PR #65, opened 2019-06-28
feat: gentx for sikka

sunnya97:


sunnya97:
@karzak Oh oops, should I make a new pull request?

PR #127, opened 2019-07-03
Genesis hash ninja

hash-ninja:
testnet 1.1


karzak:

PR #161, opened 2019-07-03
add forbole

kwunyeung:


karzak:

kwunyeung:
I was thinking leaving some tokens in the account for operations. It must be `10000000000000ukva` now?

karzak:
10000000000000 was just so the network comes up faster. If you want an additional amount for operations that’s fine too.

kwunyeung:
I have updated it anyway. Maybe I can get some tokens from the faucet if it will work. Or I can unbond some after live. It only takes 3 days to unbond on testnet which will be ok.

PR #166, opened 2019-07-03
feat: gentx for easy2stake

chinezupwnz:


karzak:

chinezupwnz:
Ok, i will resubmit

PR #218, opened 2019-07-04
feat: gentx for TRGCapital

sperlic:


sperlic:
@karzak Fixed. Thanks.

PR #252, opened 2019-09-11
Update SDK version

rhuairahrighairidh:


cli_test/cli_test.go
```diff
@@ -14,15 +14,17 @@ import (
 	"testing"
 	"time"
 
-	"github.com/tendermint/tendermint/crypto/ed25519"
-
 	"github.com/stretchr/testify/require"
 
+	"github.com/tendermint/tendermint/crypto/ed25519"
```
karzak: What is this file for? It doesn't seem to run when I do `go test -mod=readonly ./cli_test -run=*` it says `no test files`.

simulations/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+In AWS batch you define:
+
+- a "compute environment"--just how many machines you want (and of what kind)
+- a "job queue"--just a place to put jobs (pairs them with a compute environment)
+- a "job definition"--a template for jobs
+
+Then to run stuff you create "jobs" and submit them to a job queue.
+
+The number of machine running auto-scales to match the number of jobs. When there are no jobs there are no machines, so you don't pay for anything.
```
karzak: typo: `machine` -> `machines`

cli_test/cli_test.go
rhuairahrighairidh: Those tests have to be run with a build tag "cli_test". I added a test-all command to the makefile to avoid having to remember everything. We should probably make the sims and the cli tests use the same format.

rhuairahrighairidh:
Looked into AppStateDeterminism output - the failures were expected msg failures.

PR #256, opened 2019-10-10
feat: pin to cosmos-sdk 64a2741

karzak:


Makefile
```diff
@@ -108,6 +108,9 @@ test-all: build
 	@# AppStateDeterminism does not use Seed flag
 	@go test ./app -run TestAppStateDeterminism      -Enabled -Commit -NumBlocks=100 -BlockSize=200         -v -timeout 24h
 
+test:
+	@go test ./...
+
```
rhuairahrighairidh: Nit: This doesn't seem super necessary, it will only test the ./app package which only contains a couple of minor tests.

karzak: It will test all modules in the `x` folder as well.

PR #272, opened 2019-11-27
Draft params for pricefeed

karzak:
Most relevant changes to review are in `types`. This allows the following functionalities via param-change proposals:

For Pricefeed:
* Adding and removing assets
* Making assets active/inactive
* Adding and removing oracles from specific assets
* Adding feeds in new currencies (EUR, JPY, etc)

For CDP:
* Adding and removing collaterals
* Adjusting debt limit, liquidation ratio for collaterals
* Adding and removing debt assets
* Adjusting global debt limits for debt assets
* Toggling the `CircuitBreaker` to initiate system restructuring 

Interested to hear if there are features you would like to see added/removed.


rhuairahrighairidh:
Can't think of other params we need for pricefeed, these seem good.
Had some thoughts on naming:
* The Asset name confused me a bit. It's more like a pair of assets, or a market.
* If AssetCode is used to link CurrentPrices to Assets, then it will need to be contain both the base and quote assets (eg “BTC/USD”). So a name like MarketID or AssetPairID or something might be a bit clearer.
* What’s the reason for the String functions? If we can do without them the less code to maintain.
* random idea: PostedPrice could inherit from CurrentPrice, which could be rename " Price" to be more generic.

karzak:
> Can't think of other params we need for pricefeed, these seem good.
> Had some thoughts on naming:
> 
> * The Asset name confused me a bit. It's more like a pair of assets, or a market.
> * If AssetCode is used to link CurrentPrices to Assets, then it will need to be contain both the base and quote assets (eg “BTC/USD”). So a name like MarketID or AssetPairID or something might be a bit clearer.
👍I like MarketID 

> * What’s the reason for the String functions? If we can do without them the less code to maintain.

Follows convention from `cosmos-sdk` (for example. https://github.com/cosmos/cosmos-sdk/blob/master/x/gov/types/params.go). I think if we want to have query endpoints for Params, we need to have string methods. I would support keeping them. 

> * random idea: PostedPrice could inherit from CurrentPrice, which could be rename " Price" to be more generic.
👍Sounds cleaner

karzak:
Removed CDP changes for a separate PR. This is ready for re-review.

PR #275, opened 2019-12-09
R4R: CDP types and methods

karzak:
TODO 
- [x] msgs
- [x] rebase and re-integrate tests using new testing pattern


x/cdp/keeper/cdp.go
```diff
@@ -0,0 +1,407 @@
...
+		dp, found := k.GetDebt(ctx, dc.Denom)
+		if !found {
+			return types.ErrDebtNotSupported(k.codespace, dc.Denom)
+		}
+		if sdk.NewCoins(dc).IsAnyGT(dp.DebtLimit) {
+			return types.ErrExceedsDebtLimit(k.codespace, sdk.NewCoins(dc), dp.DebtLimit)
+		}
+		if dc.Amount.LT(dp.DebtFloor) {
+			return types.ErrBelowDebtFloor(k.codespace, sdk.NewCoins(dc), dp.DebtFloor)
+		}
```
denalimarsh: Should a DebtFloor check also be added somewhere in `DepositCollateral()`/`ValidatePrincipalDraw()` to prevent dust with additional principal coin types? May allow us to combine this function with `ValidatePrincipalDraw()`, but I'm not sure I fully understand all the potential implications of such an addition.

x/cdp/types/keys.go
```diff
@@ -10,6 +17,159 @@ const (
...
+//    - cdps are prefix by denom prefix so we can iterate over cdps of one type
+//    - uses : as separator, otherwise the cdp with ID 11 would be selected when iterating over denom with prefix 1
+// - 0x02<collateralDenomPrefix>:<collateralDebtRatio_Bytes>:<cdpID_Bytes>: cdpID
+// - Ox03: nextCdpID
+// - 0x04: debtDenom
+// - 0x05<depositState>:<cdpID>:<depositorAddr_bytes>: Deposit
+// - 0x06<denom>:totalPrincipal
+// - 0x07<denom>:feeRate
+// - 0x08:previousBlockTime
+// - 0x20 - 0xff are reserved for collaterals
```
rhuairahrighairidh: Remove this comment? Correct me if I'm wrong but collateral types are assigned byte ids, which as used in store keys, but their namespace doesn't conflict with the other store keys. So they can start at 0x00.

x/cdp/keeper/cdp.go
```diff
@@ -0,0 +1,407 @@
...
+	db, _ := k.GetDenomPrefix(ctx, denom)
+	store.Set(types.CollateralRatioKey(db, id, collateralRatio), types.GetCdpIDBytes(id))
+}
+
+// RemoveCdpCollateralRatioIndex deletes the cdp id from the store's index of cdps by collateral type and collateral to debt ratio
+func (k Keeper) RemoveCdpCollateralRatioIndex(ctx sdk.Context, denom string, id uint64, collateralRatio sdk.Dec) {
+	store := prefix.NewStore(ctx.KVStore(k.key), types.CollateralRatioIndexPrefix)
+	db, _ := k.GetDenomPrefix(ctx, denom)
+	store.Delete(types.CollateralRatioKey(db, id, collateralRatio))
+}
```
rhuairahrighairidh: In the `staking` module, any index updates are written into the Get/Set methods. I kind of like that approach - reduces code repitition in higher level keeper methods, and avoids the posibility that indexes are not updated when they should be.
Do you think it's worth doing for CDP indexes?

x/cdp/keeper/deposit.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+	oldCollateralToDebtRatio := k.CalculateCollateralToDebtRatio(ctx, cdp.Collateral, cdp.Principal.Add(cdp.AccumulatedFees))
+	k.RemoveCdpCollateralRatioIndex(ctx, cdp.Collateral[0].Denom, cdp.ID, oldCollateralToDebtRatio)
+
+	cdp.AccumulatedFees = cdp.AccumulatedFees.Add(fees)
+	cdp.FeesUpdated = ctx.BlockTime()
+	cdp.Collateral = cdp.Collateral.Sub(collateral)
+	k.SetCDP(ctx, cdp)
+	deposit.Amount = deposit.Amount.Sub(collateral)
+	k.SetDeposit(ctx, deposit)
```
rhuairahrighairidh: if`deposit.Amount` is reduced down to zero, the deposit is never removed from state. Think this could be exploited to add arbitrary numbers of zero size deposits.

x/cdp/types/keys.go
```diff
@@ -10,6 +17,159 @@ const (
...
+
+var sep = []byte(":")
+
+// Keys for cdp store
+// Items are stored with the following key: values
+// - 0x00<cdpOwner_Bytes>: []cdpID
+//    - One cdp owner can control one cdp per collateral type
+// - 0x01<collateralDenomPrefix>:<cdpID_Bytes>: CDP
+//    - cdps are prefix by denom prefix so we can iterate over cdps of one type
+//    - uses : as separator, otherwise the cdp with ID 11 would be selected when iterating over denom with prefix 1
```
rhuairahrighairidh: Pretty sure separators aren't needed. If each part of a key is a known fixed length so there's no need to rely on a special `:` byte to seprate them. The collateralDenomPrefix is fixed at one byte right?
Maybe just update comment, can't see any harm in leaving separators in.

x/cdp/keeper/cdp.go
karzak: Not quite sure what the suggestion is here. Looking through `staking`, the `Set` methods are pretty small and there are separate methods for indexes (ie `SetValidatorByPowerIndex`). The main difference appears to be that there is only one key prefix, which I'm not sure I can accomplish. I thought about the various indexes and iterators that were desirable for CDPs and couldn't come up with an approach that had fewer keys.

x/cdp/keeper/deposit.go
```diff
@@ -0,0 +1,216 @@
...
+func (k Keeper) WithdrawCollateral(ctx sdk.Context, owner sdk.AccAddress, depositor sdk.AccAddress, collateral sdk.Coins) sdk.Error {
+	err := k.ValidateCollateral(ctx, collateral)
+	if err != nil {
+		return err
+	}
+	cdp, found := k.GetCdpByOwnerAndDenom(ctx, owner, collateral[0].Denom)
+	if !found {
+		return types.ErrCdpNotFound(k.codespace, owner, collateral[0].Denom)
+	}
+	// withdrawals blocked if cdp is in liquidation
```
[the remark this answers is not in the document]
karzak: Not clear we reduce duplication here, as the `GetCdpByOwnerAndDenom` call is necessary, so to replace it was a 'ValidateCDPFound()' method wouldn't do much.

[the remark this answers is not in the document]
denalimarsh: Was referring to lines 73-79. It appears that my single line selection during reviews is buggy and automatically selects a 4 line block with the single selected line at the bottom, noticed it over in https://github.com/Kava-Labs/kava/pull/276 as well. That seems to be the source of the confusion here.

x/cdp/keeper/cdp.go
rhuairahrighairidh: ah yeah looking at staking again it's only the `SetUnbondingDelegations` and `SetRedelegations` methods that update their indexes. Didn't see validator indexes are not updated in `SetValidator`.

The idea was that I couldn't think of any reason you would want to set a CDP in the store, but not update the CDP indexes. Having this happen by accident could be bad. Therefore it seems safer to combine SetCDP and IndexCdpByCollateralRatio into one function.

PR #280, opened 2020-01-13
R4R: Move liquidator functions to cdp module

karzak:
Features:

CollateralAuctions are created immediately when the cdp is seized
Implemented netting of surplus and debt
Implemented starting of surplus/debt auctions 
Added gov denom to genesis state
Added params for surplus and debt auction thresholds


x/cdp/keeper/auctions.go
```diff
@@ -0,0 +1,189 @@
...
+					}
+					surplusToBurn = surplusToBurn.Sub(balance)
+				} else {
+					err = k.supplyKeeper.BurnCoins(ctx, types.LiquidatorMacc, sdk.NewCoins(sdk.NewCoin(dp.Denom, surplusToBurn)))
+					if err != nil {
+						panic(err)
+					}
+					surplusToBurn = sdk.ZeroInt()
+				}
+			}
```
rhuairahrighairidh: It looks like this can get caught in a loop on the first dp (if it's balance is < netAmount)

x/cdp/keeper/auctions.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+	return
+}
+
+// AuctionCollateral creates auctions from the input deposits which attempt to raise the corresponding amount of debt
+func (k Keeper) AuctionCollateral(ctx sdk.Context, deposits types.Deposits, debt sdk.Int, bidDenom string) {
+	auctionSize := k.getAuctionSize(ctx, deposits[0].Amount[0].Denom)
+	partialAuctionDeposits := partialDeposits{}
+	totalCollateral := deposits.SumCollateral()
+	for deposits.SumCollateral().GTE(auctionSize) {
```
rhuairahrighairidh: If total deposits are < auctionSize then no auction will be started, but the cdp will still have been seized

x/cdp/keeper/auctions.go
```diff
@@ -0,0 +1,189 @@
...
+		for i, dep := range deposits {
+			// create auctions from individual deposits that are larger than the auction size
+			k.CreateAuctionsFromDeposit(ctx, &dep, &debt, &totalCollateral, auctionSize, bidDenom)
+			if !dep.Amount[0].Amount.IsZero() {
```
rhuairahrighairidh: Found these loops a bit hard to understand, comments might help explain why things are happening

x/cdp/keeper/auctions.go
```diff
@@ -0,0 +1,189 @@
...
+	return
+}
+
+// AuctionCollateral creates auctions from the input deposits which attempt to raise the corresponding amount of debt
+func (k Keeper) AuctionCollateral(ctx sdk.Context, deposits types.Deposits, debt sdk.Int, bidDenom string) {
+	auctionSize := k.getAuctionSize(ctx, deposits[0].Amount[0].Denom)
+	partialAuctionDeposits := partialDeposits{}
+	totalCollateral := deposits.SumCollateral()
+	for deposits.SumCollateral().GTE(auctionSize) {
+		for i, dep := range deposits {
```
denalimarsh: Placeholder variables for denom, amount

x/cdp/keeper/auctions.go
```diff
@@ -0,0 +1,189 @@
...
+			}
+		}
+	}
+	if partialAuctionDeposits.SumCollateral().GT(sdk.ZeroInt()) {
+		k.CreateAuctionFromPartialDeposits(ctx, partialAuctionDeposits, &debt, &totalCollateral, partialAuctionDeposits.SumCollateral(), bidDenom)
+	}
+}
+
+// CreateAuctionsFromDeposit creates auctions from the input deposit until there is less than auctionSize left on the deposit
+func (k Keeper) CreateAuctionsFromDeposit(ctx sdk.Context, dep *types.Deposit, debt *sdk.Int, totalCollateral *sdk.Int, auctionSize sdk.Int, principalDenom string) {
```
denalimarsh: This method would be simpler if it used local variables to keep track of the data instead of updating the state via ref in each iteration of the loop. Return local variables and use them to update the state back in the primary function.

x/cdp/keeper/params.go
```diff
@@ -79,3 +78,19 @@ func (k Keeper) getLiquidationRatio(ctx sdk.Context, denom string) sdk.Dec {
 	}
 	return cp.LiquidationRatio
 }
+
+func (k Keeper) getLiquidationPenalty(ctx sdk.Context, denom string) sdk.Dec {
+	cp, found := k.GetCollateral(ctx, denom)
+	if !found {
+		panic(fmt.Sprintf("collateral not found: %s", denom))
```
denalimarsh: Will panic if collateral type type 'random' is removed in between CDP deposit of 'random' and liquidation of the CDP.

x/cdp/keeper/params.go
```diff
@@ -79,3 +78,19 @@ func (k Keeper) getLiquidationRatio(ctx sdk.Context, denom string) sdk.Dec {
...
+	}
+	return cp.LiquidationPenalty
+}
+
+func (k Keeper) getAuctionSize(ctx sdk.Context, denom string) sdk.Int {
+	cp, found := k.GetCollateral(ctx, denom)
+	if !found {
+		panic(fmt.Sprintf("collateral not found: %s", denom))
```
denalimarsh: ditto

x/cdp/keeper/auctions.go
karzak: Good catch! The outer and inner loops should be switched.

x/cdp/keeper/auctions.go
karzak: Good catch, should be checking for greater than zero.

x/cdp/keeper/params.go
karzak: Will file an issue on this topic. It's worth documenting and thinking about how we want to handle it, but for now, it would be unlikely for governance to accept a proposal to remove a collateral that still had active cdps.

x/cdp/keeper/auctions.go
```diff
@@ -38,29 +43,53 @@ func (pd partialDeposits) SumDebt() (sum sdk.Int) {
...
+			if err != nil {
+				return err
+			}
+			debt = debt.Sub(debtChange)
+			totalCollateral = totalCollateral.Sub(collateralChange)
+			dep.Amount = sdk.NewCoins(sdk.NewCoin(dep.Amount[0].Denom, dep.Amount[0].Amount.Sub(collateralChange)))
+			// if there is leftover collateral that is less than a lot
+			if !dep.Amount.IsZero() {
+				collateralAmount := dep.Amount[0].Amount
+				collateralDenom := dep.Amount[0].Denom
```
denalimarsh: Can move these up a couple of lines and replace `dep.Amount[0].Denom`, `dep.Amount[0].Amount` on line 59 with the local variables.

rhuairahrighairidh:
nit: In `seize_test.go` in setup, a new rand source doesn't need to be set up for each loop iteration.

x/cdp/keeper/auctions.go
```diff
@@ -0,0 +1,233 @@
...
+func (k Keeper) NetSurplusAndDebt(ctx sdk.Context) sdk.Error {
+	totalSurplus := k.GetTotalSurplus(ctx, types.LiquidatorMacc)
+	debt := k.GetTotalDebt(ctx, types.LiquidatorMacc)
+	netAmount := sdk.MinInt(totalSurplus, debt)
+	if netAmount.GT(sdk.ZeroInt()) {
+		surplusToBurn := netAmount
+		err := k.supplyKeeper.BurnCoins(ctx, types.LiquidatorMacc, sdk.NewCoins(sdk.NewCoin(k.GetDebtDenom(ctx), netAmount)))
+		if err != nil {
+			return err
+		}
```
rhuairahrighairidh: This if could be
```go
if netAmount.IsZero() {
    return nil
}
surplusToBurn := ...
```
It would remove a level of indentation from the subsequent section.

x/cdp/keeper/auctions.go
```diff
@@ -0,0 +1,233 @@
...
+		for surplusToBurn.GT(sdk.ZeroInt()) {
+			for _, dp := range k.GetParams(ctx).DebtParams {
+				balance := k.supplyKeeper.GetModuleAccount(ctx, types.LiquidatorMacc).GetCoins().AmountOf(dp.Denom)
+				if balance.LT(netAmount) {
```
rhuairahrighairidh: Should this be`if balance.LT(surplusToBurn) {`?
`netAmount` doesn't change in this loop.

x/cdp/keeper/auctions.go
```diff
@@ -0,0 +1,233 @@
...
+					totalCollateral = totalCollateral.Sub(collateralChange)
+					// reset partial deposits and update the deposit amount
+					partialAuctionDeposits = partialDeposits{}
+					dep.Amount = sdk.NewCoins(sdk.NewCoin(collateralDenom, collateralAmount.Sub(partialAmount)))
+				}
+			}
+			if dep.Amount.IsZero() {
+				// remove the deposit from the slice if it is empty
+				deposits = append(deposits[:i], deposits[i+1:]...)
+				i--
```
rhuairahrighairidh: my editor is saying `i--` is an ineffective assignment. Maye although it's updated here, the next for loop iteration overwrites it?
Could also be a false positive

Issue #291, opened 2020-01-15
Apply debt floor to repayments


karzak:
When a user repays stable coin debt, they should be blocked if the resulting debt balance is below the debt floor and the debt is not completely repaid.

PR #314, opened 2020-01-17
Apply debt floor to repayments

karzak:
Closes: #291

### DIFF
--- a/x/cdp/keeper/draw.go
+++ b/x/cdp/keeper/draw.go
@@ -85,14 +85,14 @@ func (k Keeper) RepayPrincipal(ctx sdk.Context, owner sdk.AccAddress, denom stri
 	if !found {
 		return types.ErrCdpNotFound(k.codespace, owner, denom)
 	}
-	err := k.ValidatePaymentCoins(ctx, cdp, payment)
-	if err != nil {
-		return err
-	}
 
 	// calculate fees
 	periods := sdk.NewInt(ctx.BlockTime().Unix()).Sub(sdk.NewInt(cdp.FeesUpdated.Unix()))
 	fees := k.CalculateFees(ctx, cdp.Principal.Add(cdp.AccumulatedFees), periods, cdp.Collateral[0].Denom)
+	err := k.ValidatePaymentCoins(ctx, cdp, payment, cdp.Principal.Add(cdp.AccumulatedFees).Add(fees))
+	if err != nil {
+		return err
+	}
 
 	// calculate fee and principal payment
 	feePayment, principalPayment := k.calculatePayment(ctx, cdp.AccumulatedFees.Add(fees), payment)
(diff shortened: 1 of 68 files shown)


x/cdp/keeper/draw.go
```diff
@@ -175,6 +175,13 @@ func (k Keeper) ValidatePaymentCoins(ctx sdk.Context, cdp types.CDP, payment sdk
 		}
 		return types.ErrInvalidPaymentDenom(k.codespace, cdp.ID, principalDenoms, paymentDenoms)
 	}
+	for _, dc := range payment {
+		dp, _ := k.GetDebt(ctx, dc.Denom)
```
rhuairahrighairidh: nit: name `GetDebtParam` might be a bit clearer

x/cdp/keeper/draw.go
```diff
@@ -175,6 +175,13 @@ func (k Keeper) ValidatePaymentCoins(ctx sdk.Context, cdp types.CDP, payment sdk
...
+		proposedBalance := cdp.Principal.AmountOf(dc.Denom).Sub(dc.Amount)
+		if proposedBalance.GT(sdk.ZeroInt()) && proposedBalance.LT(dp.DebtFloor) {
```
rhuairahrighairidh: doesn't catch case where payment is > principal
looks like it's not caught elsewehere, so someone could over repay

x/cdp/keeper/draw.go
```diff
@@ -175,8 +175,11 @@ func (k Keeper) ValidatePaymentCoins(ctx sdk.Context, cdp types.CDP, payment sdk
...
 		return types.ErrInvalidPaymentDenom(k.codespace, cdp.ID, principalDenoms, paymentDenoms)
 	}
+	if payment.IsAnyGT(debt) {
+		return types.ErrPaymentExceedsDebt(k.codespace, payment, debt)
+	}
 	for _, dc := range payment {
-		dp, _ := k.GetDebt(ctx, dc.Denom)
+		dp, _ := k.GetDebtParam(ctx, dc.Denom)
 		proposedBalance := cdp.Principal.AmountOf(dc.Denom).Sub(dc.Amount)
 		if proposedBalance.GT(sdk.ZeroInt()) && proposedBalance.LT(dp.DebtFloor) {
```
rhuairahrighairidh: This doesn't take into account the fees, so it looks like someone could choose an amount to repay that leaves the remainder under the debt floor.

rhuairahrighairidh: Actually since fees increase every block, it will be hard to specify the exact amount to repay to hit zero principle.
Maybe the cli could calulate the amount for the users? Or they specify a max-repayment which repays untill all debt is cancelled?

x/cdp/keeper/draw.go
```diff
@@ -200,10 +197,16 @@ func (k Keeper) ReturnCollateral(ctx sdk.Context, cdp types.CDP) {
...
 }
 
-func (k Keeper) calculatePayment(ctx sdk.Context, fees sdk.Coins, payment sdk.Coins) (sdk.Coins, sdk.Coins) {
+func (k Keeper) calculatePayment(ctx sdk.Context, owed sdk.Coins, fees sdk.Coins, payment sdk.Coins) (sdk.Coins, sdk.Coins) {
 	// divides repayment into principal and fee components, with fee payment applied first.
+
 	feePayment := sdk.NewCoins()
 	principalPayment := sdk.NewCoins()
+	overpayment := sdk.NewCoins()
+	if payment.IsAnyGT(owed) {
```
rhuairahrighairidh: `IsAllGT`? could be a negative coins panic in the if block

karzak: Kind of a moot point, since we're only allowing a single stable coin at this point, but neither `IsAnyGT` nor `IsAllGT` is safe. For example, if the payment due was [500usdx, 500susd] and the user input [600usdx,400susd], `IsAny` would panic, and `IsAll` would overpay. The proper was is to compare each denom and then subtract only the ones that are overpaying. 

I'll change to `IsAll` for now and leave a TODO if/when we support multiple stable denoms.

Issue #320, opened 2020-01-19
Fee calculation per block does not sum to fee calculation on liquidation


rhuairahrighairidh:
- Total fees are calculated per block, based on the block interval, and rounded.
- Fees on liquidation are calculated based on the total time since last update, and rounded.

The rounding per block means that total fees summed over blocks is < the fees calculated at liquidation.
Since new fees are minted per block, there ends up not being enough to transfer at liquidation.

karzak:
The main way this could pose a problem is when there are very few (or one) CDPs in the system, and they all get liquidated simultaneously. 

Short term, I think the fix is to limit the debt amount sent to auction to `min(debt in the cdp, debt held by the cdp module account)` 

Long term, we should remove the source of precision loss, which I believe requires something like https://github.com/cosmos/cosmos-sdk/issues/4399

What are your thoughts @rhuairahrighairidh?

rhuairahrighairidh:
Doing a min would fix the immediate problem. Though I’d be concerned that since the errors compound over time they may cause other problems.

rhuairahrighairidh:
Currently we have sort of have two versions of the same calculation.
One that calculates fees for one long period, the other that calculates total fees by summing over many smaller periods.
Both of which are “wrong” to some degree because of rounding. I believe maker gets around this by only doing the calculation once, when the global accumulator thing is updated. That keeps the two wrong by the same amount.

karzak:
Closed via #326 

A new issue can be created addressing overall issue of precision loss.

PR #326, opened 2020-01-21
Fix seize panic

rhuairahrighairidh:
Fix effects of #320 on liquidation when there's only one CDP.


denalimarsh:
@rhuairahrighairidh conflicts need resolution

PR #329, opened 2020-01-23
Fix: Avoid panics when adding cdp collaterals via governance

karzak:


denalimarsh:
Tested successfully, this is cleared for merge

PR #331, opened 2020-01-23
[WIP] example proposal.json files for param changes

denalimarsh:
Each proposal.json file is co-located with an associated genesis.json file which describes the initial state. Proposal deposit/voting time is set to 5 minutes.


contrib/proposal_examples/addcollateral/addcollateral2.json
```diff
@@ -0,0 +1,82 @@
...
+      ]
+    },
+    {
+      "subspace": "cdp",
+      "key": "DebtParams",
+      "value": [
+          {
+              "conversion_factor": "6",
+              "debt_floor": "10000000",
+              "debt_limit": [
```
karzak:
```
              "debt_limit": [
                {
                  "amount": "3000000000000",
                  "denom": "usdx"
                }
              ],
```

Remove. DebtParam doesn't have `debt_limit` field anymore.

contrib/proposal_examples/addcollateral/addcollateral2.json
```diff
@@ -0,0 +1,82 @@
...
+          }
+      ]
+    },
+    {
+      "subspace": "cdp",
+      "key": "GlobalDebtLimit",
+      "value": [
+          {
+              "global_debt_limit": [
+                  {
```
karzak: The value should just be an `sdk.Coins`:

```
"value":  [
                  {
                  "amount": "3000000000000",
                  "denom": "usdx"
                  }
              ]
```

denalimarsh:
Implemented requested changes and tested. Bunch of duplicate commits from force pushing the branch though. 

There's 1-2 other cli txs which surpass the 200,000 gas limit as well, I've been using 500,000 on every cli tx to sidestep the issue. Is there any reason why we shouldn't increase the `max_bytes` and/or `max_gas` in genesis.config?

contrib/proposal_examples/addcollateral/addcollateral2.json
denalimarsh: https://github.com/Kava-Labs/kava/pull/331/commits/2130feff6fabfb6a510cc26dde0208ac0b8531a0

contrib/proposal_examples/addcollateral/addcollateral2.json
denalimarsh: https://github.com/Kava-Labs/kava/pull/331/commits/2130feff6fabfb6a510cc26dde0208ac0b8531a0

karzak:
`max_gas`and `max_bytes` just control the per block limits. I wouldn't feel comfortable altering them due to potential issues with latency in gossiping larger blocks. The default gas estimate is set in https://github.com/cosmos/cosmos-sdk/blob/master/client/flags/flags.go and is not configurable. I'll see if anyone has opened as issue regarding that, as it should probably be configurable per-client.

Issue #337, opened 2020-01-24
Pricefeed doesn't emit events


rhuairahrighairidh:
should probably emit one when it handles messages, seems to be standard

denalimarsh:
https://github.com/Kava-Labs/kava/pull/343

PR #343, opened 2020-01-24
R4R: add events to pricefeed module

denalimarsh:
Addresses https://github.com/Kava-Labs/kava/issues/337

### DIFF
--- a/x/pricefeed/keeper/keeper.go
+++ b/x/pricefeed/keeper/keeper.go
@@ -93,6 +93,13 @@ func (k Keeper) SetCurrentPrices(ctx sdk.Context, marketID string) sdk.Error {
 	if !ok {
 		return types.ErrInvalidMarket(k.codespace, marketID)
 	}
+	// store current price
+	validPrevPrice := true
+	prevPrice, err := k.GetCurrentPrice(ctx, marketID)
+	if err != nil {
+		validPrevPrice = false
+	}
+
 	prices := k.GetRawPrices(ctx, marketID)
 	var notExpiredPrices []types.CurrentPrice
 	// filter out expired prices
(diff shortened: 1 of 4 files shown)


x/pricefeed/keeper/keeper.go
```diff
@@ -67,6 +68,16 @@ func (k Keeper) SetPrice(
...
 		}
 
+		// Emit an event containing the oracle's new price
+		ctx.EventManager().EmitEvent(
+			sdk.NewEvent(
+				types.EventTypeOracleUpdatedPrice,
+				sdk.NewAttribute(types.AttributeMarketID, fmt.Sprintf("%s", marketID)),
+				sdk.NewAttribute(types.AttributeOracle, fmt.Sprintf("%s", oracle.String())),
+				sdk.NewAttribute(types.AttributeMarketPrice, fmt.Sprintf("%s", price.String())),
+				sdk.NewAttribute(types.AttributeExpiry, fmt.Sprintf("%s", expiry.String())),
```
rhuairahrighairidh: I think this will output the time in go's custom text format. I would use some standard like a unix timestamp or https://golang.org/pkg/time/#Time.MarshalText to make it easier for clients to parse

Whichever you use, could you also update `auction`'s events to use the same?

x/pricefeed/keeper/keeper.go
```diff
@@ -102,11 +113,21 @@ func (k Keeper) SetCurrentPrices(ctx sdk.Context, marketID string) sdk.Error {
...
 
+	// Emit an event for market price update
+	ctx.EventManager().EmitEvent(
+		sdk.NewEvent(
+			types.EventTypeMarketPriceUpdated,
+			sdk.NewAttribute(types.AttributeMarketID, fmt.Sprintf("%s", marketID)),
+			sdk.NewAttribute(types.AttributeMarketPrice, fmt.Sprintf("%s", medianPrice.String())),
+		),
+	)
+
```
rhuairahrighairidh: nit: this will ouput an event every block even if the price hasn't changed

karzak: One option is to remove this event and replace it with one that only emits if the price changes. This would requires storing the previous value of each market. I don't have a strong preference between keeping it as-is or changing it.

denalimarsh: I did a thing in https://github.com/Kava-Labs/kava/pull/343/commits/fad21d56f2ccf03a6a75cd299991af0263397af1 but can't test it since it seems only events emitted by txs can be queried. It doesn't break anything but idk if it's working properly.

x/pricefeed/keeper/keeper.go
denalimarsh: https://github.com/Kava-Labs/kava/pull/343/commits/c8e1a85868d6f9f082a3a0753eb9a22049abab9f

PR #345, opened 2020-01-24
Display auction type and phase when querying auctions

### LINKED ISSUE #340: Update auctions querier to differentiate auction types and phase
- [ ] The querier shows all auctions as base auctions and does not differentiate between collateral, debt, or surplus auctions. It also doesn't indicate if a collateral auction is in forward or reverse phase. It should be updated so that it differentiates these states.

karzak:
Closes #340


rhuairahrighairidh:
got a error on unmarshalling the results of the query in the cli handlers, looks like they should be the wrapper types rather than the auction types.
Other than that looks nice

denalimarsh:
@karzak I'm not seeing phases for collateral auctions anymore, even after commit https://github.com/Kava-Labs/kava/pull/345/commits/8f588cb8ed79ae35b516797f836d9a24b5ad7fbf

PR #346, opened 2020-01-24
Add query for cdp deposits

### LINKED ISSUE #338: Users cannot query individual deposits
There is no querier to return deposits. If someone deposited to a CDP there is no way to query this seperately from the total deposited to the CDP.

Maybe deposits could be returned alongside the a CDP when querying for a specific CDP?

karzak:
Closes #338

### DIFF
--- a/x/cdp/client/cli/query.go
+++ b/x/cdp/client/cli/query.go
@@ -161,13 +161,13 @@ $ %s query %s cdps-by-ratio uatom 1.5
 // QueryCdpDepositsCmd returns the command handler for querying the deposits of a particular cdp
 func QueryCdpDepositsCmd(queryRoute string, cdc *codec.Codec) *cobra.Command {
 	return &cobra.Command{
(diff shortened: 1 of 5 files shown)


x/cdp/client/cli/query.go
```diff
@@ -157,6 +158,49 @@ $ %s query %s cdps-by-ratio uatom 1.5
 	}
 }
 
+// QueryCdpDepositsCmd returns the command handler for querying the deposits of a particular cdp
+func QueryCdpDepositsCmd(queryRoute string, cdc *codec.Codec) *cobra.Command {
+	return &cobra.Command{
+		Use:   "cdp deposits [owner-addr] [collateral-name]",
```
rhuairahrighairidh: Looks like cobra doesn't like individual commands being more than one word, can't get this to run from the cli. I think it gets confused between this and the `cdp` command.

PR #365, opened 2020-02-07
[R4R] Add Savings Rate

karzak:
Features:
* Adds a savings rate that is defined as the percentage of surplus that is redistributed back to holders of stable coins. The percentage is called `SavingsRate` set within the `DebtParam` for each debt coin. 
* Adds a param, `SavingsDistributionFrequency`, which is the number of seconds between distributions of the savings rate (default is 24 hours). 
* Modifies the surplus auction code to only start surplus auctions with the surplus that is not directed towards the savings rate

Fixes: 
* Ensures that cdp and liquidator module accounts are created at genesis
* Updates the alias file, which had gotten a bit out of date.


x/cdp/keeper/savings.go
```diff
@@ -0,0 +1,71 @@
...
+// ApplySavingsRate distributes surplus that has accumulated in the liquidator account to address holding stable coins according the the savings rate
+func (k Keeper) ApplySavingsRate(ctx sdk.Context, debtDenom string) sdk.Error {
+	dp, found := k.GetDebtParam(ctx, debtDenom)
+	if !found {
+		return types.ErrDebtNotSupported(k.codespace, debtDenom)
+	}
+	liquidatorMacc := k.supplyKeeper.GetModuleAccount(ctx, types.LiquidatorMacc)
+	totalSurplusCoins := sdk.NewCoins(sdk.NewCoin(debtDenom, liquidatorMacc.GetCoins().AmountOf(dp.Denom)))
+
+	surplusToDistribute := sdk.NewDecFromInt(liquidatorMacc.GetCoins().AmountOf(dp.Denom)).Mul(dp.SavingsRate).RoundInt()
```
denalimarsh: abci.go passes denom parameter 'usdx' to ApplySavingsRate(debtDenom) but the liquidatorMacc's debt coins are stored under denom 'debt'. Therefore,`liquidatorMacc.GetCoins().AmountOf(dp.Denom)` queries the wrong coin denom, returns 0, and the function returns on line 23 with no savings rate distribution.

denalimarsh: at least that's my working theory

x/cdp/abci.go
```diff
@@ -42,6 +47,22 @@ func BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock, k Keeper) {
...
+					sdk.NewEvent(
+						EventTypeBeginBlockerFatal,
+						sdk.NewAttribute(sdk.AttributeKeyModule, fmt.Sprintf("%s", ModuleName)),
+						sdk.NewAttribute(types.AttributeKeyError, fmt.Sprintf("%s", err)),
+					),
+				)
+			}
+		}
+		k.SetPreviousSavingsDistribution(ctx, ctx.BlockTime())
+	}
```
rhuairahrighairidh: Should paying out savings run before running surplus and debt auctions?

karzak: It should not. 
As currently implemented, surplus checks if (surplus * (1-savings_rate)) > surplus_threshold. Imagine if the surplus_threshold was 5, the current surplus is 100, and the savings_rate is 0.95. If savings rate distribution went first, 95 of the surplus would be paid out, then the surplus auction would check (5 * 0.05) = 0.25. 

Point taken on your suggestion for moving rewards to a separate module account. I agree it's hard to reason about as implemented.

karzak:
Made the suggested changes, liquidator and savings accounts are now separate as well.

x/cdp/abci.go
```diff
@@ -42,6 +47,22 @@ func BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock, k Keeper) {
...
+			err := k.ApplySavingsRate(ctx, dp.Denom)
+			if err != nil {
+				ctx.EventManager().EmitEvent(
+					sdk.NewEvent(
+						EventTypeBeginBlockerFatal,
+						sdk.NewAttribute(sdk.AttributeKeyModule, fmt.Sprintf("%s", ModuleName)),
+						sdk.NewAttribute(types.AttributeKeyError, fmt.Sprintf("%s", err)),
+					),
+				)
+			}
```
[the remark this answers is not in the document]
karzak: This is a question we should dig into further. Using the general rule "don't panic" would suggest that panicing during the operation of the chain is worse than continuing and figuring it out. However, we should do some scenario analysis and figure out if it's better to ignore that rule. Will create an issue for this.

x/cdp/types/params.go
```diff
@@ -11,34 +11,38 @@ import (
...
-	GlobalDebtLimit         sdk.Coins        `json:"global_debt_limit" yaml:"global_debt_limit"`
-	SurplusAuctionThreshold sdk.Int          `json:"surplus_auction_threshold" yaml:"surplus_auction_threshold"`
-	DebtAuctionThreshold    sdk.Int          `json:"debt_auction_threshold" yaml:"debt_auction_threshold"`
-	CircuitBreaker          bool             `json:"circuit_breaker" yaml:"circuit_breaker"`
+	CollateralParams             CollateralParams `json:"collateral_params" yaml:"collateral_params"`
+	DebtParams                   DebtParams       `json:"debt_params" yaml:"debt_params"`
+	GlobalDebtLimit              sdk.Coins        `json:"global_debt_limit" yaml:"global_debt_limit"`
+	SurplusAuctionThreshold      sdk.Int          `json:"surplus_auction_threshold" yaml:"surplus_auction_threshold"`
+	DebtAuctionThreshold         sdk.Int          `json:"debt_auction_threshold" yaml:"debt_auction_threshold"`
+	SavingsDistributionFrequency sdk.Int          `json:"savings_distribution_frequency" yaml:"savings_distribution_frequency"`
```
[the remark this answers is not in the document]
karzak: Will change, as that seems to be the pattern used in `gov`.

x/cdp/abci.go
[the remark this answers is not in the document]
karzak: #385

karzak:
Should be good for another look.

karzak:
Ran this branch on a testnet with cdps, liquidations, etc. Working as expected.

PR #369, opened 2020-02-13
R4R: update swagger to testnet 4000

karzak:
Adds the REST APIs for CDP, Auction, and Pricefeed. You can run test queries at kava-testnet-4000.kava.io


swagger-ui/swagger.yaml
```diff
@@ -364,11 +370,587 @@
...
+          200:
+            description: OK
+            schema:
+              $ref: '#/definitions/AuctionResponse'
+          400:
+            description: invalid auction id
+          500:
+            description: Internal Server Error
+    /pricefeed/postprice:
+      post:
```
rhuairahrighairidh: looks like pricefeed/postprice is still PUT not POST

PR #370, opened 2020-02-14
R4R: BEP3 module

denalimarsh:
**Description**
This PR introduces the BEP3 module. Cross-chain AtomicSwaps can be submitted by the authorized deputy, upon successful claim coins are minted to claimer and deputy. Same-chain AtomicSwaps can be submitted by any user and require coins to be held in escrow. Additional deposits are only available for same-chain AtomicSwaps and must be submitted by the original creator. Refunds are automatically processed by the EndBlocker, with escrowed tokens returned to the original creator in the case of same-chain AtomicSwaps.

**Questions**
1. `kvcli q bep3 swaps` prints AtomicSwap attribute's _swapid_ and _randomnumberhash_ as raw bytes. Are there any examples of cli queries returning a struct containing raw bytes in a cosmos module?
2. x/bep3 module tests are passing, but very recently x/crisis began failing with `panic: incorrect address length [recovered]`. Has anyone seen this before?

**Outstanding** (must be addressed before external audit)
1. Test coverage increased. Most important is swap.go keeper tests.
2. Implement any HTTP queries required by the bep3-deputy (minor)


x/bep3/keeper/htlt.go
```diff
@@ -0,0 +1,248 @@
...
+		}
+	}
+
+	ctx.EventManager().EmitEvent(
+		sdk.NewEvent(
+			types.EventTypeClaimHtlt,
+			sdk.NewAttribute(types.AttributeKeyHtltSwapID, fmt.Sprintf("%s", encodedSwapID)),
+			sdk.NewAttribute(types.AttributeKeyClaimer, fmt.Sprintf("%s", from)),
+			sdk.NewAttribute(types.AttributeKeyCoinDenom, fmt.Sprintf("%s", claimerCoins[0].Denom)),
+			sdk.NewAttribute(types.AttributeKeyCoinAmount, fmt.Sprintf("%d", claimerCoins[0].Amount.Int64())),
```
rhuairahrighairidh: Lines 175,176 will panic if someone submits a HTLT with expectedIncome=“” as claimerCoins will be empty.
In general I think using `sdk.Coin` when there should only be one coin is safer (but possibly more awkward to write).
Similarly `kepper.ValidateAsset` will panic if coins arg is nil. Not that it looks like this ever happens in the code, but could happen after some future refactor.

x/bep3/types/params.go
```diff
@@ -0,0 +1,129 @@
...
+
+// DefaultParams returns default params for bep3 module
+func DefaultParams() Params {
+	return NewParams(DefaultMinLockTime, DefaultMaxLockTime, DefaultSupportedAssets)
+}
+
+// AssetParam governance parameters for each asset within a supported chain
+type AssetParam struct {
+	Denom  string `json:"denom" yaml:"denom"`     // name of the asster
+	CoinID string `json:"coin_id" yaml:"coin_id"` // internationally recognized coin ID
```
rhuairahrighairidh: Bip44 coin type often doesn't differentiate between testnet and mainnets, so we might need another way to identify different chains. Cosmos chains all have a unique "chain-id", but I'm not sure how to create identifiers for things like eth and eth testnets.

x/bep3/types/msg.go
```diff
@@ -0,0 +1,305 @@
...
+	if err != nil {
+		panic(err)
+	}
+	return b
+}
+
+// MsgDepositHTLT defines an HTLT deposit
+type MsgDepositHTLT struct {
+	From   sdk.AccAddress `json:"from"`
+	SwapID string         `json:"swap_id"`
```
rhuairahrighairidh: Is there a reason `SwapID` is `string` compared to `[32]byte`? BEP3 lists it as 32 bytes.
The MarshalJSON method will take care of encoding it to hex automatically when a msg is encoded to json.

x/bep3/types/hash.go
```diff
@@ -0,0 +1,30 @@
...
+// CalculateSwapID calculates the hash of a RandomNumberHash, sdk.AccAddress, and string
+func CalculateSwapID(randomNumberHash string, sender sdk.AccAddress, senderOtherChain string) ([]byte, error) {
+	randomNumberHashBytes, err := HexEncodedStringToBytes(randomNumberHash)
+	if err != nil {
+		return []byte{}, err
+	}
+	senderOtherChain = strings.ToLower(senderOtherChain)
+	data := randomNumberHashBytes
+	data = append(data, []byte(sender)...)
+	data = append(data, []byte(senderOtherChain)...)
```
rhuairahrighairidh: Having `swapID` based on the string encoded address from the other chain could cause problems if they ever change the string encoding. Would it be possible to store other chain addresses as their underlying bytes instead?

x/bep3/keeper/keeper_test.go
```diff
@@ -0,0 +1,78 @@
...
+func TestHtltTestSuite(t *testing.T) {
+	suite.Run(t, new(KeeperTestSuite))
+}
+
+func htlts() types.HTLTs {
+	var htlts types.HTLTs
+	h1 := types.NewHTLT(binanceAddrs[0], kavaAddrs[0], "", "", randomNumberHashes[0], timestamps[0], coinsSingle, "bnb50000", 50500, false)
+	h2 := types.NewHTLT(binanceAddrs[1], kavaAddrs[1], "", "", randomNumberHashes[1], timestamps[1], coinsSingle, "bnb50000", 61500, false)
+	h3 := types.NewHTLT(binanceAddrs[2], kavaAddrs[2], "", "", randomNumberHashes[2], timestamps[2], coinsSingle, "bnb50000", 72500, false)
+	h4 := types.NewHTLT(binanceAddrs[3], kavaAddrs[3], "", "", randomNumberHashes[3], timestamps[3], coinsSingle, "bnb50000", 83500, false)
```
rhuairahrighairidh: The sdk formats coins as `<amount><denom>` when converting to strings

x/bep3/genesis.go
```diff
@@ -0,0 +1,31 @@
...
+		panic(fmt.Sprintf("failed to validate %s genesis state: %s", ModuleName, err))
+	}
+
+	keeper.SetParams(ctx, gs.Params)
+
+	// check if the module account exists
+	moduleAcc := supplyKeeper.GetModuleAccount(ctx, ModuleName)
+	if moduleAcc == nil {
+		panic(fmt.Sprintf("%s module account has not been set", ModuleName))
+	}
```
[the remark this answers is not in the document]
denalimarsh: The recommendation here is to remove lines 20-22, correct?

x/bep3/types/params.go
denalimarsh: Was hoping that Cosmos would follow Ethereum's standard of renaming coin Denom to indicate the network, e.g. Eth (mainnet), rEth (ropsten), and kEth (kovan), but looks like that is not the case with Cosmos.

denalimarsh: Previously the param hierarchy looked like []chain -> []assets, such that each chain had a unique ID and this was not a problem. The params have been flattened such that the bep3 module only supports one chain (i.e. the binance testnet) and assets exist under that chain. I'm very open to changing the params, but the current param structure is much simpler and also still enables us to launch the 'cross-chain testnet' from bnbchain testnet <-> kava testnet.

x/bep3/types/hash.go
denalimarsh: Yes, we could do that.

Looking into it I found that the [BEP3 documentation](https://github.com/binance-chain/BEPs/blob/master/BEP3.md#hash-timer-locked-transfer) indicates type bytes, yet the HTLTMsg in [Binance's go-sdk](https://github.com/binance-chain/go-sdk/blob/master/types/msg/msg-HTLT.go) defines them as strings:
```
RecipientOtherChain string           `json:"recipient_other_chain"`
SenderOtherChain    string           `json:"sender_other_chain"`
```

denalimarsh: Ah, found some additional info. We must implement BEP3's [common Executor interface](https://github.com/binance-chain/bep3-deputy/blob/master/common/executor.go), which defines these params in the HTLT function as strings:
```
HTLT(randomNumberHash common.Hash, timestamp int64, heightSpan int64, recipientAddr string, otherChainSenderAddr string, otherChainRecipientAddr string, outAmount *big.Int) (string, *Error)
```

x/bep3/keeper/keeper_test.go
denalimarsh: Indeed it does. Updated in https://github.com/Kava-Labs/kava/pull/370/commits/59d43fc3e0d0627c9875b5d48fe664f2133280d6.

x/bep3/keeper/htlt.go
denalimarsh: 1. Updated MsgCreateHTLT's ValidateBasic() in https://github.com/Kava-Labs/kava/pull/370/commits/fecee798f71734887c165f44e22893a4597d935a.

2. I agree that the logic here is pretty brittle. Updated ValidateAsset() in https://github.com/Kava-Labs/kava/pull/370/commits/36de1ebcaadeb0ea76f7b86709d5876f44ee1ad2. Do you have any insight into best practices for the location of these assertions? ValidateBasic() checks that the coin's amount is positive (meaning there must be at least one coin), but I feel that it may be better for the `len(coins) != 1` check to happen inside ValidateBasic() as opposed to ValidateAsset().

x/bep3/keeper/keeper_test.go
denalimarsh: Adding `expectedIncome` param validation revealed additional places where the invalid formatting was located throughout the tests. Fixed in https://github.com/Kava-Labs/kava/pull/370/commits/703353d894e5250f6d437b8d2a2e01e49e88fb6d.

denalimarsh:
> * Should `ClaimHTLT` and `RefundHTLT` be checking for expiry times? Currently looks like they can refund a HTLT at any point.

Added expiry check for claim HTLT in https://github.com/Kava-Labs/kava/pull/370/commits/4931bab72115ccc8a31d7d0ecb61933a2fe6e973. Shouldn't refunds be automatically processed for all expired HTLTs in the endblocker? HTLTs shouldn't sit in the store after expiration, but it seems better to not prevent refunds post-expiration as in the worst case it would lock funds such that they're irrecoverable. Thoughts?

x/bep3/types/msg.go
denalimarsh: Included in https://github.com/Kava-Labs/kava/pull/370/commits/d0514831d6734a722e3402739ed07fa44dfbbff3

x/bep3/types/msg.go
```diff
@@ -0,0 +1,314 @@
...
+		panic(err)
+	}
+	return b
+}
+
+// MsgClaimAtomicSwap defines a AtomicSwap claim
+type MsgClaimAtomicSwap struct {
+	From         sdk.AccAddress `json:"from"`
+	SwapID       cmm.HexBytes   `json:"swap_id"`
+	RandomNumber cmm.HexBytes   `json:"random_number"`
```
karzak: The reason it's safe to put the random number in the transaction is because the recipient address is set at the time of creation, correct?

x/bep3/keeper/swap.go
```diff
@@ -0,0 +1,340 @@
...
+	}
+
+	if atomicSwap.CrossChain {
+		err := k.ValidateCoinMint(ctx, atomicSwap.Amount)
+		if err != nil {
+			return err
+		}
+
+		// Mint full amount of this coin's associated debt coin to bep3 module for internal limit tracking
+		internalTrackingCoins, err2 := getEqualInternalTrackingCoins(atomicSwap.Amount)
```
karzak: I'm a little confused how the internal tracking coin is used. It seems like it could be replaced by a store value of `TotalSwapCoins` for each denom, and then just check that the requested swap doesn't put the total over the limit. Is there a reason to do it using coins and not just a counter?

x/bep3/keeper/swap.go
```diff
@@ -0,0 +1,340 @@
...
+	if !found {
+		return types.ErrAtomicSwapNotFound(k.codespace, swapID)
+	}
+
+	// Refund request must come from original creator or bep3 module
+	if !from.Equals(atomicSwap.From) && !k.GetBnbDeputyAddress(ctx).Equals(atomicSwap.From) {
+		return types.ErrOnlyOriginalCreator(k.codespace, from, atomicSwap.From)
+	}
+
+	if !atomicSwap.CrossChain {
```
karzak: If a swap is cross chain and it's refunded, the only thing that happens is that the swap get's deleted. Is that correct?

x/bep3/keeper/swap.go
```diff
@@ -0,0 +1,340 @@
...
+	if !found {
+		return types.ErrAssetNotSupported(k.codespace, coin.Denom)
+	}
+	if !asset.Active {
+		return types.ErrAssetNotActive(k.codespace, asset.Denom)
+	}
+	return nil
+}
+
+// ValidateCoinMint validates that an asset minted
```
karzak:
```suggestion
// ValidateCoinMint validates that an asset can be minted
```

typo?

x/bep3/types/errors.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// ErrAmountTooSmall error for when a coin amount is 0
+func ErrAmountTooSmall(codespace sdk.CodespaceType, coin sdk.Coin) sdk.Error {
+	return sdk.NewError(codespace, CodeAmountTooSmall, fmt.Sprintf("coin %s amount is below the limit for this operation", coin))
+}
+
+// ErrAtomicSwapAlreadyExists error for when an AtomicSwap with this swapID already exists
+func ErrAtomicSwapAlreadyExists(codespace sdk.CodespaceType, swapID cmn.HexBytes) sdk.Error {
+	return sdk.NewError(codespace, CodeAtomicSwapAlreadyExists, fmt.Sprintf("coin %s amount is below the limit for this operation", swapID))
```
karzak:
```suggestion
	return sdk.NewError(codespace, CodeAtomicSwapAlreadyExists, fmt.Sprintf("swap %s already exists", swapID))
```

denalimarsh: Ah, this is why I was seeing a different error message.

x/bep3/keeper/swap.go
rhuairahrighairidh: Not sure if it helps in this case, but supply has methods that return total amount of any coin.

denalimarsh:
Pushed a [major refactor](https://github.com/Kava-Labs/kava/pull/370/commits/40edb8c627ef547d0f67fc0fb5559385d250cd05) based on the BEP3 [EthSwapper.sol](https://github.com/binance-chain/bep3-smartcontracts/blob/master/ethereum/contracts/ETHAtomicSwapper.sol) smart contract. Instead of attempting to get _everything_ working at once, I simplified this PR so that it contains just core logic for cross-chain swaps - most of the changes are in keeper/swap.go. It also contains most of the requested revisions.

Instead of deleting AtomicSwaps, they're kept in the store and have their status updated. This works due to checks in Refund() and Claim(). But as this is based on Solidity programming paradigms, it'll better for us to iterate over expired AtomicSwaps and update status to expired instead of updating it on the first requested refund. Iteration logic is [here](a65ee245180c75a585742cf939618d550239d6c7), but should be brought back into the module in a follow-on PR.

Makefile
```diff
@@ -76,8 +76,8 @@ build-linux: go.sum
 	LEDGER_ENABLED=false GOOS=linux GOARCH=amd64 $(MAKE) build
 
 install: go.sum
-	go install -mod=readonly $(BUILD_FLAGS) ./cmd/kvd
-	go install -mod=readonly $(BUILD_FLAGS) ./cmd/kvcli
+	go install $(BUILD_FLAGS) ./cmd/kvd
```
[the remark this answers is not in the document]
denalimarsh: added back in https://github.com/Kava-Labs/kava/pull/370/commits/40edb8c627ef547d0f67fc0fb5559385d250cd05

x/bep3/genesis.go
```diff
@@ -0,0 +1,31 @@
...
+}
+
+// ExportGenesis writes the current store values
+// to a genesis file, which can be imported again
+// with InitGenesis
+func ExportGenesis(ctx sdk.Context, k Keeper) (data types.GenesisState) {
+	params := k.GetParams(ctx)
+	return types.NewGenesisState(params)
```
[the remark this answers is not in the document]
denalimarsh: ImportGenesis logic included in https://github.com/Kava-Labs/kava/pull/370/commits/40edb8c627ef547d0f67fc0fb5559385d250cd05. Left TODO on ExportGenesis which should be addressed prior to the approval of this PR

x/bep3/types/genesis.go
```diff
@@ -0,0 +1,42 @@
+package types
+
+import (
+	"bytes"
+)
+
+// GenesisState - all bep3 state that must be provided at genesis
```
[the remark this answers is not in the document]
denalimarsh: added in https://github.com/Kava-Labs/kava/pull/370/commits/40edb8c627ef547d0f67fc0fb5559385d250cd05

x/bep3/types/params.go
```diff
@@ -0,0 +1,134 @@
...
+	DefaultBnbDeputyAddress sdk.AccAddress = sdk.AccAddress("kava1xy7hrjy9r0algz9w3gzm8u6mrpq97kwta747gj")
+	DefaultMinLockTime      int64          = 20
+	DefaultMaxLockTime      int64          = 200
+	DefaultSupportedAssets                 = AssetParams{AssetParam{Denom: "kava", CoinID: "459", Limit: 1, Active: false}}
+)
+
+// Params governance parameters for bep3 module
+type Params struct {
+	BnbDeputyAddress sdk.AccAddress `json:"bnb_deputy_address" yaml:"bnb_deputy_address"` // Bnbchain deputy address
+	MinLockTime      int64          `json:"min_lock_time" yaml:"min_lock_time"`           // AtomicSwap minimum lock time
```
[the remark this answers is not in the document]
denalimarsh: Updated in https://github.com/Kava-Labs/kava/pull/370/commits/40edb8c627ef547d0f67fc0fb5559385d250cd05. The refactor removes references to BnbDeputyAddress param as well, but I've left it in for now until we're 100% sure that the final deliverable won't include it.

x/bep3/types/msg.go
denalimarsh: That is correct. Keeper logic has been refactored to make that clearer.

x/bep3/keeper/swap.go
denalimarsh: Added PrefixAssetSupplyKey to the store in https://github.com/Kava-Labs/kava/pull/370/commits/40edb8c627ef547d0f67fc0fb5559385d250cd05. This has the additional benefit of allowing us to import/export genesis coin supply amounts.

x/bep3/keeper/swap.go
denalimarsh: That was correct, but the new refactor now returns deposited coins to the original swap creator and updates the swap's status to 'Expired'.

x/bep3/genesis.go
```diff
@@ -0,0 +1,64 @@
...
+// ExportGenesis writes the current store values
+// to a genesis file, which can be imported again
+// with InitGenesis
+func ExportGenesis(ctx sdk.Context, k Keeper) (data types.GenesisState) {
+	params := k.GetParams(ctx)
+
+	genAssetSupplies := []sdk.Coin{}
+	// TODO: Add k.IterateAssetSupplies
+
+	genAtomicSwaps := types.GenesisAtomicSwaps{}
```
karzak: I think it would be safe to follow the CDP module pattern and just marshal/unmarshall all swaps at the time of InitGenesis/Export Genesis. Because expiry is checked at each state transition, importing expired swaps doesn't seem to pose any risk. In that case, `GenesisState` would look like. 
```
type Swaps []Swap
type GenesisState struct {
	Params      Params             `json:"params" yaml:"params"`
	AtomicSwaps Swaps `json:"atomic_swaps" yaml:"atomic_swaps"`
	Assets      []sdk.Coin         `json:"assets" yaml:"assets"`
}
```

x/bep3/types/params.go
```diff
@@ -0,0 +1,134 @@
...
+// DefaultParams returns default params for bep3 module
+func DefaultParams() Params {
+	return NewParams(DefaultBnbDeputyAddress, DefaultMinBlockLock, DefaultMaxBlockLock, DefaultSupportedAssets)
+}
+
+// AssetParam governance parameters for each asset within a supported chain
+type AssetParam struct {
+	Denom  string `json:"denom" yaml:"denom"`     // name of the asster
+	CoinID string `json:"coin_id" yaml:"coin_id"` // internationally recognized coin ID
+	Limit  int64  `json:"limit" yaml:"limit"`     // asset supply limit
```
karzak: nit: I would use `sdk.Int`

x/bep3/genesis.go
denalimarsh: Updated in https://github.com/Kava-Labs/kava/pull/370/commits/61daca47e743c4472c8c30a47d575c2cb89af591.
- Is a 'PreviousBlockHeight' required, or is that maintained by the blockchain already?
- Is InitGenesis the best place to set all AssetSupplies to `sdk.NewCoin(denom, sdk.NewInt(0))`? I'm setting each AssetSupply to 0 in CreateAtomicSwap() as they're witnessed, which feels like the wrong way to approach this.

x/bep3/keeper/swap.go
```diff
@@ -0,0 +1,204 @@
...
+			return types.ErrInvalidTimestamp(k.codespace)
+		}
+	}
+
+	// Sanity check on recipient address
+	if recipient.Empty() {
+		return sdk.ErrInvalidAddress("invalid (empty) bidder address")
+	}
+
+	err := k.validateCoinDeposit(ctx, amount)
```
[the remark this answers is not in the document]
denalimarsh: Addressed in https://github.com/Kava-Labs/kava/pull/370/commits/e2d53b089ef535245de68ac301a3ebd480cc3dcf.

x/bep3/keeper/keeper.go
```diff
@@ -40,11 +40,15 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 	return ctx.Logger().With("module", fmt.Sprintf("x/%s", types.ModuleName))
 }
 
+// ------------------------------------------
+//				Atomic Swaps
```
denalimarsh: These are centered in my IDE but spaced awkwardly on git. I like these section headers as they organize similar code into chunks. Does anyone have strong opinions for/against them?

x/bep3/keeper/swap.go
```diff
@@ -0,0 +1,204 @@
...
+			types.EventTypeClaimAtomicSwap,
+			sdk.NewAttribute(types.AttributeKeyClaimSender, fmt.Sprintf("%s", from)),
+			sdk.NewAttribute(types.AttributeKeyRecipient, fmt.Sprintf("%s", atomicSwap.Recipient)),
+			sdk.NewAttribute(types.AttributeKeyAtomicSwapID, fmt.Sprintf("%s", hex.EncodeToString(atomicSwap.GetSwapID()))),
+			sdk.NewAttribute(types.AttributeKeyRandomNumberHash, fmt.Sprintf("%s", hex.EncodeToString(atomicSwap.RandomNumberHash))),
+			sdk.NewAttribute(types.AttributeKeyRandomNumber, fmt.Sprintf("%s", randomNumber)),
+		),
+	)
+
+	return nil
```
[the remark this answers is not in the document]
denalimarsh: Updated in https://github.com/Kava-Labs/kava/pull/370/commits/414abd7f744a51584141bed8575801993455c582.

When the chain's block height is greater than a swap's expiration block its status is updated to expired. Claims can only occur before expiration, refunds can only occur after expiration. Once the swap is claimed/refunded, it is removed from the primary swap store.

By deleting the swap immediately, the completed status isn't used. Another option is update status to completed upon claim/refund, then delete all swaps with completed status after the delay. I favor the current implementation as 'completed' swaps have no utility.

x/bep3/types/params.go
denalimarsh: Addressed in https://github.com/Kava-Labs/kava/pull/370/commits/f73778e91076600eed68817474ce2e1e92593daa.

x/bep3/keeper/swap.go
```diff
@@ -0,0 +1,204 @@
...
+			sdk.NewAttribute(types.AttributeKeySender, fmt.Sprintf("%s", atomicSwap.Sender)),
+			sdk.NewAttribute(types.AttributeKeyAtomicSwapID, fmt.Sprintf("%s", hex.EncodeToString(atomicSwap.GetSwapID()))),
+			sdk.NewAttribute(types.AttributeKeyRandomNumberHash, fmt.Sprintf("%s", hex.EncodeToString(atomicSwap.RandomNumberHash))),
+		),
+	)
+
+	return nil
+}
+
+// GetAllAtomicSwaps returns all AtomicSwaps from the store
```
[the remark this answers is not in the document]
denalimarsh: Updated in https://github.com/Kava-Labs/kava/pull/370/commits/e2d53b089ef535245de68ac301a3ebd480cc3dcf.

Worth noting that the CDP module defines GetAllCdps() in cdp.go but IterateAllCdps() in keeper.go. The Auction and CDP modules approach store management very differently in  keeper.go. Auction defines basic get/set/iterate methods in keeper.go with complex logic elsewhere. CDP module defines basic store interactions in cdp.go along with the associated complex logic.

I think both approaches have merit - we should discuss a standardized solution.

x/bep3/keeper/keeper.go
denalimarsh: Related to https://github.com/Kava-Labs/kava/pull/370#discussion_r383744618

rhuairahrighairidh: I like headers as well. Though I don't see them much in other code, not sure if there's reasons against them I'm missing.

x/bep3/genesis.go
denalimarsh: Discussed with @rhuairahrighairidh - we agree that our options are:
- leaving the code as is i.e. set an asset's supply the first time it is witnessed
- adding asset supply to params and requiring that asset supply params are always 0 in InitGenesis, then using them to set asset supplies in the store; upon chain restart, the exported asset supplies in the genesis state are used to update the supply from 0 to the actual count.

x/bep3/client/cli/tx.go
```diff
@@ -0,0 +1,160 @@
...
+		},
+	}
+}
+
+// GetCmdClaimAtomicSwap cli command for claiming an atomic swap
+func GetCmdClaimAtomicSwap(cdc *codec.Codec) *cobra.Command {
+	return &cobra.Command{
+		Use:     "claim [swap-id] [random-number]",
+		Short:   "claim coins in an atomic swap using the secret number",
+		Example: "bep3 claim 6682c03cc3856879c8fb98c9733c6b0c30758299138166b6523fe94628b1d3af 123456789 --from accA",
```
karzak:
```suggestion
		Example: fmt.Sprintf("%s query %s claim 6682c03cc3856879c8fb98c9733c6b0c30758299138166b6523fe94628b1d3af 123456789 --from accA", version.ClientName, types.ModuleName),
```
nit: examples are a little nicer if they mirror cli syntax exactly

x/bep3/client/rest/query.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/gorilla/mux"
+
+	"github.com/cosmos/cosmos-sdk/client/context"
+	"github.com/cosmos/cosmos-sdk/types/rest"
+	"github.com/kava-labs/kava/x/bep3/types"
+)
+
+func registerQueryRoutes(cliCtx context.CLIContext, r *mux.Router) {
+	r.HandleFunc("/bep3/parameters", queryParamsHandlerFn(cliCtx)).Methods("GET")
```
karzak: Make note to add rest client methods, can be in a follow-on pr

x/bep3/genesis.go
```diff
@@ -0,0 +1,51 @@
...
+func InitGenesis(ctx sdk.Context, keeper Keeper, supplyKeeper types.SupplyKeeper, gs GenesisState) {
+	if err := gs.Validate(); err != nil {
+		panic(fmt.Sprintf("failed to validate %s genesis state: %s", ModuleName, err))
+	}
+
+	// Set each AssetSupply and store total coin count
+	totalAssetSupplyCoins := sdk.NewCoins()
+	for _, asset := range gs.AssetSupplies {
+		keeper.SetAssetSupply(ctx, asset, []byte(asset.Denom))
+		totalAssetSupplyCoins.Add(sdk.NewCoins(asset))
```
karzak:
```suggestion
		totalAssetSupplyCoins = totalAssetSupplyCoins.Add(sdk.NewCoins(asset))
```

I assume you want to update the state of `totalAssetSupplyCoins`

x/bep3/genesis.go
```diff
@@ -0,0 +1,51 @@
...
+
+	// Set each AtomicSwap and store total coin count
+	totalAtomicSwapCoins := sdk.NewCoins()
+	for _, swap := range gs.AtomicSwaps {
+		if swap.Validate() != nil {
+			panic(fmt.Sprintf("invalid swap %s", swap.GetSwapID()))
+		}
+		keeper.SetAtomicSwap(ctx, swap)
+		keeper.InsertIntoByBlockIndex(ctx, swap)
+		totalAtomicSwapCoins.Add(swap.GetModuleAccountCoins())
```
karzak: see above

x/bep3/types/params.go
```diff
@@ -0,0 +1,134 @@
...
+
+// Parameter keys
+var (
+	KeyBnbDeputyAddress = []byte("BnbDeputyAddress")
+	KeyMinBlockLock     = []byte("MinBlockLock")
+	KeyMaxBlockLock     = []byte("MaxBlockLock")
+	KeySupportedAssets  = []byte("SupportedAssets")
+
+	AbsoluteMaximumBlockLock int64 = 10000
+	AbsoluteMinimumBlockLock int64 = 10
```
karzak: Given ~7 second block times, this is only ~1 minute, I would increase it to 50

x/bep3/types/params.go
```diff
@@ -0,0 +1,134 @@
...
+	DefaultMinBlockLock      int64 = 20
```
karzak: I would set this to 80

x/bep3/types/params.go
```diff
@@ -0,0 +1,134 @@
...
+	DefaultMaxBlockLock      int64 = 200
```
karzak: I would set this to 600

x/bep3/types/params.go
```diff
@@ -0,0 +1,134 @@
...
+	DefaultSupportedAssets         = AssetParams{AssetParam{Denom: "kava", CoinID: "459", Limit: sdk.NewInt(1), Active: false}}
```
karzak:
```suggestion
	DefaultSupportedAssets         = AssetParams{AssetParam{Denom: "ukava", CoinID: "459", Limit: sdk.NewInt(1), Active: false}}
```

x/bep3/keeper/swap.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	// The heightSpan period should be more than 10 minutes and less than one week
+	// Assume average block time interval is 10 second. 10 mins = 60 blocks, 1 week = 60480 blocks
+	if heightSpan < k.GetMinBlockLock(ctx) || heightSpan > k.GetMaxBlockLock(ctx) {
+		return types.ErrInvalidHeightSpan(k.codespace, heightSpan, k.GetMinBlockLock(ctx), k.GetMaxBlockLock(ctx))
+	}
+
+	// Validate that timestamp is within reasonable bounds
+	// Assuming a block time of 10 seconds, the timestamp must be in range [-15 mins, 30 mins] of the current time
```
karzak: Confused as to what this validation is checking. It says it's checking the timestamp (which I assume is a unix timestamp like 1582914311), but is then checking the block height. Clarify in comment perhaps.

x/bep3/abci.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	abci "github.com/tendermint/tendermint/abci/types"
+)
+
+// BeginBlocker runs at the start of every block
+func BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock, k Keeper) {
+	err := k.UpdateExpiredAtomicSwaps(ctx)
+	if err != nil {
```
karzak: To match bnb implementation, we can add method to delete  expired swaps after 2 weeks. Can be in follow on PR.

x/bep3/types/msg.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+	if !msg.Amount.IsAllPositive() {
+		return sdk.ErrInternal(fmt.Sprintf("the swapped out coin must be positive"))
+	}
+	if len(msg.ExpectedIncome) > MaxExpectedIncomeLength {
+		return sdk.ErrInternal(fmt.Sprintf("the length of expected income should be less than %d", MaxExpectedIncomeLength))
+	}
+	expectedIncomeCoins, err := sdk.ParseCoins(msg.ExpectedIncome)
+	if err != nil || expectedIncomeCoins == nil {
+		return sdk.ErrInternal(fmt.Sprintf("expected income %s must be in valid format e.g. kava10000", msg.ExpectedIncome))
```
karzak:
```suggestion
		return sdk.ErrInternal(fmt.Sprintf("expected income %s must be in valid format e.g. 10000ukava", msg.ExpectedIncome))
```

nit: match cosmos sdk parsing rather than bnbchain

denalimarsh: Fixed in https://github.com/Kava-Labs/kava/pull/370/commits/ac92e26999cf32a5fae8296ba92a05b1da0cf357

x/bep3/types/params.go
denalimarsh: Fixed in https://github.com/Kava-Labs/kava/pull/370/commits/ac92e26999cf32a5fae8296ba92a05b1da0cf357

x/bep3/types/params.go
denalimarsh: Set in https://github.com/Kava-Labs/kava/pull/370/commits/ac92e26999cf32a5fae8296ba92a05b1da0cf357

x/bep3/types/params.go
denalimarsh: Set in https://github.com/Kava-Labs/kava/pull/370/commits/ac92e26999cf32a5fae8296ba92a05b1da0cf357

x/bep3/types/params.go
denalimarsh: Set in https://github.com/Kava-Labs/kava/pull/370/commits/ac92e26999cf32a5fae8296ba92a05b1da0cf357.

x/bep3/genesis.go
denalimarsh: Fixed in https://github.com/Kava-Labs/kava/pull/370/commits/aefaf6fdc9582bf6c83b94700b06889d27bdb39b. Updated to make naming conventions as well it make it more readable in https://github.com/Kava-Labs/kava/pull/370/commits/ac92e26999cf32a5fae8296ba92a05b1da0cf357.

x/bep3/genesis.go
denalimarsh: Fixed in https://github.com/Kava-Labs/kava/pull/370/commits/aefaf6fdc9582bf6c83b94700b06889d27bdb39b

x/bep3/client/rest/query.go
denalimarsh: Implemented `/bep3/swap` in https://github.com/Kava-Labs/kava/pull/370/commits/ac92e26999cf32a5fae8296ba92a05b1da0cf357. Follow-on PR is still needed to add the other rest client methods.

x/bep3/client/cli/query.go
```diff
@@ -0,0 +1,173 @@
...
+			randomNumberHash := types.CalculateRandomHash(randomNumber, timestamp)
+			return cliCtx.PrintOutput(hex.EncodeToString(randomNumberHash))
+		},
+	}
+}
+
+// QueryCalcSwapIDCmd calculates the swapID for a random number hash, sender, and sender other chain
+func QueryCalcSwapIDCmd(queryRoute string, cdc *codec.Codec) *cobra.Command {
+	return &cobra.Command{
+		Use:     "calc-id [randomNumberHash] [sender] [senderOtherChain]",
```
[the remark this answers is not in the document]
denalimarsh: Addressed in https://github.com/Kava-Labs/kava/pull/370/commits/ac92e26999cf32a5fae8296ba92a05b1da0cf357

x/bep3/client/cli/tx.go
denalimarsh: Updated cmds in tx.go in https://github.com/Kava-Labs/kava/pull/370/commits/ac92e26999cf32a5fae8296ba92a05b1da0cf357

x/bep3/keeper/swap.go
denalimarsh: Updated comment to clarify that timestamp represents block height in https://github.com/Kava-Labs/kava/pull/370/commits/ac92e26999cf32a5fae8296ba92a05b1da0cf357. I agree that this is confusing and will make a note to check with Binance to confirm that this implementation of timestamp is correct.

x/bep3/client/rest/query.go
denalimarsh: Issue https://github.com/Kava-Labs/kava/issues/384

x/bep3/keeper/swap.go
denalimarsh: This also assumes a block time of 10 seconds as opposed to 7 seconds and should be updated to match the new 7-second standard.

x/bep3/abci.go
denalimarsh: Issue https://github.com/Kava-Labs/kava/issues/384

x/bep3/client/cli/tx.go
```diff
@@ -35,10 +36,11 @@ func GetTxCmd(cdc *codec.Codec) *cobra.Command {
 // GetCmdCreateAtomicSwap cli command for creating atomic swaps
 func GetCmdCreateAtomicSwap(cdc *codec.Codec) *cobra.Command {
 	return &cobra.Command{
-		Use:     "create [to] [recipient-other-chain] [sender-other-chain] [hashed-secret] [timestamp] [coins] [expected-income] [height-span] [cross-chain]",
-		Short:   "create a new atomic swap",
-		Example: "bep3 create kava15qdefkmwswysgg4qxgqpqr35k3m49pkx2jdfnw 0x9eB05a790e2De0a047a57a22199D8CccEA6d6D5A '' 0677bd8a303dd981810f34d8e5cc6507f13b391899b84d3c1be6c6045a17d747 100 100xrp 99xrp 1000 true --from accA",
-		Args:    cobra.ExactArgs(9),
+		Use:   "create [to] [recipient-other-chain] [sender-other-chain] [hashed-secret] [timestamp] [coins] [expected-income] [height-span] [cross-chain]",
+		Short: "create a new atomic swap",
+		Example: fmt.Sprintf("%s query %s create kava15qdefkmwswysgg4qxgqpqr35k3m49pkx2jdfnw 0x9eB05a790e2De0a047a57a22199D8CccEA6d6D5A '' 0677bd8a303dd981810f34d8e5cc6507f13b391899b84d3c1be6c6045a17d747 100 100xrp 99xrp 100 true --from accA",
```
karzak:
```suggestion
		Example: fmt.Sprintf("%s tx %s create kava15qdefkmwswysgg4qxgqpqr35k3m49pkx2jdfnw 0x9eB05a790e2De0a047a57a22199D8CccEA6d6D5A '' 0677bd8a303dd981810f34d8e5cc6507f13b391899b84d3c1be6c6045a17d747 100 100xrp 99xrp 100 true --from accA",
```

my bad 😂  - this should be tx, not query

x/bep3/client/cli/tx.go
```diff
@@ -99,7 +101,7 @@ func GetCmdClaimAtomicSwap(cdc *codec.Codec) *cobra.Command {
 	return &cobra.Command{
 		Use:     "claim [swap-id] [random-number]",
 		Short:   "claim coins in an atomic swap using the secret number",
-		Example: "bep3 claim 6682c03cc3856879c8fb98c9733c6b0c30758299138166b6523fe94628b1d3af 123456789 --from accA",
+		Example: fmt.Sprintf("%s query %s claim 6682c03cc3856879c8fb98c9733c6b0c30758299138166b6523fe94628b1d3af 123456789 --from accA", version.ClientName, types.ModuleName),
```
karzak:
```suggestion
		Example: fmt.Sprintf("%s tx %s claim 6682c03cc3856879c8fb98c9733c6b0c30758299138166b6523fe94628b1d3af 123456789 --from accA", version.ClientName, types.ModuleName),
```

x/bep3/client/cli/tx.go
```diff
@@ -134,7 +136,7 @@ func GetCmdRefundAtomicSwap(cdc *codec.Codec) *cobra.Command {
 	return &cobra.Command{
 		Use:     "refund [swap-id]",
 		Short:   "refund the coins in an atomic swap",
-		Example: "bep3 refund 6682c03cc3856879c8fb98c9733c6b0c30758299138166b6523fe94628b1d3af --from accA",
+		Example: fmt.Sprintf("%s query %s refund 6682c03cc3856879c8fb98c9733c6b0c30758299138166b6523fe94628b1d3af --from accA", version.ClientName, types.ModuleName),
```
karzak:
```suggestion
		Example: fmt.Sprintf("%s tx %s refund 6682c03cc3856879c8fb98c9733c6b0c30758299138166b6523fe94628b1d3af --from accA", version.ClientName, types.ModuleName),
```

denalimarsh: https://github.com/Kava-Labs/kava/pull/370/commits/22e94e1af802b66aa28faa6c564ae0772250be84

x/bep3/client/cli/tx.go
denalimarsh: https://github.com/Kava-Labs/kava/pull/370/commits/22e94e1af802b66aa28faa6c564ae0772250be84

x/bep3/client/cli/tx.go
denalimarsh: https://github.com/Kava-Labs/kava/pull/370/commits/22e94e1af802b66aa28faa6c564ae0772250be84

denalimarsh:
Awesome! Pushed the final revisions, this PR is ready for merge.

PR #372, opened 2020-02-20
[R4R] Prepare develop for merge

karzak:
* Updates the structure of swagger and contrib directories to disambiguate mainnet versus testnet resources. 
* Updates readme to reflect proper versions for mainnet/testnet


contrib/testnet-4000/README.md
```diff
@@ -0,0 +1,62 @@
...
+Resources and examples for running and interacting with kava-testnet-4000
+
+## Rest server requests
+
+### Setup
+
+ Before making a request, query account information for the signing account. Note the 'accountnumber' and 'sequence' fields, we'll need them later in order to send our request:
+
+```bash
+    kvcli q auth account $(kvcli keys show accB -a)
```
denalimarsh: This is here because the cli can't access the DB once the rest sever is started. It works, but users will want to make multiple requests and won't want to manually track their sequence. We could change it to an HTTP request to the rest server endpoint `auth/account/{address}` and move it to the "Create CDP example request" section.

Pretty sure I originally put these steps together, so the confusing user experience is on me. If we're in agreement about the change I can push it up.

PR #376, opened 2020-02-21
[R4R] This is the implementation of the total circulating supply

jmaheswaran:
Will implement [todo]:
- cli interface
- rest api endpoint
- iterate over all account objects and on each object count up how much is in "circulating supply"
- implement different totaling functionality for each of the different types of accounts (ModuleAccounts, vesting accounts (linear vesting), periodic vesting accounts and any others)
- implements tests
- documentation of implemented functionality


jmaheswaran:
Fixed include issues - all CircleCI tests pass now

jmaheswaran:
Builds successfully and unit tests pass

jmaheswaran:
Querying commands for `validator-vesting` module do not appear in the list of query commands when running a `kava` node and querying using `kvcli`. I'm not sure why this is. does `GetQueryValidatorVesting` in `client/query.go` need to be registered somewhere?

x/validator-vesting/internal/keeper/querier.go
```diff
@@ -0,0 +1,62 @@
...
+
+			// periodic vesting account
+			vacc, ok := acc.(vesting.PeriodicVestingAccount)
+			if ok {
+				balance := vacc.GetCoins().AmountOf("ukava")
+				if balance.IsZero() {
+					return false
+				}
+				spendableBalance := vacc.SpendableCoins(ctx.BlockTime()).AmountOf("ukava")
+				circulatingSupply = circulatingSupply.Add(sdk.MinInt(balance, spendableBalance))
```
rhuairahrighairidh: What's the reason for calling `MinInt`?
Is there a situation when `SpendableCoins` are greater than an account's coins?

x/validator-vesting/client/cli/query.go
```diff
@@ -0,0 +1,48 @@
...
+			cliCtx := context.NewCLIContext().WithCodec(cdc)
+
+			res, _, err := cliCtx.QueryWithData(fmt.Sprintf("custom/%s/%s", queryRoute, types.QueryCirculatingSupply), nil)
+			if err != nil {
+				fmt.Printf("could not get total circulating supply\n")
+				return nil
+			}
+
+			var out types.TotalCirculatingSupply
+			cdc.MustUnmarshalJSON(res, &out)
```
rhuairahrighairidh: Looks like your querier returns a json encoded `sdk.Int`, but this will try and unmarshal it into the `TotalCirculatingSupply` type.
You could maybe get away without the type and just unmarshal the querier response into an `sdk.Int`.

x/validator-vesting/internal/keeper/querier.go
```diff
@@ -0,0 +1,62 @@
...
+			// exclude module account
+			_, ok := acc.(supplyexported.ModuleAccountI)
+			if ok {
+				return false
+			}
+
+			// periodic vesting account
+			vacc, ok := acc.(vesting.PeriodicVestingAccount)
+			if ok {
+				balance := vacc.GetCoins().AmountOf("ukava")
```
rhuairahrighairidh: I think the "ukava" coin denom shouldn't be hard coded.
Possible options:
- get the denom from another module, ie `stakingKeeper.BondDenom(ctx)`
- just return all coins, and leave it up to the user to parse out the coins they're interested in

x/validator-vesting/internal/keeper/querier.go
karzak: Indeed yes. Spendable coins is a badly named method from the vesting account interface that in fact returns the theoretical spendable balance based the progress of vesting at the current block time. If some of those coins have already been spent, the spendable coins will be greater than the balance.

jmaheswaran:
Tested `cli` and `rest` endpoints with @karzak

x/validator-vesting/client/cli/query.go
jmaheswaran: I fixed this to use `sdk.Int` instead

x/validator-vesting/client/cli/query.go
```diff
@@ -0,0 +1,72 @@
...
+			}
+
+			var out sdk.Dec
+			cdc.MustUnmarshalJSON(res, &out)
+			return cliCtx.PrintOutput(out)
+		},
+	}
+}
+
+// QueryTotalSupplyCmd queries the toal supply of ukava
```
karzak:
```suggestion
// QueryTotalSupplyCmd queries the total supply of ukava
```

jmaheswaran:
Fixed typo

PR #380, opened 2020-02-26
Add min bid increments

### LINKED ISSUE #373: Change auction end blocker to begin blocker
Begin blockers are better for actions that should be triggered at certain points in time. ie the auctions should be closed after their end time. This is better in a begin blocker as the auctions can be closed before there is the chance for any txs to interact with it.

Note: current implementation checks for auction expiry in the keeper methods so should be safe, but using a begin blocker avoids closed auctions hanging around for a block after they're expired.

rhuairahrighairidh:
- adds min bid increments to auctions based on percentages of previous bid
- also fixes #373


x/auction/types/params_test.go
```diff
@@ -1,33 +1,101 @@
...
+				MaxAuctionDuration:  1 * time.Hour,
+				BidDuration:         24 * time.Hour,
+				IncrementSurplus:    d("0.05"),
+				IncrementDebt:       d("0.05"),
+				IncrementCollateral: d("0.05"),
+			},
+			true,
+		},
+		{
+			"negative increment surplus",
```
karzak: These tests are hitting error "bid duration param cannot be larger than max auction duration" and not the expected errors.

x/auction/types/params_test.go
```diff
@@ -1,33 +1,101 @@
...
+				MaxAuctionDuration:  1 * time.Hour,
+				BidDuration:         24 * time.Hour,
+				IncrementSurplus:    d("-0.05"),
+				IncrementDebt:       d("0.05"),
+				IncrementCollateral: d("0.05"),
+			},
+			true,
+		},
+		{
+			"negative increment debt",
```
karzak: see above

x/auction/types/params_test.go
```diff
@@ -1,33 +1,101 @@
...
+				MaxAuctionDuration:  1 * time.Hour,
+				BidDuration:         24 * time.Hour,
+				IncrementSurplus:    d("0.05"),
+				IncrementDebt:       d("-0.05"),
+				IncrementCollateral: d("0.05"),
+			},
+			true,
+		},
+		{
+			"negative increment collateral",
```
karzak: ditto

x/auction/types/params_test.go
rhuairahrighairidh: good catch

x/auction/types/params.go
```diff
@@ -17,26 +17,36 @@ const (
 	DefaultBidDuration time.Duration = 1 * time.Hour
 )
 
-// Parameter keys
 var (
+	// DefaultIncrement is the smallest percent change a new bid must have from the old one
+	DefaultIncrement sdk.Dec = sdk.MustNewDecFromStr("0.05")
```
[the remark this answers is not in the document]
rhuairahrighairidh: agreed
some references for the future:
- [art auction bid increments](https://theauctioncollective.com/blog/bidding-increments/) they seem to be in the range 5%-8%
- maker is using 3% collateral, 2% surplus, 3% debt

PR #381, opened 2020-02-26
[R4R] Update fees risky cdps

jmaheswaran:
This is for the implementation to automatically update fees for risky CDPs


x/cdp/keeper/fees.go
```diff
@@ -27,6 +28,46 @@ func (k Keeper) CalculateFees(ctx sdk.Context, principal sdk.Coins, periods sdk.
...
+// set cdp.FeesLastUpdated to the current block time (ie. ctx.BlockTime())
+
+// UpdateFeesForRiskyCdps calculates fees for risky CDPs
+// The overall logic is first select the CDPs with 10% of the liquidation ratio
+// Then we call calculate fees on each of those CDPs
+// Next we store the result of the fees in the cdp.AccumulatedFees field
+// Finally we set the cdp.FeesUpdated time to the current block time (ctx.BlockTime()) since that
+// is when we made the update
+// TODO - QUESTION - this method signature should only take (ctx sdk.Context, cp types.Params) as parameters? Is this correct?
+// TODO - QUESTION - is types.CollateralParam the correct type?
```
jmaheswaran: @karzak Are these the correct function parameter?

x/cdp/keeper/fees.go
```diff
@@ -27,6 +28,46 @@ func (k Keeper) CalculateFees(ctx sdk.Context, principal sdk.Coins, periods sdk.
...
+func (k Keeper) UpdateFeesForRiskyCdps(ctx sdk.Context, cp types.CollateralParam) {
+
+	// first calculate the target ratio based on liquidation ratio plus ten percent
+	value, err := sdk.NewDecFromStr("1.1")
+	if err != nil {
+		fmt.Printf("got error: %s", err) //  TODO - QUESTION - is there another method we should use for error logging instead of
```
jmaheswaran: @karzak I just printed the error out here - is there an error logging function or anything that I should call instead?

x/cdp/keeper/fees.go
```diff
@@ -27,6 +28,46 @@ func (k Keeper) CalculateFees(ctx sdk.Context, principal sdk.Coins, periods sdk.
...
+		// get the number of periods
+		periods := sdk.NewInt(ctx.BlockTime().Unix()).Sub(sdk.NewInt(cdp.FeesUpdated.Unix()))
+		// now calcuate and store additional fees
+		additionalFees := k.CalculateFees(ctx, cdp.Principal, periods, cp.Denom)
+		// now add the additional fees to the accumulated fees for the cdp
+		cdp.AccumulatedFees.Add(additionalFees)
+		// and set the fees updated time to the current block time since we just updated it
+		cdp.FeesUpdated = ctx.BlockTime()
+		return false // TODO - QUESTION - is this the correct thing to return?? The return is
+		// not used in this case as far as I can tell
```
jmaheswaran: @karzak I'm not too sure what the `boolean` return should be representing. Is returning `false` a default or should it vary depending on something else?

x/cdp/abci.go
```diff
@@ -21,6 +21,10 @@ func BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock, k Keeper) {
 			k.HandleNewDebt(ctx, cp.Denom, dp.Denom, timeElapsed)
 		}
 
+		// IMPLEMENT HERE FEE UPDATING
+		// call our update fees method for the risky cdps
+		k.UpdateFeesForRiskyCdps(ctx, cp)
```
jmaheswaran: @karzak Is just calling the function here OK? Since it updates the state of the CDPs I assuming just calling is sufficient and no need to do any other state updates?

jmaheswaran:
Built successfully and existing tests passed

jmaheswaran:
What would be a good testing strategy for this update?

x/cdp/abci.go
karzak:
```suggestion
		err := k.UpdateFeesForRiskyCdps(ctx, cp)
```

karzak: Yes, but I would have the function return an error and follow the pattern of emitting the error if it's not `nil` (see how errors are emitted a few lines below)

```
if err != nil {
			ctx.EventManager().EmitEvent(
				sdk.NewEvent(
					EventTypeBeginBlockerFatal,
					sdk.NewAttribute(sdk.AttributeKeyModule, fmt.Sprintf("%s", ModuleName)),
					sdk.NewAttribute(types.AttributeKeyError, fmt.Sprintf("%s", err)),
				),
			)
		}
```

x/cdp/keeper/fees.go
```diff
@@ -1,6 +1,7 @@
 package keeper
 
 import (
+	"fmt"
```
karzak:
```suggestion
```

x/cdp/keeper/fees.go
```diff
@@ -27,6 +28,46 @@ func (k Keeper) CalculateFees(ctx sdk.Context, principal sdk.Coins, periods sdk.
...
+
+// UpdateFeesForRiskyCdps calculates fees for risky CDPs
+// The overall logic is first select the CDPs with 10% of the liquidation ratio
+// Then we call calculate fees on each of those CDPs
+// Next we store the result of the fees in the cdp.AccumulatedFees field
+// Finally we set the cdp.FeesUpdated time to the current block time (ctx.BlockTime()) since that
+// is when we made the update
+// TODO - QUESTION - this method signature should only take (ctx sdk.Context, cp types.Params) as parameters? Is this correct?
+// TODO - QUESTION - is types.CollateralParam the correct type?
+func (k Keeper) UpdateFeesForRiskyCdps(ctx sdk.Context, cp types.CollateralParam) {
```
karzak:
```suggestion
func (k Keeper) UpdateFeesForRiskyCdps(ctx sdk.Context, cp types.CollateralParam) sdk.Error {
```

x/cdp/keeper/fees.go
karzak:
```suggestion
		return err
```

x/cdp/keeper/fees.go
karzak: the boolean tells the `IterateCdpsByCollateralRatio` function if it should stop iterating. Since we aren't looking for specific cdp or some other termination condition, we can always return `false`

x/cdp/keeper/fees.go
```diff
@@ -27,6 +28,46 @@ func (k Keeper) CalculateFees(ctx sdk.Context, principal sdk.Coins, periods sdk.
...
+		// now calcuate and store additional fees
+		additionalFees := k.CalculateFees(ctx, cdp.Principal, periods, cp.Denom)
+		// now add the additional fees to the accumulated fees for the cdp
+		cdp.AccumulatedFees.Add(additionalFees)
+		// and set the fees updated time to the current block time since we just updated it
+		cdp.FeesUpdated = ctx.BlockTime()
+		return false // TODO - QUESTION - is this the correct thing to return?? The return is
+		// not used in this case as far as I can tell
+	})
+	// this function does not return anything
```
karzak:
```suggestion
	return nil
```

go.sum
```diff
@@ -354,6 +354,7 @@ golang.org/x/tools v0.0.0-20181030221726-6c7e314b6563/go.mod h1:n7NCudcB/nEzxVGm
 golang.org/x/tools v0.0.0-20190114222345-bf090417da8b/go.mod h1:n7NCudcB/nEzxVGmLbDWY5pfWTLqBcC2KZ6jyYvM4mQ=
 golang.org/x/tools v0.0.0-20190311212946-11955173bddd/go.mod h1:LCzVGOaR6xXOjkQ3onu1FJEFr0SW1gC7cKk1uF8kGRs=
 golang.org/x/tools v0.0.0-20190425150028-36563e24a262/go.mod h1:RgjU9mgBXZiqYHBnxXauZ1Gv1EHHAz9KjViQ78xBX0Q=
+golang.org/x/tools v0.0.0-20190524140312-2c0ae7006135 h1:5Beo0mZN8dRzgrMMkDp0jc8YXQKx9DiJ2k1dkvGsn5A=
```
karzak:
```suggestion
```

generally best not to push changes to `go.sum`

x/cdp/keeper/fees.go
karzak: `(ctx sdk.Context, cp types.CollateralParam)` are fine. Since it looks like only `cp.Denom` is used I would change it to

`(ctx sdk.Context, collateralDenom string)` and replace references to `cp.Denom` with `collateralDenom`

x/cdp/keeper/fees.go
```diff
@@ -27,6 +28,46 @@ func (k Keeper) CalculateFees(ctx sdk.Context, principal sdk.Coins, periods sdk.
...
+	// now iterate over all the cdps based on collateral ratio
+	k.IterateCdpsByCollateralRatio(ctx, cp.Denom, targetRatio, func(cdp types.CDP) bool {
+		// get the number of periods
+		periods := sdk.NewInt(ctx.BlockTime().Unix()).Sub(sdk.NewInt(cdp.FeesUpdated.Unix()))
+		// now calcuate and store additional fees
+		additionalFees := k.CalculateFees(ctx, cdp.Principal, periods, cp.Denom)
+		// now add the additional fees to the accumulated fees for the cdp
+		cdp.AccumulatedFees.Add(additionalFees)
+		// and set the fees updated time to the current block time since we just updated it
+		cdp.FeesUpdated = ctx.BlockTime()
```
karzak:
```suggestion
		cdp.FeesUpdated = ctx.BlockTime()
		collateralToDebtRatio := k.CalculateCollateralToDebtRatio(ctx, cdp.Collateral, cdp.Principal.Add(cdp.AccumulatedFees))
	        k.SetCdpAndCollateralRatioIndex(ctx, cdp, collateralToDebtRatio)
```

Note, you haven't updated the state of the CDP in the store. https://github.com/Kava-Labs/kava/blob/4d21230925baf19d55d0fe3ed59703939348dfb6/x/cdp/keeper/deposit.go#L47 is a pretty generic example of updating the cdp state after altering the fees.

x/cdp/keeper/fees.go
jmaheswaran: 🆗

x/cdp/keeper/fees.go
jmaheswaran: 🆗 👍

x/cdp/abci.go
```diff
@@ -21,6 +21,19 @@ func BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock, k Keeper) {
 			k.HandleNewDebt(ctx, cp.Denom, dp.Denom, timeElapsed)
 		}
 
+		// call our update fees method for the risky cdps
+		e := k.UpdateFeesForRiskyCdps(ctx, cp.Denom)
```
karzak:
```suggestion
		err := k.UpdateFeesForRiskyCdps(ctx, cp.Denom)
```

nit: stick to convention of the rest of the codebase and call error results `err`

x/cdp/keeper/fees.go
```diff
@@ -27,6 +27,38 @@ func (k Keeper) CalculateFees(ctx sdk.Context, principal sdk.Coins, periods sdk.
 	return newFees
 }
 
+// UpdateFeesForRiskyCdps calculates fees for risky CDPs
+// The overall logic is first select the CDPs with 10% of the liquidation ratio
+// Then we call calculate fees on each of those CDPs
+// Next we store the result of the fees in the cdp.AccumulatedFees field
+// Finally we set the cdp.FeesUpdated time to the current block time (ctx.BlockTime()) since that
+// is when we made the update
+func (k Keeper) UpdateFeesForRiskyCdps(ctx sdk.Context, collateralDenom string) sdk.Error {
```
karzak:
```suggestion
func (k Keeper) UpdateFeesForRiskyCdps(ctx sdk.Context, collateralDenom string, marketID string) sdk.Error {
        price, err := k.pricefeedKeeper.GetCurrentPrice(ctx, marketID)
	if err != nil {
		return err
	}
	// first calculate the target ratio based on liquidation ratio plus ten percent
	value, err := sdk.NewDecFromStr("1.1")
	if err != nil {
		return err
	}
	targetRatio := k.getLiquidationRatio(ctx, collateralDenom).Mul(value) // corresponds to 110% of the liquidation ratio
        normalizedRatio := sdk.OneDec().Quo(price.Price.Quo(liquidationRatio))
        
	// now iterate over all the cdps based on collateral ratio
	k.IterateCdpsByCollateralRatio(ctx, collateralDenom, normalizedRatio, func(cdp types.CDP) bool {
```

CDPs need to be selected based on the collateral:debt ratio. Currently, we are selecting based on the usd_value_of_collateral:debt ratio. This suggestions gives the correct approach. Note that `CollateralParam` contains `MarketID` as a field, so we can call the function with `cp.MarketID`.

jmaheswaran:
Added updates. Builds and passes tests now.

jmaheswaran:
Completed tests, debugged the issues around the 110% cutoff normalized ratio, tests cover a CDP where the fees should be updated (and are, and check the fee amount), and also a case where a CDP should NOT have fees updated (and the fees correctly are not updated). No errors are emitted originating from the update fees for risky cdps code, they are just passed through from other modules (for example getting the current price) so there is no "unhappy path" to cover with these tests.

jmaheswaran:
Note: a test in another piece of code is breaking `TestGetSetCdp` since that relies on `	t, found := suite.keeper.GetCDP(suite.ctx, "xrp", types.DefaultCdpStartingID)` ie the default starting CDP ID. Should we remove the CDPs after we add them to fix or do something else?

jmaheswaran:
Fixed the breaking test by first checking the number of CDPs and offsetting from there. All test now pass successfully.

jmaheswaran:
Made changes to `fees_test.go` - tests are still all passing on CircleCI. Will now make changes to `cdp_test.go` and check the tests again.

jmaheswaran:
Looks like all tests do indeed pass even with the reverted `cdp_test`. Recommended changes have now been made.

x/cdp/keeper/fees.go
```diff
@@ -27,6 +27,46 @@ func (k Keeper) CalculateFees(ctx sdk.Context, principal sdk.Coins, periods sdk.
...
 
+// UpdateFeesForRiskyCdps calculates fees for risky CDPs
+// The overall logic is first select the CDPs with 10% of the liquidation ratio
+// Then we call calculate fees on each of those CDPs
+// Next we store the result of the fees in the cdp.AccumulatedFees field
+// Finally we set the cdp.FeesUpdated time to the current block time (ctx.BlockTime()) since that
+// is when we made the update
+func (k Keeper) UpdateFeesForRiskyCdps(ctx sdk.Context, collateralDenom string, marketID string) sdk.Error {
+
+	// TODO - QUESTION - should this be modified to get different prices for different assets in future?
```
karzak:
```suggestion
	
```

Comment can be removed -`UpdateFeesForRiskyCdps` is called in a for-loop in the begin blocker for each asset (xrp, btc, etc), so it is already doing this.

x/cdp/spec/04_begin_block.md
```diff
@@ -13,6 +14,13 @@ At the start of every block the BeginBlocker of the cdp module:
 - An equal amount of debt coins are minted and sent to the system's CDP module account.
 - An equal amount of stable asset coins are minted and sent to the system's liquidator module account
 
+## Update risky cdps
+
+- UpdateFeesForRiskyCdps calculates fees for risky CDPs
+- Select the CDPs with 10% of the liquidation ratio - the risky CDPs
+- Calculate additional accumulated fees on each of those CDPs
+- Update the feeds updated time for the CDP to the current block time
```
karzak:
```suggestion
- Update the fees updated time for the CDP to the current block time
```

jmaheswaran:
Fixed typo and old comment

jmaheswaran:
All tests pass successfully. Should be ready to merge now.

@karzak

PR #386, opened 2020-03-04
Committee Gov Module

rhuairahrighairidh:
In case something going wrong, we need a fast governance mechanism. This module creates "committees" that can vote on governance proposals quickly to update params, shutdown msg types, etc.
Disabling msg types will be handled by a different module.

See `x/committee/spec/README.md` for an overview of how committees work.

Open Questions
- do we need more features from committees like 'no' or 'veto' votes?
- should committee IDs be strings rather than uints?
- I had to make the committee submit-proposal cli cmd work a a bit differently from how gov does it to avoid forking sdk code. We could replace gov's submit-proposal cmd so they work the same way and have the same UX.
- Is there enough tests?


x/committee/types/querier.go
```diff
@@ -0,0 +1,49 @@
+package types
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// Query endpoints supported by the Querier
+const (
+	//QueryParams     = "params"
```
denalimarsh, line 9:
```suggestion
```

x/committee/keeper/proposal.go
```diff
@@ -0,0 +1,162 @@
...
+		return 0, sdk.ErrUnauthorized("committee does not have permissions to enact proposal")
+	}
+
+	// Check proposal is valid
+	if err := k.ValidatePubProposal(ctx, pubProposal); err != nil {
+		return 0, err
+	}
+
+	// Get a new ID and store the proposal
+	deadline := ctx.BlockTime().Add(com.MaxProposalDuration)
```
denalimarsh, line 33: Each proposal before a committee is subject to constant duration `MaxProposalDuration`. Consider renaming to `ProposalDuration` since proposal duration times don't vary by committee.

x/committee/keeper/proposal_test.go
```diff
@@ -0,0 +1,324 @@
...
+			pubProposal: UnregisteredProposal{gov.TextProposal{Title: "A Title", Description: "A description of this proposal."}},
+			expectPass:  false,
+		},
+		{
+			name:        "invalid (nil)",
+			pubProposal: nil,
+			expectPass:  false,
+		},
+		{
+			name: "invalid (proposal handler fails)",
```
karzak, line 298: Add case for a valid param change proposal that passes

x/committee/keeper/keeper_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	// delete from store
+	suite.keeper.DeleteCommittee(suite.ctx, com.ID)
+
+	// check does not exist
+	_, found = suite.keeper.GetCommittee(suite.ctx, com.ID)
+	suite.False(found)
+}
+
+func (suite *KeeperTestSuite) TestGetSetProposal() {
```
karzak, line 64: hyper-nit: keep naming consistent 
```suggestion
func (suite *KeeperTestSuite) TestGetSetDeleteProposal() {
```

x/committee/keeper/keeper_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	// delete from store
+	suite.keeper.DeleteProposal(suite.ctx, prop.ID)
+
+	// check does not exist
+	_, found = suite.keeper.GetProposal(suite.ctx, prop.ID)
+	suite.False(found)
+}
+
+func (suite *KeeperTestSuite) TestGetSetVote() {
```
karzak, line 89: ditto:
```suggestion
func (suite *KeeperTestSuite) TestGetSetDeleteVote() {
```

x/committee/client/rest/query.go
```diff
@@ -0,0 +1,275 @@
...
+
+		// Query
+		res, err := common.QueryProposer(cliCtx, proposalID)
+		if err != nil {
+			rest.WriteErrorResponse(w, http.StatusInternalServerError, err.Error())
+			return
+		}
+
+		// Write response
+		rest.PostProcessResponse(w, cliCtx, res)
```
karzak, line 193:
```suggestion
		cliCtx = cliCtx.WithHeight(height)
		rest.PostProcessResponse(w, cliCtx, res)
```

x/committee/types/permissions.go
```diff
@@ -0,0 +1,82 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+// TODO add more permissions?
+// - limit parameter changes to be within small ranges
+// - allow community spend proposals
+// - allow committee change proposals
```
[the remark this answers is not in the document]
rhuairahrighairidh: Yeah committees could change themselves or other committees. Though currently it isn't enabled purely because setting up the keeper was a bit awkward (to create a keeper, you pass a gov proposal router, which contains a reference to the same committee keeper).
If we wanted this, we could create a permission to stop committees aribtrarily giving themselves more permissions

x/committee/client/cli/tx.go
```diff
@@ -0,0 +1,235 @@
...
+				},
+			},
+			sdk.MustNewDecFromStr("0.8"),
+			time.Hour*24*7,
+		),
+	)
+	exampleChangeProposalBz, err := cdc.MarshalJSONIndent(exampleChangeProposal, "", "  ")
+	if err != nil {
+		panic(err)
+	}
```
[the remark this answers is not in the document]
rhuairahrighairidh: The Long field value is evaluated when the cmd struct is created, which happens when all the commands are created in kvcli main.go. So if there's a problem it should panic when any command runs.
But it would nicer if problems were caught sooner, so I'll add a test.

x/committee/client/cli/cli_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/kava-labs/kava/x/committee/client/cli"
+)
+
+type CLITestSuite struct {
+	suite.Suite
+	cdc *codec.Codec
+}
+
+func (suite *CLITestSuite) SetupTest() {
+	ahpp := app.NewTestApp()
```
fedekunze, line 19: is this a typo? seems to be in purpose

x/committee/genesis.go
```diff
@@ -0,0 +1,58 @@
...
+
+	nextID, err := keeper.GetNextProposalID(ctx)
+	if err != nil {
+		panic(err)
+	}
+	committees := []types.Committee{}
+	keeper.IterateCommittees(ctx, func(com types.Committee) bool {
+		committees = append(committees, com)
+		return false
+	})
```
fedekunze, line 40: I'd create a `GetAllCommittees` funtion

x/committee/genesis.go
```diff
@@ -0,0 +1,58 @@
...
+	proposals := []types.Proposal{}
+	votes := []types.Vote{}
+	keeper.IterateProposals(ctx, func(p types.Proposal) bool {
```
fedekunze, line 43: ditto

x/committee/genesis.go
```diff
@@ -0,0 +1,58 @@
...
+		proposals = append(proposals, p)
+		keeper.IterateVotes(ctx, p.ID, func(v types.Vote) bool {
```
fedekunze, line 45: ditto

x/committee/handler.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		return err.Result()
+	}
+
+	// Enact a proposal if it has enough votes
+	passes, err := k.GetProposalResult(ctx, msg.ProposalID)
+	if err != nil {
+		return err.Result()
+	}
+	if passes {
```
fedekunze, line 66: use inverse for ease of reading: `if !passes { ... }`

x/committee/keeper/proposal.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+	return nil
+}
+
+// CloseExpiredProposals removes proposals (and associated votes) that have past their deadline.
+// TODO rename to RemoveExpiredProposals?
+func (k Keeper) CloseExpiredProposals(ctx sdk.Context) {
+
+	k.IterateProposals(ctx, func(proposal types.Proposal) bool {
+		if proposal.HasExpiredBy(ctx.BlockTime()) {
```
fedekunze, line 134: use inverse

x/committee/keeper/proposal.go
```diff
@@ -0,0 +1,197 @@
...
+func (k Keeper) DeleteProposalAndVotes(ctx sdk.Context, proposalID uint64) {
+	var votes []types.Vote
+	k.IterateVotes(ctx, proposalID, func(vote types.Vote) bool {
+		votes = append(votes, vote)
+		return false
+	})
+
+	k.DeleteProposal(ctx, proposalID)
+	for _, v := range votes {
+		k.DeleteVote(ctx, v.ProposalID, v.Voter)
```
fedekunze, line 195: you can delete the votes on the iterator above

x/committee/keeper/querier.go
```diff
@@ -0,0 +1,188 @@
...
+func queryProposals(ctx sdk.Context, path []string, req abci.RequestQuery, keeper Keeper) ([]byte, sdk.Error) {
+	var params types.QueryCommitteeParams
+	err := keeper.cdc.UnmarshalJSON(req.Data, &params)
+	if err != nil {
+		return nil, sdk.ErrUnknownRequest(sdk.AppendMsgToErr("incorrectly formatted request data", err.Error()))
+	}
+
+	proposals := []types.Proposal{}
+	keeper.IterateProposals(ctx, func(p types.Proposal) bool {
+		if p.CommitteeID == params.CommitteeID {
```
fedekunze, line 90: use inverse

x/committee/keeper/querier.go
```diff
@@ -0,0 +1,188 @@
...
+func queryVotes(ctx sdk.Context, path []string, req abci.RequestQuery, keeper Keeper) ([]byte, sdk.Error) {
+	var params types.QueryProposalParams
+	err := keeper.cdc.UnmarshalJSON(req.Data, &params)
+
+	if err != nil {
+		return nil, sdk.ErrUnknownRequest(sdk.AppendMsgToErr("incorrectly formatted request data", err.Error()))
+	}
+
+	votes := []types.Vote{}
+	keeper.IterateVotes(ctx, params.ProposalID, func(v types.Vote) bool {
```
fedekunze, line 135: create GetAllProposalVotes

x/committee/keeper/querier_test.go
```diff
@@ -0,0 +1,247 @@
...
+func (suite *QuerierTestSuite) TestQueryCommittees() {
+	ctx := suite.ctx.WithIsCheckTx(false)
+	// Set up request query
+	query := abci.RequestQuery{
+		Path: strings.Join([]string{custom, types.QuerierRoute, types.QueryCommittees}, "/"),
+	}
+
+	// Execute query and check the []byte result
+	bz, err := suite.querier(ctx, []string{types.QueryCommittees}, query)
+	suite.NoError(err)
```
fedekunze, line 99: we should be using `suite.Require().` everywhere because the default is an assertion.
```suggestion
	suite.Require().NoError(err)
```

x/committee/proposal_handler.go
```diff
@@ -0,0 +1,66 @@
...
+
+func handleCommitteeChangeProposal(ctx sdk.Context, k Keeper, committeeProposal CommitteeChangeProposal) sdk.Error {
+	if err := committeeProposal.ValidateBasic(); err != nil {
+		return ErrInvalidCommittee(DefaultCodespace, err.Error())
+	}
+
+	// Remove all committee's ongoing proposals
+	var proposals []Proposal
+	k.IterateProposals(ctx, func(p Proposal) bool {
+		if p.CommitteeID == committeeProposal.NewCommittee.ID {
```
fedekunze, line 33: use inverse

x/committee/proposal_handler.go
```diff
@@ -0,0 +1,66 @@
...
+			proposals = append(proposals, p)
+		}
+		return false
+	})
+	for _, p := range proposals { // split loops to avoid updating the db while iterating
+		k.DeleteProposalAndVotes(ctx, p.ID)
```
fedekunze, line 39: you can delete the proposal votes on the iterator above

x/committee/proposal_handler.go
```diff
@@ -0,0 +1,66 @@
...
+	// Remove all committee's ongoing proposals
+	var proposals []Proposal
+	k.IterateProposals(ctx, func(p Proposal) bool {
+		if p.CommitteeID == committeeProposal.CommitteeID {
+			proposals = append(proposals, p)
+		}
+		return false
+	})
+	for _, p := range proposals { // split loops to avoid updating the db while iterating
+		k.DeleteProposalAndVotes(ctx, p.ID)
```
fedekunze, line 61: ditto

x/committee/proposal_handler_test.go
```diff
@@ -0,0 +1,257 @@
...
+				NewCommitteeGenState(suite.app.Codec(), suite.testGenesis),
+			)
+			suite.ctx = suite.app.NewContext(true, abci.Header{Height: 1, Time: testTime})
+			handler := committee.NewProposalHandler(suite.keeper)
+
+			// get proposals and votes for target committee
+			var proposals []committee.Proposal
+			var votes []committee.Vote
+			suite.keeper.IterateProposals(suite.ctx, func(p committee.Proposal) bool {
+				if p.CommitteeID == tc.proposal.NewCommittee.ID {
```
fedekunze, line 140: inverse

x/committee/proposal_handler_test.go
```diff
@@ -0,0 +1,257 @@
...
+				NewCommitteeGenState(suite.app.Codec(), suite.testGenesis),
+			)
+			suite.ctx = suite.app.NewContext(true, abci.Header{Height: 1, Time: testTime})
+			handler := committee.NewProposalHandler(suite.keeper)
+
+			// get proposals and votes for target committee
+			var proposals []committee.Proposal
+			var votes []committee.Vote
+			suite.keeper.IterateProposals(suite.ctx, func(p committee.Proposal) bool {
+				if p.CommitteeID == tc.proposal.CommitteeID {
```
fedekunze, line 218: ditto

x/committee/types/codec.go
```diff
@@ -0,0 +1,57 @@
...
+// ModuleCdc generic sealed codec to be used throughout module
+var ModuleCdc *codec.Codec
+
+func init() {
+	cdc := codec.New()
+	RegisterModuleCodec(cdc)
+	ModuleCdc = cdc.Seal()
+}
+
+// TODO decide if not using gov's Content type would be better
```
fedekunze, line 19: do the proposals of this module need to be duplicated?

x/committee/client/cli/cli_test.go
rhuairahrighairidh: fixed

x/committee/keeper/proposal_test.go
```diff
@@ -0,0 +1,324 @@
...
+		{
+			name:      "not enough votes",
+			committee: normalCom,
+			votes: []types.Vote{
+				{ProposalID: defaultID, Voter: suite.addresses[0]},
+			},
+			proposalPasses: false,
+			expectPass:     true,
+		},
+	}
```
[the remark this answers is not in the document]
rhuairahrighairidh: Good idea, I added a handler test for this case. Turns out this only applies to community spend proposals.

x/committee/types/gov_proposal.go
```diff
@@ -0,0 +1,114 @@
...
+func (ccp CommitteeChangeProposal) ProposalType() string { return ProposalTypeCommitteeChange }
+
+// ValidateBasic runs basic stateless validity checks
+func (ccp CommitteeChangeProposal) ValidateBasic() sdk.Error {
+	if err := govtypes.ValidateAbstract(DefaultCodespace, ccp); err != nil {
+		return err
+	}
+	if err := ccp.NewCommittee.Validate(); err != nil {
+		return ErrInvalidCommittee(DefaultCodespace, err.Error())
+	}
```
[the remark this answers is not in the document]
rhuairahrighairidh: The committees use the id as their key in the store, so it should be impossible for two committees to end up in the store with the same IDs (as keys are unique).
I guess you could end up with two stored comittees with different key IDs, but the same ID field. But `SetCommittee` ensures the key and field are the same, so as long as committees aren't set by another method I think it's ok.

x/committee/handler.go
rhuairahrighairidh: I'm not sure how to apply your suggestion here

x/committee/genesis.go
rhuairahrighairidh: I created a load of these, now most code doesn't need to use the iterators.

x/committee/types/types.go
```diff
@@ -0,0 +1,132 @@
...
+
+	if len(c.Members) == 0 {
+		return fmt.Errorf("committee %d invalid: cannot have zero members", c.ID)
+	}
+
+	if len(c.Description) > MaxCommitteeDescriptionLength {
+		return fmt.Errorf("invalid description")
+	}
+
+	if c.VoteThreshold.IsNil() || c.VoteThreshold.IsNegative() || c.VoteThreshold.GT(sdk.NewDec(1)) {
```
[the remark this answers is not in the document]
rhuairahrighairidh: Yeah thats probably not a feature we want.

x/committee/keeper/proposal.go
```diff
@@ -0,0 +1,162 @@
...
+
+// EnactProposal makes the changes proposed in a proposal.
+func (k Keeper) EnactProposal(ctx sdk.Context, proposalID uint64) sdk.Error {
+	pr, found := k.GetProposal(ctx, proposalID)
+	if !found {
+		return types.ErrUnknownProposal(k.codespace, proposalID)
+	}
+
+	// Run the proposal's changes through the associated handler, but using a cached version of state to ensure changes are not permanent if an error occurs.
+	handler := k.router.GetRoute(pr.ProposalRoute())
```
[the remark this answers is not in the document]
rhuairahrighairidh: I beefed up the `ValidatePubProposal` method to deal with another problem so I've used that as the check method. One downside is now the proposal is run twice (once to check, then for real).

x/committee/handler.go
fedekunze:
```go
if !passes {
 ctx.EventManager().EmitEvent(
		sdk.NewEvent(
			sdk.EventTypeMessage,
			sdk.NewAttribute(sdk.AttributeKeyModule, types.AttributeValueCategory),
			sdk.NewAttribute(sdk.AttributeKeySender, msg.Voter.String()),
		),
	)
return sdk.Result{Events: ctx.EventManager().Events()}
}
err = k.EnactProposal(ctx, msg.ProposalID)
...
```

x/committee/keeper/proposal_test.go
```diff
@@ -0,0 +1,324 @@
...
+				[]params.ParamChange{{
+					Subspace: "non existant",
+					Key:      "non existant",
+					Value:    "nonsense",
+				}},
+			),
+			expectPass: false,
+		},
+		// Some proposals can cause the proposal handler to panic.
+		// However panics will be caught when the proposal is first submitted so should avoid making it onto the chain.
```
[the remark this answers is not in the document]
rhuairahrighairidh: currently there is not

x/committee/types/types.go
```diff
@@ -0,0 +1,132 @@
...
+		return fmt.Errorf("invalid threshold")
+	}
+
+	if c.MaxProposalDuration < 0 {
```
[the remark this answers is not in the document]
rhuairahrighairidh: I think for `ProposalDuration` exactly equal to zero, voters won't be able to vote because the blockTime must be < proposal.deadline (not ==). But could set the ProposalDuration to a tiny number to get the ability to vote within block.

x/committee/types/types.go
```diff
@@ -0,0 +1,132 @@
...
+import (
+	"fmt"
+	"time"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/x/gov"
+	"gopkg.in/yaml.v2"
+)
+
+const MaxCommitteeDescriptionLength int = 5000
```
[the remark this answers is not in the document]
rhuairahrighairidh: changed it 512, enough for a paragraph of text

x/committee/types/codec.go
rhuairahrighairidh: The committee proposal interface type is the same as the one gov uses for proposals. So any gov proposal also works as a committee proposal. Though they need to be registered on the gov module cdc to allow them to be marshalled in gov submit proposal msgs.

x/committee/keeper/querier_test.go
rhuairahrighairidh: I'll punt this to an issue for now as most of our modules aren't using `Require()`.

x/committee/types/permissions.go
```diff
@@ -0,0 +1,82 @@
+package types
+
+import (
+	"github.com/cosmos/cosmos-sdk/x/gov"
+	"github.com/cosmos/cosmos-sdk/x/params"
+)
+
+func init() {
```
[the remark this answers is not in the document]
rhuairahrighairidh: I'm going to leave it where it is for now, following the pattern in other modules where init functions only modify gov's codec not committee's codec.

x/committee/types/committee.go
```diff
@@ -0,0 +1,140 @@
...
+
+// ------------------------------------------
+//				Committees
+// ------------------------------------------
+
+// A Committee is a collection of addresses that are allowed to vote and enact any governance proposal that passes their permissions.
+type Committee struct {
+	ID               uint64           `json:"id" yaml:"id"`
+	Description      string           `json:"description" yaml:"description"`
+	Members          []sdk.AccAddress `json:"members" yaml:"members"`
```
fedekunze, line 22: q: how are members added? through regular param change proposals?

rhuairahrighairidh: The module has a couple of gov proposals that create/update/delete committees. So its up to the normal gov to add members, change permissions etc.
In theory these proposals could be submitted to committees, allowing them to change themselves or each other, but it's disabled currently.

PR #387, opened 2020-03-04
[R4R] Kava distribution module

karzak:
Implementation of `kavadist`, a module for minting kava into a new module account via a governance controlled inflation schedule.


x/kavadist/keeper/mint.go
```diff
@@ -0,0 +1,49 @@
...
+		return nil
+	}
+	timeElapsed := sdk.NewInt(ctx.BlockTime().Unix() - previousBlockTime.Unix())
+
+	for _, p := range params.Periods {
+		if p.Start.Before(ctx.BlockTime()) && (p.End.After(ctx.BlockTime()) || p.End.Equal(ctx.BlockTime())) {
+			totalSupply := k.supplyKeeper.GetSupply(ctx).GetTotal().AmountOf(types.GovDenom)
+			scalar := sdk.NewInt(1000000000000000000)
+			inflationInt := p.Inflation.Mul(sdk.NewDecFromInt(scalar)).TruncateInt()
+			accumulator := sdk.NewDecFromInt(cdptypes.RelativePow(inflationInt, timeElapsed, scalar)).Mul(sdk.SmallestDec())
```
rhuairahrighairidh: Is there a relation between the `1000000000000000000` scalar and the `sdk.SmallestDec`? They're just to convert back and forth between 18 decimal place `sdk.Dec` type right?
Nit: a comment or two might help understanding of the mathy bits.

x/kavadist/keeper/mint.go
```diff
@@ -0,0 +1,49 @@
...
+		)
+		return nil
+	}
+	previousBlockTime, found := k.GetPreviousBlockTime(ctx)
+	if !found {
+		previousBlockTime = ctx.BlockTime()
+		k.SetPreviousBlockTime(ctx, previousBlockTime)
+		return nil
+	}
+	timeElapsed := sdk.NewInt(ctx.BlockTime().Unix() - previousBlockTime.Unix())
```
rhuairahrighairidh: This doesn't take into account the period start/end times when calculating timeElapsed. So the newly minted coins for a period could be up to one block's worth different from the "ideal" amount. It could catch us out if the last blocktime is really long in the case of a chain halt. Probably worth a comment or fix.

karzak:
Ready for re-review. Updated to handle period transitions, added basic params validation tests.

PR #390, opened 2020-03-06
[R4R] Swagger Rest Automating Testing 

jmaheswaran:
Using `dredd` to perform automated testing of `rest` endpoints to the `kava` blockchain based on the `swagger.yaml` api definition/specification.


jmaheswaran:
Working on writing more hooks to create transactions and send to the blockchain so we can run tests on the other endpoints that require certain transactions to have been posted to the chain

jmaheswaran:
There are now 57 tests passing and 5 that are failing. Test results have been added to the pull request. Of the 5 remaining failing tests these are the remaining ones:

1) - `unbonding_delegations` for a delegator, validator pair. The test fails because there are currently no unbonding delegations on the chain. There need to be at least one for the test to pass in its current form. I tried sending a `NewDelegationMsg` to the chain which did successfully create a delegation, however not an unbonding delegation. I could not find any unbonding delegation message in the `staking` or `bank` or `gov` modules so am not sure how to create one over the rest api.

2) `/txs/{transaction hash}` - this test currently fails because it takes a transaction hash and sends it to the rest api. While we can create a transaction and post it to the chain using a dredd hook, I do not know how we can take the resulting transaction hash and pass it to dredd for use in testing the `yaml` file. The example variables are usually taken from the `yaml` file. However each time you start a new chain and post a transaction to it, you could end up with a different transaction hash so I am not sure how to statically include this in the `yaml`. If we can find a way to get the valid transaction hash and pass it to dredd the test should pass.

3) `POST` to `/txs` - this takes a valid signed transaction and sends it to the blockchain. It is failing because I do not know a way to get a consistent signature. The signature depends on the `sequence number`. So we need a way to get the `sequence number` from the chain and then pass it to dredd somehow. This runs into the same issue as the other failing `txs/{hash}` test where dredd usually reads static examples as tests from the `yaml` whereas the signature could be different every time.

4) `POST` to `/txs/encode` - the request format for this test now seems to be correct is a valid transaction. However the server throws an internal error - `  "error": "JSON encoding of interfaces require non-empty type field."` I am not sure why this is. In the requests there are valid labels (types) for each of the fields. I think maybe there JSON is being unmarshalled into a struct internally that has some extra fields that are then causing issues when the data is being `Amino` encoded, but am not too sure.

5)`POST` to `/txs/decode` - running into an issue similar to the above `encode` endpoint. We are getting an error `"error": "UnmarshalBinaryBare expected to read prefix bytes 282816A9 (since it is registered concrete) but got F0625DEE..."`. I am assuming if we can get the `encode` endpoint to work then we can pass that output to the `decode` endpoint and hopefully that would get this test to pass.

rhuairahrighairidh:
For 1), `MsgUndelegate` should do what you're after. You'll probably need to create a delegation first, then undelegate it.

For 2), the tx hash, looks like it can be calculated from just the tx object. I tried it out locally and I seemed to be getting the same tx hash for a transaction across chain restarts.
I think is how tendermint calculates it: 
```go
import "github.com/tendermint/tendermint/types"
/ get a StdTx from somewhere, eg:
stdTx := auth.StdTx{}
// marshal StdTx to bytes with amino
txbytes := cdc.MarshalBinaryLengthPrefixed(stdTx)
// cast to tendermint's Tx type and use it's hash function to get the hash
hash := types.Tx(txBytes).Hash()
```

For 3) Ideally the example tx in the swagger is the one used in the test (compared ot creating a new one in the hooks). It makes sure the example is valid.
If the tx is from an address that is never used elsewhere in the test, then the sequence number should be the same each run. The account number is static per run, and I believe just depends on the order accounts are added to genesis in startchain.sh

For 4) do you think you can catch the actual json being sent to the api from a hook? If we can see it then it'll probably be obvious what's wrong. (could also be bugs in the endpoint itself)

jmaheswaran:
1) worked - the `unbonding_delegation` test is now passing

jmaheswaran:
2) the `GET` to `/tx/{hash}` now works too

jmaheswaran:
4) and 5) are now fixed and working

jmaheswaran:
3) now fixed.

All the tests are now passing. Ready for review.

swagger-ui/hooks.go
```diff
@@ -0,0 +1,379 @@
+package main
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/snikch/goodman/hooks"
+	trans "github.com/snikch/goodman/transaction"
```
karzak: This requirement doesn't appear in `go.mod`.

swagger-ui/startchain.sh
```diff
@@ -0,0 +1,38 @@
...
+#   address: kavavaloper1ffv7nhd3z6sych2qpqkk03ec6hzkmufyz4scd0
+
+faucet="chief access utility giant burger winner jar false naive mobile often perfect advice village love enroll embark bacon under flock harbor render father since"
+# kava1xy7hrjy9r0algz9w3gzm8u6mrpq97kwta747gj
+
+# address: kava1ls82zzghsx0exkpr52m8vht5jqs3un0ceysshz
+# address: kavavaloper1ls82zzghsx0exkpr52m8vht5jqs3un0c5j2c04
+
+# Remove any existing data directory
+rm -rf ~/.kvd
```
karzak: Avoid deleting the default installation of kvd and kvcli by using the --home flag with `kvd init`

Something like 
`kvd init --chain-id --home /tmp`
I believe you have to add the home flag to all additional  kvd commands

testthing/test.go
```diff
@@ -0,0 +1,374 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	// helper methods for transactions
+	cdc := app.MakeCodec() // make codec for the app
+
+	// create a keybase
+	// TODO - IMPORTANT - this needs to be set manually and does NOT work with tilde i.e. ~/ does NOT work
+	keybase, err := keys.NewKeyBaseFromDir("/Users/john/.kvcli/")
```
karzak: You can obtain the homedir programmatically using `home, err := os.UserHomeDir()`

swagger-ui/startchain.sh
jmaheswaran: ok I will address

testthing/test.go
jmaheswaran: OK I will do that

swagger-ui/hooks.go
jmaheswaran: ok I will add to the go.mod file

testthing/test.go
jmaheswaran: done

swagger-ui/startchain.sh
jmaheswaran: ok it has been updated now to use a temporary directory instead of the default installation

jmaheswaran:
@karzak comments have been addressed

swagger-ui/hooks.go
```diff
@@ -0,0 +1,378 @@
...
+	keybase, err := keys.NewKeyBaseFromDir(home + "/kavatmp/")
+	if err != nil {
+		panic(err)
+	}
+
+	return keybase
+}
+
+// sendMsgToBlockchain sends a message to the blockchain via the rest api
+func sendMsgToBlockchain(cdc *amino.Codec, address string, keyname string,
```
karzak: I think this method should return the tx hash, and then for transactions that need to be included we should include a `awaitTxConfirmed()` method which does something like 
```
confrirmed = false
for !confirmed {
  resp, err := http.Get("https://localhost:1317/txs/{hash}")
  // sleep for 5 seconds if tx not found and poll again
 // when tx is found, set confirmed to true and return
}
```

I think this will solve the issue of having to run the tests twice.

swagger-ui/startchain.sh
```diff
@@ -0,0 +1,51 @@
...
+rm -rf ~/kavatmp
+
+# create validator key
+printf "$password\n$validatorMnemonic\n" | kvcli keys add vlad --recover --home ~/kavatmp
+# create faucet key
+printf "$password\n$faucet\n" | kvcli --home ~/kavatmp keys add faucet --recover --home ~/kavatmp
+
+
+# Create new data directory
+kvd --home ~/kavatmp init --chain-id=testing vlad # doesn't need to be the same as the validator
```
karzak:
```suggestion
{
kvd --home ~/kavatmp init --chain-id=testing vlad # doesn't need to be the same as the validator
} > /dev/null 2>&1
```

Avoids dumping the output to console

swagger-ui/startchain.sh
```diff
@@ -0,0 +1,51 @@
...
+kvcli --home ~/kavatmp config chain-id testing # or set trust-node true
+
+# add validator account to genesis
+kvd --home ~/kavatmp add-genesis-account $(kvcli --home ~/kavatmp keys show vlad -a) 10000000000000stake
+# add faucet account to genesis
+kvd --home ~/kavatmp add-genesis-account $(kvcli --home ~/kavatmp keys show faucet -a) 10000000000000stake,1000000000000xrp,100000000000btc
+
+# Create a delegation tx for the validator and add to genesis
+printf "$password\n" | kvd --home ~/kavatmp gentx --name vlad --home-client ~/kavatmp
+kvd --home ~/kavatmp collect-gentxs
```
karzak:
```suggestion
{
kvd --home ~/kavatmp collect-gentxs
} > /dev/null 2>&1
```
ditto

swagger-ui/testing-thoughts.md
```diff
@@ -0,0 +1,3 @@
+
```
karzak: delete this file

testthing/test.go
```diff
@@ -0,0 +1,349 @@
+////////////////////////////////////////////////////////////////////////////////
```
karzak: fine keeping this file, just change the folder name to `debugging_tools`

swagger-ui/how_to_run_dredd_tests.md
```diff
@@ -0,0 +1,20 @@
+Instructions on how to run the `dredd` tests
+
+1) Create the genesis state for the blockchain, start the blockchain and the rest server by running the following from the `swagger-ui` directory (`ctrc-c` to stop both the blockchain and the rest server):
```
karzak: Mention that dredd needs to be installed globally using np,

swagger-ui/testing-thoughts.md
jmaheswaran: 🆗

testthing/test.go
jmaheswaran: 🆗

swagger-ui/how_to_run_dredd_tests.md
jmaheswaran: 🆗

jmaheswaran:
Updated to a new workflow so that the tests now all pass on the first and every run of `dredd`.

Added the updated files and also updated the instructions.

jmaheswaran:
Incorporated code review feedback. 

Made all the setup and tests runs and shutdown and cleanup run from one script.

Added a command to the `Makefile` to run all the tests. 

Run `make test_dredd` from the `kava` folder to run all the setup and `dredd` tests in one command.

rest_test/debugging_tools/test.go
```diff
@@ -0,0 +1,363 @@
+////////////////////////////////////////////////////////////////////////////////
+////
+//// **IMPORTANT NOTE FOR CODE REVIEW**
+////
+//// This file is used for testing and debugging. Much of the code it contains
+//// is similar or the same to the `hooks.go` file.
+////
+////////////////////////////////////////////////////////////////////////////////
```
rhuairahrighairidh: Suggestion: name this file to clarify that it sets up the tests, rather than being purely a debugging tool. It'll help our future selves when we come back to update it and have forgotten how it all worked. Something like `setup/setup_test.go` ?

rest_test/test-results
```diff
@@ -0,0 +1,181 @@
+[32mpass[39m: GET (200) /node_info duration: 79ms
```
rhuairahrighairidh: Is there a reason to keep this file? If not I'd remove it to keep things tidy.

rest_test/hooks.go
```diff
@@ -0,0 +1,378 @@
+package main
```
rhuairahrighairidh: If we're not using hooks I think we should remove this file. It will always be in the git history if we want it back in the future.

rest_test/swagger.yaml
```diff
@@ -0,0 +1,2758 @@
+---
```
rhuairahrighairidh: It would be good to have only one copy of the `swagger.yml` in the repo. Then there's only one file to update in the future when we add rest endpoints. Also two files can get out of sync, making it unclear to third parties which one describes our current rest interface.
Maybe just use the one in swagger-ui for now and remove this one?

rest_test/run_all_tests.sh
```diff
@@ -0,0 +1,109 @@
+#! /bin/bash
```
rhuairahrighairidh: Is there a reason for keeping this file if we have `run_all_tests_from_make.sh`? if not I'd remove it to keep the folder tidy

rest_test/run_all_tests_from_make.sh
```diff
@@ -0,0 +1,109 @@
...
+
+# build the go setup test file
+rm -f rest_test/test
+go build rest_test/debugging_tools/test.go & showLoading "Building go test file, please wait"
+
+# Create new data directory
+{
+kvd --home ~/kavatmp init --chain-id=testing vlad # doesn't need to be the same as the validator
+} > /dev/null 2>&1
+kvcli --home ~/kavatmp config chain-id testing # or set trust-node true
```
rhuairahrighairidh: `kvd` and `kvcli` should have their own separate home folders. Doesn't look like it's actually causing problems here, but it could trigger weird bugs in the future if they're overwritting each other's files.

Couple of suggestions
- if you define a variable like `kvdHome=/some/path` then pass flags like `--home $kvdHome` it will be easier to change the folder location in the future.
- If you put the home folders in `/tmp`, then they won't clutter or overwrite anything in the use's home directory, and they'll be cleaned up automatically by the os.

rest_test/debugging_tools/test.go
jmaheswaran: 🆗 👌

rest_test/test-results
jmaheswaran: No reason. I will remove.

rest_test/hooks.go
jmaheswaran: 🆗

rest_test/swagger.yaml
jmaheswaran: 🆗

rest_test/run_all_tests.sh
jmaheswaran: No reason. I will remove.

rest_test/run_all_tests_from_make.sh
jmaheswaran: 🆗

jmaheswaran:
Incorporated code review feedback from @rhuairahrighairidh 

- separate data directories for `kvd` and `kvcli`
- use variables to store data directories, and store in `tmp` for easier cleanup
- clean up bash script and check for `dredd` errors
- remove old bash script and test results file
- remove extra `yaml` file and fix issue in the other `yaml` file
- remove `hooks` file
- rename other test file
- fixed issues in the bash script
- modified go code to take kvcli dir from command line and pass from bash script

jmaheswaran:
`make test_dredd` from `kava` directory to run all tests. The code review comments, modifications and suggestions have been incorporated.

jmaheswaran:
Ready to merge @karzak

PR #394, opened 2020-03-11
[R4R] fix: update genesis for guide

karzak:
Adds a required field for USDX savings. Removes gentxs because I think people will need to create them separately.


denalimarsh:
I successfully tested a similar genesis file locally yesterday, so these changes will work.

PR #397, opened 2020-03-13
R4R: BEP3 additional features + module test suite

denalimarsh:
Changes in this PR:

Store swaps for a week prior to deletion
 - implemented new store key AtomicSwapLongtermStorage
 - swaps closed and moved to longterm storage upon successful claim/refund
 - update BeginBlocker to delete closed swaps after 1 week

Implemented asset supply
 - incoming, current, outgoing supply that are modified by swaps
 - updated genesis
 - tests + integrated into swap tests

Module test suite: abci, handler, genesis tests

Updated code coverage:
```bash
- github.com/kava-labs/kava/x/bep3        0.372s  coverage: 70.3% of statements
- github.com/kava-labs/kava/x/bep3/keeper 0.560s  coverage: 88.0% of statements
- github.com/kava-labs/kava/x/bep3/types  0.133s  coverage: 42.9% of statements
```

This PR is **no longer blocked**. Incoming and outgoing swaps have been tested against bnbchain testnet via the bep3-deputy.


rhuairahrighairidh:
I went over the bep3 code in more detail. Looks pretty solid, couple of points of feedback:

Requests for change:

- in https://github.com/Kava-Labs/kava/blob/1d0def0a9f4f28d0db41758770c871e009cdfb35/x/bep3/keeper/swap.go#L34 we should use `ctx.BlockTime` rather than `tmtime.Now`. Time adds a source of indeterminism in module code can lead to chain forks/halts, where validators get different results on applying the same tx. For example if their clocks are out of sync by more than 15mins one some validators would accept this tx and others would reject it. It would also cause problems replaying txs at a later time.

Nits (up to you if you want to change them):

- The asset supply limit can be sort of bypassed because the asset supply is only increased for completed swaps. So if the supply was near the limit, and several swaps were submitted, each of them individually won't cross the limit, but once they all go through the limit will be crossed.

- The `rand.Int(_, max)` function generates random numbers up to but not including `max`. So I think you don't need to minus the 1 off 2^256, but check my maths there.

- `MsgCreateAtomicSwap.ValidateBasic` could check for negative value on the `int64` types. Looks like it's not a problem right now because negative values are caught by the keeper. But for example if `heightSpan` (in `keeper.CreateAtomicSwap`) could be negative then the type cast to uint64 in [`GetAtomicSwapByHeightKey`](https://github.com/Kava-Labs/kava/blob/1d0def0a9f4f28d0db41758770c871e009cdfb35/x/bep3/types/keys.go#L35) would convert the negative num to a really large positive one and the swap would never expire.

- Probably worth it to have a test for the `CreateAtomicSwap` keeper method that checks a valid swap was created, and checks for errors on invalid input.

x/bep3/keeper/keeper.go
```diff
@@ -132,15 +132,53 @@ func (k Keeper) IterateAtomicSwapsByBlock(ctx sdk.Context, inclusiveCutoffTime u
...
+// ------------------------------------------
+//		Atomic Swap Longterm Storage Index
+// ------------------------------------------
+
+// InsertIntoLongtermStorage adds a swap ID and deletion time into the longterm storage index.
+// Completed swaps are stored for 1 week.
+func (k Keeper) InsertIntoLongtermStorage(ctx sdk.Context, atomicSwap types.AtomicSwap) {
+	store := prefix.NewStore(ctx.KVStore(k.key), types.AtomicSwapLongtermStoragePrefix)
+	// Assuming block time of 7 seconds, there are 86,400 blocks in a week
+	store.Set(types.GetAtomicSwapByHeightKey(atomicSwap.ClosedBlock+int64(86400), atomicSwap.GetSwapID()), atomicSwap.GetSwapID())
```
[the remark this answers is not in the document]
denalimarsh: https://github.com/Kava-Labs/kava/pull/397/commits/8f4130888e2c6d784166f8018180d49c844c83ec

x/bep3/keeper/keeper.go
```diff
@@ -132,15 +132,53 @@ func (k Keeper) IterateAtomicSwapsByBlock(ctx sdk.Context, inclusiveCutoffTime u
...
+}
+
+// RemoveFromLongtermStorage removes a swap from the into the longterm storage index
+func (k Keeper) RemoveFromLongtermStorage(ctx sdk.Context, atomicSwap types.AtomicSwap) {
+	store := prefix.NewStore(ctx.KVStore(k.key), types.AtomicSwapLongtermStoragePrefix)
+	store.Delete(types.GetAtomicSwapByHeightKey(atomicSwap.ClosedBlock+int64(86400), atomicSwap.GetSwapID()))
```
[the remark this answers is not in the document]
denalimarsh: https://github.com/Kava-Labs/kava/pull/397/commits/8f4130888e2c6d784166f8018180d49c844c83ec

x/bep3/keeper/asset_test.go
```diff
@@ -52,15 +52,66 @@ func (suite *AssetTestSuite) SetupTest() {
...
+			args{
+				denom:        "xyz",
+				inSwapSupply: 10000,
+				assetSupply:  50000,
+			},
+			false,
+		},
+	}
+	for _, tc := range testCases {
+		_, _ = suite.keeper.LoadAssetSupply(suite.ctx, tc.args.denom)
```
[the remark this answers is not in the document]
denalimarsh: https://github.com/Kava-Labs/kava/pull/397/commits/adb98f8687b0e54fff2d319d756fcd53517cb746

x/bep3/abci_test.go
```diff
@@ -0,0 +1,237 @@
...
+			firstCtx:        suite.ctx,
+			action:          Refund,
+			secondCtx:       suite.ctx.WithBlockHeight(suite.ctx.BlockHeight() + types.DefaultLongtermStorageDuration),
+			expectInStorage: false,
+		},
+	}
+
+	for _, tc := range testCases {
+		// Reset keeper and run the initial begin blocker
+		suite.ResetKeeper()
```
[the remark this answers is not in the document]
denalimarsh: https://github.com/Kava-Labs/kava/pull/397/commits/adb98f8687b0e54fff2d319d756fcd53517cb746

x/bep3/genesis_test.go
```diff
@@ -0,0 +1,206 @@
...
+				gs := baseGenState()
+				gs.Params.SupportedAssets[1].Denom = "bnb"
+				return app.GenesisState{"bep3": bep3.ModuleCdc.MustMarshalJSON(gs)}
+			},
+			expectPass: false,
+		},
+	}
+
+	for _, tc := range testCases {
+		if tc.expectPass {
```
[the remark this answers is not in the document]
denalimarsh: https://github.com/Kava-Labs/kava/pull/397/commits/adb98f8687b0e54fff2d319d756fcd53517cb746

x/bep3/genesis.go
```diff
@@ -13,37 +13,68 @@ func InitGenesis(ctx sdk.Context, keeper Keeper, supplyKeeper types.SupplyKeeper
...
-	}
-
-	// Set each AtomicSwap and store total coin count
 	for _, swap := range gs.AtomicSwaps {
 		if swap.Validate() != nil {
 			panic(fmt.Sprintf("invalid swap %s", swap.GetSwapID()))
 		}
+
+		// Confirm that the asset is supported and active
+		err := keeper.ValidateLiveAsset(ctx, swap.Amount[0])
```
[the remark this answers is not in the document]
denalimarsh: https://github.com/Kava-Labs/kava/pull/397/commits/adb98f8687b0e54fff2d319d756fcd53517cb746

x/bep3/genesis_test.go
```diff
@@ -0,0 +1,206 @@
...
+			suite.NotPanics(func() {
```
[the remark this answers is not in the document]
denalimarsh: https://github.com/Kava-Labs/kava/pull/397/commits/adb98f8687b0e54fff2d319d756fcd53517cb746

x/bep3/types/keys.go
```diff
@@ -22,16 +22,20 @@ const (
...
+
 // Key prefixes
 var (
-	AtomicSwapKeyPrefix     = []byte{0x00} // prefix for keys that store AtomicSwaps
-	AtomicSwapByBlockPrefix = []byte{0x01} // prefix for keys of the AtomicSwapsByBlock index
-	AssetSupplyKeyPrefix    = []byte{0x02} // prefix for keys that store global asset supply counts
+	AtomicSwapKeyPrefix             = []byte{0x00} // prefix for keys that store AtomicSwaps
+	AtomicSwapByBlockPrefix         = []byte{0x01} // prefix for keys of the AtomicSwapsByBlock index
+	AssetSupplyKeyPrefix            = []byte{0x02} // prefix for keys that store global asset supply counts
+	AtomicSwapLongtermStoragePrefix = []byte{0x04} // prefix for keys of the AtomicSwapLongtermStorage index
```
karzak:
```suggestion
	AtomicSwapLongtermStoragePrefix = []byte{0x03} // prefix for keys of the AtomicSwapLongtermStorage index
```
nit: use sequential key prefixes

x/bep3/types/params.go
```diff
@@ -18,15 +18,15 @@ var (
 	AbsoluteMinimumBlockLock int64 = 50
 	DefaultMinBlockLock      int64 = 80
 	DefaultMaxBlockLock      int64 = 600
-	DefaultSupportedAssets         = AssetParams{AssetParam{Denom: "kava", CoinID: "459", Limit: sdk.NewInt(1), Active: false}}
+	DefaultSupportedAssets         = AssetParams{AssetParam{Denom: "kava", CoinID: 459, Limit: sdk.NewInt(1), Active: false}}
```
karzak: do we want denom `ukava` here?

x/bep3/types/swap.go
```diff
@@ -57,26 +80,41 @@ func (a AtomicSwap) Validate() error {
...
-		Timestamp:           timestamp,
-		Sender:              sender,
-		Recipient:           recipient,
-		SenderOtherChain:    senderOtherChain,
-		RecipientOtherChain: recipientOtherChain,
-		ClosedBlock:         closedBlock,
-		Status:              status,
-	}
+// String implements stringer
+func (a AtomicSwap) String() string {
```
karzak: Use raw string (`) instead of (") to avoid having to manually add new lines.  Same for `AssetSupply` string method. 
See https://github.com/Kava-Labs/kava/blob/42bd5e149f176c2031193bc5333f6c775f586cf8/x/cdp/types/cdp.go#L36 for example.

x/bep3/types/params.go
denalimarsh: This is just a placeholder, "kava"/"ukava" isn't used anywhere - just "bnb" is transfered back and forth. We should actually probably change this to "bnb", mark it as active, and update the limit to be the expected asset limit at launch.

x/bep3/types/swap.go
denalimarsh: After making the modifications the output result is:
```bash
"Atomic Swap:\n\t\t    ID:                       213FBD68B56B411CCB4D11BC1C36614105644D1DFE90581AE496BDA7EFD9AE68\n\t\t
  \   Status:                   Open\n\t\t    Amount:                   1111111bnb\n\t\t
  \   Random number hash:       095a454bb7c0dd88a766be0907e9ffcca470470d73a987150e49d7a707533129\n\t\t
  \   Expire height:            599\n\t\t    Timestamp:                1585154146\n\t\t
  \   Sender:                   kava1vry5lhegzlulehuutcr7nmdlmktw88awp0a39p\n\t\t
  \   Recipient:                kava1sl8glhaa9f9tep0d9h8gdcfmwcatghtdrfcd2x\n\t\t
  \   Sender other chain:       tbnb10uypsspvl6jlxcx5xse02pag39l8xpe7a3468h\n\t\t
  \   Recipient other chain:    tbnb1etjxt3t5r9899w3tf7dnktdue5ngu4ncdwckvt\n\t\t
  \   Closed block:             0\n\t\t    Cross chain:              true\n\t\t    Direction:
  \               Outgoing"
```

After updating line 148 of `client/cli/query.go` from `cdc.MustUnmarshalJSON(res, &atomicSwap).String()` to `cdc.MustUnmarshalJSON(res, &atomicSwap)`, the output result is:
```bash
swap: null
amount:
- denom: bnb
  amount: "1111111"
random_number_hash:
- 158
- 241
- 81
- 52
- 158
- 150
- 232
- 68
- 85
- 223
- 235
- 234
- 205
- 109
- 228
- 140
- 9
- 128
- 136
- 154
- 147
- 109
- 255
- 82
- 38
- 38
- 235
- 29
- 76
- 107
- 19
- 172
expire_height: 366
timestamp: 1585154354
sender: kava1sl8glhaa9f9tep0d9h8gdcfmwcatghtdrfcd2x
recipient: kava1sl8glhaa9f9tep0d9h8gdcfmwcatghtdrfcd2x
sender_other_chain: tbnb10uypsspvl6jlxcx5xse02pag39l8xpe7a3468h
recipient_other_chain: tbnb1etjxt3t5r9899w3tf7dnktdue5ngu4ncdwckvt
closed_block: 0
status: 1
cross_chain: true
direction: 1
```

For the 2nd option to work we'll need to return the string representation of the hex-encoded `random_number_hash` bytes, drop the `swap: null` interface from the results, and add an ID field. This would probably necessitate a new AugmentedAtomicSwap object. Worth noting that `status` and `direction` both have custom `UnmarshalJSON` methods that should return them as strings "Open" and "Incoming", but that isn't happening here.

I've already taken a stab at the string representation but I'm not sure how to get the `random_number_hash` string into the query results. The original string concatenation was a solution to get around this and the other requirements of an AugmentedAtomicSwap.

x/bep3/types/params.go
denalimarsh: https://github.com/Kava-Labs/kava/pull/397/commits/6eea7377ae5b59fa2b7a13b6acd998cd71d16c75

x/bep3/types/keys.go
denalimarsh: https://github.com/Kava-Labs/kava/pull/397/commits/6eea7377ae5b59fa2b7a13b6acd998cd71d16c75

x/bep3/types/swap.go
karzak: fine to leave it as is then.

rhuairahrighairidh: `cliCtx.PrintOutput(atomicSwap)` defaults to printing the yaml encoded version of atomicSwap. So if you created a `MarshalYaml()` method for the random number hash type (or the swap type) you can control the output.

PR #419, opened 2020-04-07
Add Auction Simulations

rhuairahrighairidh:
TODOs
- [x] operations
- [x] params
- [x] finish invariants
- [x] generate better random bid durations
- [ ] tidy up genesis by fixing problems below

problems:
- the sdk sims setup new accounts with only one coin denom, we want more denoms
- don't know how to update the module accounts and supply when creating a genesis state
- operations fail frequently because generation doesn't account for min bid increments


x/auction/simulation/genesis.go
```diff
@@ -2,21 +2,137 @@ package simulation
...
 
-	// TODO implement this fully
-	// - randomly generating the genesis params
-	// - overwriting with genesis provided to simulation
-	auctionGenesis := types.DefaultGenesisState()
+	auctionGenesis := types.NewGenesisState(
+		types.DefaultNextAuctionID,
+		types.NewParams(
+			GenMaxAuctionDuration(simState.Rand),
+			GenBidDuration(simState.Rand),
```
karzak, line 45: Should check that `BidDuration < MaxDuration`

x/auction/simulation/genesis.go
```diff
@@ -2,21 +2,170 @@ package simulation
...
+	maxTimePerBlock time.Duration = 10000 * time.Second
+
+	// Calculate the average block time
+	AverageBlockTime time.Duration = (maxTimePerBlock - minTimePerBlock) / 2
+	// MaxBidDuration is a crude way of ensuring that BidDuration ≤ MaxAuctionDuration for all generated params
+	MaxBidDuration time.Duration = AverageBlockTime * 50
+)
+
+func GenBidDuration(r *rand.Rand) time.Duration {
+	d, err := RandomPositiveDuration(r, 0, MaxBidDuration)
```
karzak, line 33: Outside the scope of this PR, but these values made me consider #426

x/auction/simulation/operations/msg.go
```diff
@@ -0,0 +1,180 @@
...
+			_, err := generateBidAmount(r, auc, acc, blockTime)
+			if err == ErrorNotEnoughCoins {
+				return false // keep searching
+			} else if err != nil {
+				panic(err) // raise errors
+			}
+			return true // found valid pair
+		})
+		if !found {
+			return noOpMsg, nil, nil // TODO op message?
```
karzak, line 64: I think we should log when this occurs - I would be curious to know the frequency and to inspect the state when it does. Maybe:
```suggestion
			return NewOperationMsgBasic(auction.ModuleName, "no-operation (could not find bidder)", "", false, nil)
```

x/auction/simulation/operations/msg.go
```diff
@@ -0,0 +1,180 @@
...
+func submitMsg(ctx sdk.Context, handler sdk.Handler, msg sdk.Msg) (ok bool) {
+	ctx, write := ctx.CacheContext()
+	ok = handler(ctx, msg).IsOK()
+	if ok {
+		write()
+	}
+	return ok
+}
+
+var ErrorNotEnoughCoins = errors.New("account doesn't have enough coins")
```
karzak, line 93: nit: move this to the top level var declaration block

x/auction/simulation/operations/msg.go
```diff
@@ -0,0 +1,180 @@
...
+	"time"
+
+	"github.com/cosmos/cosmos-sdk/baseapp"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/x/auth"
+	authexported "github.com/cosmos/cosmos-sdk/x/auth/exported"
+	"github.com/cosmos/cosmos-sdk/x/simulation"
+
+	"github.com/kava-labs/kava/x/auction"
+	"github.com/kava-labs/kava/x/auction/keeper"
```
karzak, line 17: nit: you could use only `auction` methods rather than import the keeper and types

x/auction/simulation/operations/msg.go
rhuairahrighairidh: As suggested, I renamed the OpMsg "name" to clarify the reason for there being no operation.
I also added the error log from submitting msgs to the "comment" field of the op msgs to aid in debugging.

PR #421, opened 2020-04-08
[R4R] bump SDK version to v0.38.3

fedekunze:


x/bep3/types/msg.go
```diff
@@ -91,49 +95,55 @@ func (msg MsgCreateAtomicSwap) GetSigners() []sdk.AccAddress {
...
-		return sdk.ErrInternal(fmt.Sprintf("expected income %s must be in valid format e.g. 10000ukava", msg.ExpectedIncome))
+		return fmt.Errorf("expected income %s must be in valid format e.g. 10000ukava", msg.ExpectedIncome)
 	}
 	if expectedIncomeCoins.IsAnyGT(msg.Amount) {
-		return sdk.ErrInternal(fmt.Sprintf("expected income %s cannot be greater than amount %s", msg.ExpectedIncome, msg.Amount.String()))
+		return fmt.Errorf("expected income %s cannot be greater than amount %s", msg.ExpectedIncome, msg.Amount.String())
 	}
 	if msg.HeightSpan <= 0 {
-		return sdk.ErrInternal("height span  must be positive")
+		return errors.New("height span  must be positive")
```
denalimarsh, line 146:
```suggestion
		return errors.New("height span must be positive")
```

x/bep3/types/msg.go
```diff
@@ -182,15 +192,18 @@ func (msg MsgClaimAtomicSwap) GetSigners() []sdk.AccAddress {
...
 
 // ValidateBasic validates the MsgClaimAtomicSwap
-func (msg MsgClaimAtomicSwap) ValidateBasic() sdk.Error {
+func (msg MsgClaimAtomicSwap) ValidateBasic() error {
+	if msg.From.Empty() {
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidAddress, "sender address cannot be empty")
+	}
 	if len(msg.From) != AddrByteCount {
-		return sdk.ErrInternal(fmt.Sprintf("the expected address length is %d, actual length is %d", AddrByteCount, len(msg.From)))
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "actual address lenght ≠ expected length (%d ≠ %d)", len(msg.From), AddrByteCount)
```
denalimarsh, line 200:
```suggestion
		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "actual address length ≠ expected length (%d ≠ %d)", len(msg.From), AddrByteCount)
```

x/bep3/types/params.go
```diff
@@ -102,45 +118,114 @@ func ParamKeyTable() params.KeyTable {
 // pairs of bep3 module's parameters.
 // nolint
 func (p *Params) ParamSetPairs() params.ParamSetPairs {
+	// TODO: Write validation functions
```
denalimarsh, line 121: Same as other comment - can we remove this TODO?

x/bep3/types/params.go
```diff
@@ -102,45 +118,114 @@ func ParamKeyTable() params.KeyTable {
...
+		return err
+	}
+
+	if err := validateMaxBlockLockParam(p.MaxBlockLock); err != nil {
+		return err
+	}
+
 	if p.MinBlockLock >= p.MaxBlockLock {
-		return fmt.Errorf("maximum block lock must be greater than minimum block lock")
+		return fmt.Errorf("minimum block lock ≥ maximum block lock, got %d ≥ %d", p.MinBlockLock, p.MaxBlockLock)
```
denalimarsh, line 145:
```suggestion
		return fmt.Errorf("minimum block lock cannot be ≥ maximum block lock, got %d ≥ %d", p.MinBlockLock, p.MaxBlockLock)
```

x/cdp/keeper/cdp.go
```diff
@@ -340,72 +341,73 @@ func (k Keeper) SetGovDenom(ctx sdk.Context, denom string) {
...
-		return types.ErrCollateralNotSupported(k.codespace, collateralDenom)
+		return sdkerrors.Wrap(types.ErrCollateralNotSupported, collateralDenom)
 	}
+
 	for _, dc := range principal {
 		totalPrincipal := k.GetTotalPrincipal(ctx, collateralDenom, dc.Denom).Add(dc.Amount)
 		collateralLimit := cp.DebtLimit.AmountOf(dc.Denom)
 		if totalPrincipal.GT(collateralLimit) {
-			return types.ErrExceedsDebtLimit(k.codespace, sdk.NewCoins(sdk.NewCoin(dc.Denom, totalPrincipal)), sdk.NewCoins(sdk.NewCoin(dc.Denom, collateralLimit)))
+			return sdkerrors.Wrapf(types.ErrExceedsDebtLimit, "payment %s > debt %s", sdk.NewCoins(sdk.NewCoin(dc.Denom, totalPrincipal)), sdk.NewCoins(sdk.NewCoin(dc.Denom, collateralLimit)))
```
denalimarsh, line 391: nit: Error message is a bit deceiving. The individual payment isn't greater than the outstanding debt limit, the payment puts the outstanding debt above the limit.

x/cdp/keeper/cdp.go
```diff
@@ -340,72 +341,73 @@ func (k Keeper) SetGovDenom(ctx sdk.Context, denom string) {
...
 		}
 		globalLimit := k.GetParams(ctx).GlobalDebtLimit.AmountOf(dc.Denom)
 		if totalPrincipal.GT(globalLimit) {
-			return types.ErrExceedsDebtLimit(k.codespace, sdk.NewCoins(sdk.NewCoin(dc.Denom, totalPrincipal)), sdk.NewCoins(sdk.NewCoin(dc.Denom, globalLimit)))
+			return sdkerrors.Wrapf(types.ErrExceedsDebtLimit, "payment %s > debt %s", sdk.NewCoin(dc.Denom, totalPrincipal), sdk.NewCoin(dc.Denom, globalLimit))
```
denalimarsh, line 395: ditto. Also might be good to differentiate between which limit (collateral vs. global) is being hit in the error message.

x/cdp/keeper/cdp.go
```diff
@@ -340,72 +341,73 @@ func (k Keeper) SetGovDenom(ctx sdk.Context, denom string) {
...
+func (k Keeper) ValidateCollateralizationRatio(ctx sdk.Context, collateral sdk.Coins, principal sdk.Coins, fees sdk.Coins) error {
 	//
 	collateralizationRatio, err := k.CalculateCollateralizationRatio(ctx, collateral, principal, fees)
 	if err != nil {
 		return err
 	}
 	liquidationRatio := k.getLiquidationRatio(ctx, collateral[0].Denom)
 	if collateralizationRatio.LT(liquidationRatio) {
-		return types.ErrInvalidCollateralRatio(k.codespace, collateral[0].Denom, collateralizationRatio, liquidationRatio)
+		return sdkerrors.Wrapf(types.ErrInvalidCollateralRatio, "owner %s, collateral ratio %s, liquidation ration %s", collateral[0].Denom, collateralizationRatio, liquidationRatio)
```
denalimarsh, line 410:
```suggestion
		return sdkerrors.Wrapf(types.ErrInvalidCollateralRatio, "owner %s, collateral ratio %s, liquidation ratio %s", collateral[0].Denom, collateralizationRatio, liquidationRatio)
```

x/cdp/keeper/deposit.go
```diff
@@ -40,45 +41,45 @@ func (k Keeper) DepositCollateral(ctx sdk.Context, owner sdk.AccAddress, deposit
...
+		return sdkerrors.Wrapf(types.ErrCdpNotFound, "owner %s, collateral %s", owner, collateral[0].Denom)
 	}
 	deposit, found := k.GetDeposit(ctx, cdp.ID, depositor)
 	if !found {
-		return types.ErrDepositNotFound(k.codespace, depositor, cdp.ID)
+		return sdkerrors.Wrapf(types.ErrDepositNotFound, "depositor %s, collateral %s", depositor, collateral[0].Denom)
 	}
 	if collateral.IsAnyGT(deposit.Amount) {
-		return types.ErrInvalidWithdrawAmount(k.codespace, collateral, deposit.Amount)
+		return sdkerrors.Wrapf(types.ErrInvalidWithdrawAmount, "owner %s, collateral %s", collateral, deposit.Amount)
```
denalimarsh, line 71: `owner` is out of place in this error message.

x/bep3/types/msg.go
```diff
@@ -36,21 +40,21 @@ var (
...
-	Timestamp           int64          `json:"timestamp"  yaml:"timestamp"`
-	Amount              sdk.Coins      `json:"amount"  yaml:"amount"`
-	ExpectedIncome      string         `json:"expected_income"  yaml:"expected_income"`
-	HeightSpan          int64          `json:"height_span"  yaml:"height_span"`
-	CrossChain          bool           `json:"cross_chain"  yaml:"cross_chain"`
+	From                sdk.AccAddress   `json:"from"  yaml:"from"`
+	To                  sdk.AccAddress   `json:"to"  yaml:"to"`
+	RecipientOtherChain string           `json:"recipient_other_chain"  yaml:"recipient_other_chain"`
+	SenderOtherChain    string           `json:"sender_other_chain"  yaml:"sender_other_chain"`
+	RandomNumberHash    tmbytes.HexBytes `json:"random_number_hash"  yaml:"random_number_hash"`
```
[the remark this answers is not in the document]
fedekunze: the `HexBytes` was moved to `tmbytes` package

x/auction/keeper/bidding_test.go
```diff
@@ -55,319 +56,347 @@ func TestAuctionBidding(t *testing.T) {
...
 		expectedEndTime time.Time
 		expectedBidder  sdk.AccAddress
 		expectedBid     sdk.Coin
-		expectpass      bool
+		expectPass      bool
+		expectPanic     bool
 	}{
 		{
 			"basic: auction doesn't exist",
 			auctionArgs{Surplus, "", c("token1", 1), c("token2", 1), sdk.Coin{}, []sdk.AccAddress{}, []sdk.Int{}},
```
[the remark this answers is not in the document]
denalimarsh: Related to this [comment](https://github.com/Kava-Labs/kava/pull/421/files/#r410400213)
> Should attempting to bid on an auction that doesn't exist cause a panic? Perhaps @rhuairahrighairidh can weigh in as he wrote the auction module.

x/auction/keeper/auctions_test.go
```diff
@@ -267,30 +267,31 @@ func TestStartSurplusAuction(t *testing.T) {
...
 			args{cdp.LiquidatorMacc, c("stable", 10), "gov"},
-			true,
+			true, false,
 		},
 		{
 			"no module account",
 			someTime,
 			args{"nonExistentModule", c("stable", 10), "gov"},
-			false,
+			false, true,
```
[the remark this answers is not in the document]
fedekunze: yeah this is independent of this issue and can be done separately

x/cdp/keeper/cdp.go
```diff
@@ -340,72 +341,73 @@ func (k Keeper) SetGovDenom(ctx sdk.Context, denom string) {
...
+func (k Keeper) ValidatePrincipalAdd(ctx sdk.Context, principal sdk.Coins) error {
 	for _, dc := range principal {
 		dp, found := k.GetDebtParam(ctx, dc.Denom)
 		if !found {
-			return types.ErrDebtNotSupported(k.codespace, dc.Denom)
+			return sdkerrors.Wrap(types.ErrDebtNotSupported, dc.Denom)
 		}
 		if dc.Amount.LT(dp.DebtFloor) {
-			return types.ErrBelowDebtFloor(k.codespace, sdk.NewCoins(dc), dp.DebtFloor)
+			return sdkerrors.Wrapf(types.ErrBelowDebtFloor, "proposed %s < minimum %s", sdk.NewCoins(dc), dp.DebtFloor)
```
rhuairahrighairidh, line 363:
```suggestion
			return sdkerrors.Wrapf(types.ErrBelowDebtFloor, "proposed %s < minimum %s", sdk.Coins{dc}, dp.DebtFloor)
```
While we're changing the errors do you want to remove `NewCoins` from within error messages? If `dc` is invalid in some way (such < 0) then `NewCoins` will panic. The simulator caught this in auction.

x/auction/simulation/params.go
```diff
@@ -17,29 +17,29 @@ func ParamChanges(r *rand.Rand) []simulation.ParamChange {
 	// as strings in JSON (such as time.Duration) have the escaped quotes.
 	// TODO should we encode the values properly with ModuleCdc.MustMarshalJSON()?
 	return []simulation.ParamChange{
-		simulation.NewSimParamChange(types.ModuleName, string(types.KeyBidDuration), "",
+		simulation.NewSimParamChange(types.ModuleName, string(types.KeyBidDuration),
 			func(r *rand.Rand) string {
-				return fmt.Sprintf("\"%d\"", GenBidDuration(r))
+				return fmt.Sprintf("%d", GenBidDuration(r))
```
rhuairahrighairidh, line 22: Should these have escaped quotes?
My understandig was that param changes are stored as json encoded param values. ie time.Duration encodes to a string.

x/auction/simulation/operations.go
```diff
@@ -60,41 +83,56 @@ func SimulateMsgPlaceBid(authKeeper auth.AccountKeeper, keeper auction.Keeper) s
...
+			helpers.DefaultGenTxGas,
+			chainID,
+			[]uint64{bidderAcc.GetAccountNumber()},
+			[]uint64{bidderAcc.GetSequence()},
+			bidder.PrivKey,
+		)
+
+		_, result, err := app.Deliver(tx)
+		if err != nil {
+			return simulation.NoOpMsg(types.ModuleName), nil, err
```
rhuairahrighairidh, line 121: Will returning an error here stop the simulation?

x/cdp/keeper/cdp.go
fedekunze: I'd propose instead: `"proposed %s < minimum %s", dc, dp.DebtFloor)` since it's the same

rhuairahrighairidh: ah yeah, thats nicer

x/kavadist/types/params.go
```diff
@@ -90,22 +90,54 @@ func ParamKeyTable() params.KeyTable {
...
+
+func validatePeriodsParams(i interface{}) error {
+	periods, ok := i.(Periods)
+	if !ok {
+		return fmt.Errorf("invalid parameter type: %T", i)
+	}
+
 	prevEnd := tmtime.Canonical(time.Unix(0, 0))
-	for _, pr := range p.Periods {
+	now := time.Now()
```
rhuairahrighairidh, line 123: I think this could cause validators to disagree on the hash of a block if they have different clock times, and a param change proposal is accepted with a period with an End time that sits in between the validators' clock times.
I'd remove the `time.Now()` call.

x/auction/simulation/operations.go
fedekunze: correct 👍

x/auction/simulation/params.go
fedekunze: good point. Idk tbh, maybe it's the same? The SDK uses both from what I'm seeing

x/bep3/types/msg.go
```diff
@@ -91,49 +95,55 @@ func (msg MsgCreateAtomicSwap) GetSigners() []sdk.AccAddress {
...
+		return fmt.Errorf("the length of random number hash should be %d", RandomNumberHashLength)
 	}
 	if msg.Timestamp <= 0 {
-		return sdk.ErrInternal("timestamp must be positive")
+		return errors.New("timestamp must be positive")
 	}
-	if !msg.Amount.IsAllPositive() {
-		return sdk.ErrInternal(fmt.Sprintf("the swapped out coin must be positive"))
+	if !msg.Amount.IsValid() {
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidCoins, msg.Amount.String())
```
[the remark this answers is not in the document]
fedekunze: good catch. Added that check

x/bep3/keeper/swap.go
```diff
@@ -7,40 +7,37 @@ import (
...
 	// Unix timestamp must be in range [-15 mins, 30 mins] of the current time
 	pastTimestampLimit := ctx.BlockTime().Add(time.Duration(-15) * time.Minute).Unix()
 	futureTimestampLimit := ctx.BlockTime().Add(time.Duration(30) * time.Minute).Unix()
 	if timestamp < pastTimestampLimit || timestamp >= futureTimestampLimit {
-		return types.ErrInvalidTimestamp(k.codespace)
-	}
-
-	// Sanity check on recipient address
-	if recipient.Empty() {
-		return sdk.ErrInvalidAddress("invalid (empty) recipient address")
```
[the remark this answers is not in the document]
fedekunze: I moved it to `msg.ValidateBasic`

x/validator-vesting/types/validator_vesting_account.go
```diff
@@ -199,14 +201,102 @@ func (vva ValidatorVestingAccount) Validate() error {
...
+	ValidatorAddress       sdk.ConsAddress       `json:"validator_address" yaml:"validator_address"`
+	ReturnAddress          sdk.AccAddress        `json:"return_address" yaml:"return_address"`
+	SigningThreshold       int64                 `json:"signing_threshold" yaml:"signing_threshold"`
+	CurrentPeriodProgress  CurrentPeriodProgress `json:"current_period_progress" yaml:"current_period_progress"`
+	VestingPeriodProgress  []VestingProgress     `json:"vesting_period_progress" yaml:"vesting_period_progress"`
+	DebtAfterFailedVesting sdk.Coins             `json:"debt_after_failed_vesting" yaml:"debt_after_failed_vesting"`
+}
+
+// MarshalJSON returns the JSON representation of a PeriodicVestingAccount.
+func (vva ValidatorVestingAccount) MarshalJSON() ([]byte, error) {
```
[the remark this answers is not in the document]
fedekunze: I'll create an issue

x/auction/simulation/params.go
rhuairahrighairidh: After some digging, I think we should do this:
```suggestion
				return string(types.ModuleCdc.MustMarshalJSON(GenBidDuration(r)))
```
When these generated values (string types) are set in the params keeper (through the param gov handler) they're type cast to `[]byte` then unmarshalled from json (using amino) ,  then validated, then marshalled back to json bytes and then set in the param store.

So I think the string formatting is effectively marshalling the types to json, which sounds like a bad idea. I propose replacing it with the proper json marshaller.

I looked at v0.38 sdk modules. Whether they use quotes or not mirrors amino's decision to use quotes when marshalling to json. Which itself seems a bit arbitrary: uint16 - no quotes, uint64 - quotes. It's also inconsistent with golang's encoding/json library (eg marshalling to json with encoding/json then unmarhaling with amino can produce incorrect values)

I'll create an issue

PR #427, opened 2020-04-12
[R4R] Remove non-determinism from bep3 sims

karzak:


x/bep3/simulation/genesis.go
```diff
@@ -53,7 +52,6 @@ func GenMaxBlockLock(r *rand.Rand, minBlockLock int64) int64 {
 func GenSupportedAssets(r *rand.Rand) types.AssetParams {
 	var assets types.AssetParams
 	for i := 0; i < (r.Intn(10) + 1); i++ {
```
rhuairahrighairidh, line 54: nit: this line will generate a new random number every iteration of the loop as the `r.Intn(10)` expression is evaluated each time
I think the end result of iterating some random number of times is the same, but might be cleaner to pick the random number outside the loop.

denalimarsh:
utACK

PR #428, opened 2020-04-12
[R4R] Pricefeed cleanup

### LINKED ISSUE #416: Pricefeed code cleanup
- [ ] Use byte prefix instead of string in `key.go`
- [ ] Remove unused key prefixes
- [ ] Define types for arrays rather than using `[]type` 
- [ ] Use `NewType()` functions instead of direct type instantiation (in `keeper.go` and `markets.go`)
- [ ] Rename `SetCurrentPrices` to `SetCurrentPrice` (this affects other modules, so ideally don't PR this until there aren't a lot of outstanding PRs)

karzak:
Closes #416


x/pricefeed/keeper/keeper.go
```diff
@@ -194,10 +183,10 @@ func (k Keeper) GetCurrentPrice(ctx sdk.Context, marketID string) (types.Current
...
 
 // GetRawPrices fetches the set of all prices posted by oracles for an asset
-func (k Keeper) GetRawPrices(ctx sdk.Context, marketID string) []types.PostedPrice {
+func (k Keeper) GetRawPrices(ctx sdk.Context, marketID string) types.PostedPrices {
 	store := ctx.KVStore(k.key)
-	bz := store.Get([]byte(types.RawPriceFeedPrefix + marketID))
-	var prices []types.PostedPrice
+	bz := store.Get(types.RawPriceKey(marketID))
+	var prices types.PostedPrices
 	k.cdc.MustUnmarshalBinaryBare(bz, &prices)
```
fedekunze, line 190: nit, this will panic if it fails. I'd use `UnmarshalBinaryBare` instead

x/pricefeed/types/market.go
```diff
@@ -68,6 +89,15 @@ Price: %s
 Expiry: %s`, pp.MarketID, pp.OracleAddress, pp.Price, pp.Expiry))
```
fedekunze, line 89: I'd use yaml for strings:

```go
func (pp PostedPrice) String() string {
	out, _ := yaml.Marshal(pp)
	return string(out)
}
```

fedekunze: same for other types

x/pricefeed/keeper/keeper.go
fedekunze: also check for `len(bz) == 0` in case the PostedPrices are not found

PR #429, opened 2020-04-13
[R4R] Incentives tests

karzak:


x/incentive/keeper/rewards_test.go
```diff
@@ -0,0 +1,341 @@
...
+	reward := types.NewReward(true, "bnb", c("ukava", 1000000000), time.Hour*7*24, time.Hour*24*365, time.Hour*7*24)
+	suite.NotPanics(func() {
+		suite.keeper.CreateNewRewardPeriod(suite.ctx, reward)
+	})
+	_, found = suite.keeper.GetRewardPeriod(suite.ctx, "bnb")
+	suite.True(found)
+
+}
+
+func (suite *KeeperTestSuite) TestCreateAndDeleteRewardsPeriods() {
```
denalimarsh, line 164: While there are only 2 cases here it should still use table tests. We think that table tests should be used for all tests, even if they only have 1 case, as it itemizes the logic being tested and enables future extensibility.

x/incentive/keeper/rewards_test.go
```diff
@@ -0,0 +1,341 @@
...
+
+	_, found := suite.keeper.GetRewardPeriod(suite.ctx, "bnb")
+	suite.True(found)
+	_, found = suite.keeper.GetRewardPeriod(suite.ctx, "xrp")
+	suite.False(found)
+
+}
+
+func (suite *KeeperTestSuite) TestApplyRewardsToCdps() {
+	suite.setupCdpChain()
```
denalimarsh, line 182: `setupCdpChain()` is only called once, but I understand why it's been moved outside of the test since it's a large function. The downside is that it's not very clear what state is - for example, is the expectation of 3 claims in state on line 192 a valid expectation?

denalimarsh: Not sure there's an easy solution here. What about moving `setupCdpChain()` to keeper_test.go and using it as the default test setup function for all/most methods? Then state expectations will be the same from test to test.

x/incentive/keeper/rewards_test.go
```diff
@@ -0,0 +1,341 @@
...
+	suite.NotPanics(func() {
+		suite.keeper.ApplyRewardsToCdps(suite.ctx)
+	})
+	_, found = suite.keeper.GetClaimPeriod(suite.ctx, 1, "bnb")
+	suite.True(found)
+	testID := suite.keeper.GetNextClaimPeriodID(suite.ctx, "bnb")
+	suite.Equal(uint64(2), testID)
+}
+
+func (suite *KeeperTestSuite) addObjectsToStore() {
```
denalimarsh, line 207: Shared functionality that can be moved to keeper_test.go as it's called from both rewards_test.go and querier_test.go.

x/incentive/types/params_test.go
```diff
@@ -21,129 +22,130 @@ type ParamTestSuite struct {
...
+						Denom:         "bnb",
+						Reward:        sdk.NewCoin("ukava", sdk.NewInt(10000000000)),
+						Duration:      time.Hour * 24 * 7,
+						TimeLock:      time.Hour * 8766,
+						ClaimDuration: time.Hour * 24 * 14,
+					},
+				},
+			},
 			expectPass: true,
 		},
```
denalimarsh, line 42: Add case for `Active: false`

x/incentive/types/params_test.go
```diff
@@ -21,129 +22,130 @@ type ParamTestSuite struct {
...
 		paramTest{
-			params:     p5,
+			name: "zero claim duration",
+			params: types.Params{
+				Active: true,
+				Rewards: types.Rewards{
+					types.Reward{
+						Active:        true,
+						Denom:         "bnb",
+						Reward:        sdk.NewCoin("ukava", sdk.NewInt(10000000000)),
```
denalimarsh, line 110: Since sdk.NewCoin() panics on an empty denom, could a Reward with `sdk.NewCoin("", sdk.NewInt(10000000000))` be submitted in params via governance? I believe proposals are json - so if the coin is first instantiated within param validation could it cause a panic if `len(denom) > 0` isn't checked?

x/incentive/types/params_test.go
```diff
@@ -21,129 +22,130 @@ type ParamTestSuite struct {
...
+			params: types.Params{
+				Active: true,
+				Rewards: types.Rewards{
+					types.Reward{
+						Active:        true,
+						Denom:         "bnb",
+						Reward:        sdk.NewCoin("ukava", sdk.NewInt(0)),
+						Duration:      time.Hour * 24 * 7,
+						TimeLock:      time.Hour * 8766,
+						ClaimDuration: time.Hour * 0,
```
denalimarsh, line 130:
```suggestion
						ClaimDuration: time.Hour * 24 * 14,
```

Test case will fail regardless of Reward's amount due to duplicated invalid ClaimDuration.

x/incentive/types/params_test.go
```diff
@@ -21,129 +22,130 @@ type ParamTestSuite struct {
...
 		},
 		paramTest{
-			params:     p6,
+			name: "zero reward",
+			params: types.Params{
+				Active: true,
+				Rewards: types.Rewards{
+					types.Reward{
+						Active:        true,
+						Denom:         "bnb",
```
denalimarsh, line 126: Denom length isn't checked, so params accepts `""` as a valid denom here.

x/incentive/keeper/querier_test.go
```diff
@@ -0,0 +1,32 @@
...
+	claimQueryParams := types.NewQueryClaimsParams(addr, "bnb")
+	query := abci.RequestQuery{
+		Path: strings.Join([]string{"custom", types.QuerierRoute, types.QueryGetClaims}, "/"),
+		Data: types.ModuleCdc.MustMarshalJSON(claimQueryParams),
+	}
+	bz, err = querier(suite.ctx, []string{types.QueryGetClaims}, query)
+
+	var claims types.Claims
+	suite.Nil(types.ModuleCdc.UnmarshalJSON(bz, &claims))
+	suite.Equal(1, len(claims))
```
denalimarsh, line 31: Add deep equals e.g. `suite.Equal(claimObj, returnedClaimObj)`

x/incentive/keeper/payout_test.go
```diff
@@ -0,0 +1,311 @@
...
+	supplyExported "github.com/cosmos/cosmos-sdk/x/supply/exported"
+	"github.com/kava-labs/kava/app"
+	"github.com/kava-labs/kava/x/cdp"
+	"github.com/kava-labs/kava/x/incentive/types"
+	"github.com/kava-labs/kava/x/kavadist"
+	validatorvesting "github.com/kava-labs/kava/x/validator-vesting"
+	abci "github.com/tendermint/tendermint/abci/types"
+)
+
+func (suite *KeeperTestSuite) TestSendCoinsToPeriodicVestingAccount() {
```
denalimarsh, line 20: Refactor to table test by generalizing the following code block
```
        suite.True(ok)
	expectedPeriods = append(expectedPeriods, vesting.Period{1, cs(c("ukava", 100))})
	suite.Equal(expectedPeriods, vacc.VestingPeriods)
	suite.Equal(cs(c("ukava", 600)), vacc.OriginalVesting)
	suite.Equal(cs(c("ukava", 600)), vacc.Coins)
	suite.Equal(int64(117), vacc.EndTime)
```

x/incentive/keeper/payout_test.go
```diff
@@ -0,0 +1,311 @@
...
+	suite.setupChain()
+	err := suite.keeper.SendTimeLockedCoinsToAccount(suite.ctx, kavadist.ModuleName, suite.addrs[2], cs(c("ukava", 100)), 5)
+	suite.Equal(types.CodeInvalidAccountType, err.Result().Code)
+	macc := suite.getModuleAccount(cdp.ModuleName)
+	err = suite.keeper.SendTimeLockedCoinsToAccount(suite.ctx, kavadist.ModuleName, macc.GetAddress(), cs(c("ukava", 100)), 5)
+	suite.Equal(types.CodeInvalidAccountType, err.Result().Code)
+}
+
+func (suite *KeeperTestSuite) TestPayoutClaim() {
+	suite.setupClaims()
```
denalimarsh, line 144: Could put setup here since `suite.setupClaims()` is only called once.

x/incentive/keeper/payout_test.go
```diff
@@ -0,0 +1,311 @@
...
+func (suite *KeeperTestSuite) TestSendCoinsToInvalidAccount() {
+	suite.setupChain()
+	err := suite.keeper.SendTimeLockedCoinsToAccount(suite.ctx, kavadist.ModuleName, suite.addrs[2], cs(c("ukava", 100)), 5)
+	suite.Equal(types.CodeInvalidAccountType, err.Result().Code)
+	macc := suite.getModuleAccount(cdp.ModuleName)
+	err = suite.keeper.SendTimeLockedCoinsToAccount(suite.ctx, kavadist.ModuleName, macc.GetAddress(), cs(c("ukava", 100)), 5)
+	suite.Equal(types.CodeInvalidAccountType, err.Result().Code)
+}
+
+func (suite *KeeperTestSuite) TestPayoutClaim() {
```
denalimarsh, line 143: Hard to understand the intent of this test without comments.

x/incentive/keeper/payout_test.go
```diff
@@ -0,0 +1,311 @@
...
+	suite.Equal(cs(c("ukava", 500)), vacc.OriginalVesting)
+
+	err = suite.keeper.PayoutClaim(suite.ctx, suite.addrs[1], "bnb", 1)
+	suite.NoError(err)
+	acc = suite.getAccount(suite.addrs[1])
+	vacc, ok = acc.(*vesting.PeriodicVestingAccount)
+	suite.True(ok)
+	suite.Equal(cs(c("ukava", 100)), vacc.OriginalVesting)
+
+	err = suite.keeper.PayoutClaim(suite.ctx, suite.addrs[3], "bnb", 1)
```
denalimarsh, line 159: It's not clear why the claim shouldn't be found here.

x/incentive/keeper/payout_test.go
```diff
@@ -0,0 +1,311 @@
...
+	_, found = suite.keeper.GetClaim(suite.ctx, suite.addrs[0], "bnb", 1)
+	suite.False(found)
+	_, found = suite.keeper.GetClaimPeriod(suite.ctx, 1, "xrp")
+	suite.True(found)
+	_, found = suite.keeper.GetClaim(suite.ctx, suite.addrs[0], "xrp", 1)
+	suite.True(found)
+
+}
+
+func (suite *KeeperTestSuite) setupChain() {
```
denalimarsh, line 189: Move to top of file or refactor to keeper_test.go.

x/incentive/keeper/payout_test.go
```diff
@@ -0,0 +1,311 @@
...
+	suite.keeper = tApp.GetIncentiveKeeper()
+	suite.ctx = ctx
+	suite.addrs = addrs
+}
+
+func (suite *KeeperTestSuite) setupClaims() {
+	suite.setupChain()
+	cp1 := types.NewClaimPeriod("bnb", 1, suite.ctx.BlockTime().Add(time.Hour*168), time.Hour*8766)
+	suite.keeper.SetClaimPeriod(suite.ctx, cp1)
+	c1 := types.NewClaim(suite.addrs[0], c("ukava", 100), "bnb", 1)
```
denalimarsh, line 241: Changing 'ukava' to 'xyzxyz' causes payout to fail in `TestPayoutClaim()`.

x/incentive/keeper/payout_test.go
```diff
@@ -0,0 +1,311 @@
...
+	ak := suite.app.GetAccountKeeper()
+	return ak.GetAccount(suite.ctx, addr)
+}
+
+func (suite *KeeperTestSuite) getModuleAccount(name string) supplyExported.ModuleAccountI {
+	sk := suite.app.GetSupplyKeeper()
+	return sk.GetModuleAccount(suite.ctx, name)
+}
+
+func (suite *KeeperTestSuite) setupExpiredClaims() {
```
denalimarsh, line 259: Since the BeginBlocker is used to expire the valid claims in `TestDeleteExpiredClaimPeriods()` I believe we can remove this function and just use `setupClaims()`.

x/incentive/keeper/payout_test.go
```diff
@@ -0,0 +1,311 @@
...
+	c1 := types.NewClaim(addrs[0], c("ukava", 1000000), "bnb", 1)
+	c2 := types.NewClaim(addrs[0], c("ukava", 1000000), "xrp", 1)
+	suite.keeper.SetClaim(ctx, c1)
+	suite.keeper.SetClaim(ctx, c2)
+	suite.app = tApp
+	suite.ctx = ctx
+	suite.addrs = addrs
+}
+
+func stringPva(pva *vesting.PeriodicVestingAccount) string {
```
denalimarsh, line 288: Delete as this is unused. If `PeriodicVestingAccount.String()` isn't available in a future sdk version then should submit a PR to cosmos-sdk with this function.

x/incentive/keeper/payout_test.go
```diff
@@ -0,0 +1,311 @@
...
+	AccountNumber: %d
+	Sequence: %d`,
+		pva.GetAddress(), pva.GetCoins(), pva.GetOriginalVesting(), pva.StartTime, pva.EndTime, (pva.VestingPeriods), pva.GetAccountNumber(), pva.GetSequence())
+}
+
+func stringPeriod(p vesting.Period) string {
+	return fmt.Sprintf("\t\tLength: %d, Amount: %s", p.Length, p.Amount)
+}
+
+func stringPeriods(periods vesting.Periods) string {
```
denalimarsh, line 305: Same here

x/incentive/handler_test.go
```diff
@@ -0,0 +1,73 @@
...
+	})
+	c1 := incentive.NewClaim(suite.addrs[0], c("ukava", 1000000), "bnb", 1)
+	suite.NotPanics(func() {
+		suite.keeper.SetClaim(suite.ctx, c1)
+	})
+}
+
+func (suite *HandlerTestSuite) TestMsgClaimReward() {
+	suite.addClaim()
+	msg := incentive.NewMsgClaimReward(suite.addrs[0], "bnb")
```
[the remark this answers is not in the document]
karzak: A little confused as to what the suggestion is here. `NewClaim` is always called with `ukava` as the argument to reward because `types.GovDenom` is hardcoded as the argument to `keeper.AddToClaim` in rewards.go. Do you think NewClaim should throw if `types.GovDenom` is not the denom passed to reward?

x/incentive/keeper/rewards_test.go
```diff
@@ -0,0 +1,341 @@
...
+	var genIDs types.GenesisClaimPeriodIDs
+	suite.keeper.IterateClaimPeriodIDKeysAndValues(suite.ctx, func(denom string, id uint64) (stop bool) {
+		genID := types.GenesisClaimPeriodID{Denom: denom, ID: id}
+		genIDs = append(genIDs, genID)
+		return false
+	})
+	suite.Equal(2, len(genIDs))
+}
+
+func (suite *KeeperTestSuite) TestHelperMethods() {
```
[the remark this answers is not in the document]
karzak: Split this up as I agree it's unclear

x/incentive/keeper/rewards_test.go
karzak: I like using default `SetupTest()` with no state, as I think it's easier to reason about most tests if they explicitly add state and don't add state from other modules unless necessary. For big integration tests like this, I'm fine having a separate helper function and just using comments to describe what the expected state is. I agree it's a preference thing and probably doesn't have one 'right' answer.

x/incentive/types/params_test.go
karzak: Added a check for `denom == ""`

x/incentive/handler_test.go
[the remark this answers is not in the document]
denalimarsh: That would have caught the issue where `types.GovDenom` wasn't passed to `keeper.AddToClaim`, but with the fix the hardcoded value should be fine. Marking as resolved.

x/incentive/types/params_test.go
denalimarsh: This has me concerned about other coin denoms in other modules' params. We should check that we're validating them appropriately everywhere.

denalimarsh: It seems like this is an issue that someone else has thought about. I guess it makes sense for all param validation to be handled at the module level, but I'm interested if there any proposal safeguards in place in the Cosmos SDK for preventing `sdk.Coins` from causing a panic @fedekunze

PR #434, opened 2020-04-14
[R4R] Fix pricefeed sims so does not always go to zero

### LINKED ISSUE #430: Pricefeed simulator's price drifts towards 0
Looks like the price's generated by the pricefeed simultor operations gradually tends towards 0 even with different random seeds.
Not sure what's going on, maybe something to do with rounding?

There's a script to generate prices on branch `price-gen-script`. Running `go test ./x/pricefeed/simulator/operations` will write out a json file with prices (outputs to the `pricefeed/simulator/operations` folder).

jmahess:
Problems were:
- multiplying by the random variable resulted in lower value on average (e.g. 100 x 1.2 x .08 = 96 which is less than 100)
- simulation random number generation is biased towards zero and the "upper bound" but the upper bound can be less than the actual upper bound because it is represented by an integer which can result in truncation. This results in it likely tending down due to the zero and truncation bias (functions in `simulation` e.g. `simulation.RandIntBetween` and the random decimal function)

Update:
- used a new random walk approach. See below for graphs of the new simulation and explanation of how it works

closes #430


x/pricefeed/simulation/operations/msg.go
```diff
@@ -101,26 +107,36 @@ func getExpiryTime(ctx sdk.Context) (t time.Time) {
 // pickNewRandomPrice picks a new random price given the current price
 // It takes the current price then generates a random number to multiply it by to create variation while
 // still being in the similar range. Random walk style.
```
rhuairahrighairidh, line 109: nit: these are a bit out of date given the new changes

In general I'm liking the comments, but it's worth remembering they can get out of sync as code changes (or is copy and pasted to other areas), which can lead to misleading comments.

The general advice is to make sure comments add useful information not obvious from the code itself. For example something like this wouldn't add anything:
```go
    // get next time by adding period to current time
    nextTime := currentTime.Add(period)
```

jmahess:
> nfortunately it looks like `ctx.WithBlockHeight(1)` doesn't work for accessing the genesis prices. It seems to just return the current block's price. I can't think of a way to get the genesis price into the operation func, but I'll think more tomorrow.
> 
> Alternatively if we could get the multiplying algorithm from before to work, then we wouldn't need genesis.

Does is matter if we are pulling the price from the current block or the genesis block? The idea is just to scale the random deviations based on the asset price e.g. BTC may swing by tens of dollars whereas XRP may swing by cents

x/pricefeed/simulation/operations/msg.go
jmahess: 🆗

x/pricefeed/simulation/operations/msg.go
```diff
@@ -101,26 +107,36 @@ func getExpiryTime(ctx sdk.Context) (t time.Time) {
...
+	// Pick random price multiplier
+	limit := sdk.MustNewDecFromStr("0.2")
+	got := simulation.RandomDecAmount(r, limit) // random in the range 0-0.2
 
-	randomPriceMultiplier := simulation.RandomDecAmount(r, got) // get a random number
 	if err != nil {
-		fmt.Errorf("Error generating random price multiplier\n")
+		fmt.Errorf("error generating random price multiplier")
 		return sdk.ZeroDec(), err
 	}
```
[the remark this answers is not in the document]
jmahess: 🆗

jmahess:
> Is it possible to have a random walk that moves some multiple each iteration rather than a fixed additive amount?

Yes. I have implemented a version just using multiplication which does not always end up in zero: https://github.com/Kava-Labs/kava/pull/436

jmahess:
Updating to use the multiplication method.

jmahess:
PR now merges into `master`

jmahess:
Implemented the new random walk strategy. First tried generating the entire series of data points every time but this significantly slows down the simulations. So instead I store the most recent previous price and the take the next step on the random walk based on this. I bounded the prices between zero and twice the genesis price. This means the prices go up and down within a reasonable / realistic range similar to the real assets. I tried running the simulation several times with BTC and BNB and ended up with reasonable values (not tending to zero or infinity). I have included the test file so that it can be replicated to verify not tending to zero or infinity, but will remove the test file before we merge to `master` (test file: https://github.com/Kava-Labs/kava/pull/434/commits/d2f49d1aa2633a4bc08a82ee349ed3a1564de61e#diff-f475e1f75d205858d7fe5205bc816be1

@karzak

jmahess:
Screen Shot 2020-04-15 at 4 05 41 PM

jmahess:
Another run using lower "volatility". This shows the price of BTC varying. This amount of volatility may be more realistic.

Screen Shot 2020-04-15 at 4 27 25 PM

jmahess:
Updated to put upper bound at 10x genesis price. Also put the volatility back to the original high volatility.

Graph of results:

Screen Shot 2020-04-15 at 5 27 28 PM

jmahess:
Updated to use smallest dec instead of zero @karzak

x/pricefeed/simulation/operations/pricegen_test.go
```diff
@@ -0,0 +1,40 @@
+package operations
```
karzak, line 1: We can delete this again

x/pricefeed/simulation/operations/msg.go
```diff
@@ -98,29 +123,29 @@ func getExpiryTime(ctx sdk.Context) (t time.Time) {
...
-		return sdk.ZeroDec(), err
+func pickNewRandomPrice(r *rand.Rand, marketID string) (newPrice sdk.Dec) {
+	startPrice := getStartPrice(marketID)
+	maxPrice := sdk.MustNewDecFromStr("10.0").Mul(startPrice)
+	recentPrice := getRecentPrice(marketID)
+	increment := getIncrement(marketID)
+
+	upDown := r.Intn(2)
+	if upDown == 0 {
+		if maxPrice.Sub(increment).LTE(recentPrice) {
```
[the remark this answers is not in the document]
jmahess: yes you are right. will make the change

x/pricefeed/simulation/operations/msg.go
```diff
@@ -98,29 +123,29 @@ func getExpiryTime(ctx sdk.Context) (t time.Time) {
...
+			newPrice = maxPrice
+		} else {
+			newPrice = recentPrice.Add(increment)
+		}
+	} else {
+		if sdk.SmallestDec().Add(increment).GTE(recentPrice) {
```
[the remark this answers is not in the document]
jmahess: 🆗

x/pricefeed/simulation/operations/pricegen_test.go
jmahess: ok. deleted.

jmahess:
Made the updates as suggested in the code review. Feel free to merge whenever or let me know if anything else needs updating @karzak

x/pricefeed/simulation/operations/msg.go
```diff
@@ -98,29 +123,29 @@ func getExpiryTime(ctx sdk.Context) (t time.Time) {
 	return t
 }
 
-// pickNewRandomPrice picks a new random price given the current price
```
fedekunze, line 101: nit, I liked the comments here. Can you maybe readd them?

jmahess: @rhuairahrighairidh

PR #440, opened 2020-04-16
[R4R] Update sim tooling

### LINKED ISSUE #437: Should TestAppStateDeterminism run each seed 5 times?
In `app/sim_test.go (TestAppStateDeterminism)`, 3 seeds are tested, and each seed is run 5 times. 

```
numSeeds := 3
numTimesToRunPerSeed := 5
```

1. I can't think of a good reason to run an individual seed more than twice. 
2. Further, I think that `NumSeeds` should be 1, and if you want to really test determinism, you run the function many times (one seed per run), rather than having an arbitrary floor of three seeds per run. 

Thoughts?

karzak:
closes #437 

* Went with Ruaridh's suggestion to just remove the multiple runs from `TestAppStateDeterminism` and used `config.Seed` for the run. This also has the advantage of increasing reproducability, as previously seeds were chosen with `rand.int63()`, which mean't you had to hard code the value if you wanted to do the same seed twice. 

**Note** this means that future determinism testing should test multiple seeds in independent runs.

* Added a new `test-basic` method that runs the tests for each module (should be fast since most will be cached) and a 5 block sim. Total run time is closer to 30seconds.

### DIFF
--- a/app/sim_test.go
+++ b/app/sim_test.go
@@ -662,12 +662,12 @@ func TestAppStateDeterminism(t *testing.T) {
 	config.OnOperation = false
 	config.AllInvariants = false
 
-	numSeeds := 3
-	numTimesToRunPerSeed := 5
+	numSeeds := 1
+	numTimesToRunPerSeed := 2
 	appHashList := make([]json.RawMessage, numTimesToRunPerSeed)
 
 	for i := 0; i < numSeeds; i++ {
-		config.Seed = rand.Int63()
+		config.Seed = 8674665223082153551
 
 		for j := 0; j < numTimesToRunPerSeed; j++ {
 			logger := log.NewNopLogger()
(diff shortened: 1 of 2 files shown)


rhuairahrighairidh:
I think running more than one seed per test gets higher code coverage. It should increase the likelihood of executing the line that produces undeterministic behaviour. Though running the sim for longer would probably be equivalent, and having the test pick a hardcoded number of seeds randomly sounds awkward to use.

We could remove that seed loop entirely? Then I believe the seed can be set with the `-Seed` flag, and whoever is running the test is free to pick the seed and numblocks as needed. They can run high numBlocks and/or lots of different seeds for a big test, or less for CI.

app/sim_test.go
```diff
@@ -662,8 +662,8 @@ func TestAppStateDeterminism(t *testing.T) {
 	config.OnOperation = false
 	config.AllInvariants = false
 
-	numSeeds := 3
-	numTimesToRunPerSeed := 5
+	numSeeds := 1
```
fedekunze, line 665: hmm the idea is to run the non-determinism with 2 different seeds at least. Eg: one seed could run deterministic transactions but others not.
```suggestion
	numSeeds := 2
```

PR #443, opened 2020-04-16
[R4R] Fix pricefeed non-determinism

karzak:
Fixes issue where multiple calls to `SimulateFromSeed` with the same seed would use the old 'recentPrice' variables and thus not be deterministic between runs. Uses  `sync.Once` to initialize a random walk for each market at the beginning of sim. 

Note: I'm pretty sure this means that calling `SimulateFromSeed` multiple times in the same simulation with different seeds, as is done in `TestAppStateDeterminism`, will create the same array of prices for all runs. There is probably a way to reinitialize the prices whenever the seed changes, but I'll call this sufficient for now, as our general simulation strategy strategy is to run lots of separate simulations with different seeds.

### DIFF
--- a/x/pricefeed/simulation/operations/msg.go
+++ b/x/pricefeed/simulation/operations/msg.go
@@ -3,6 +3,7 @@ package operations
 import (
 	"fmt"
 	"math/rand"
+	"sync"
 	"time"
 
 	"github.com/cosmos/cosmos-sdk/baseapp"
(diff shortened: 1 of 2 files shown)


x/pricefeed/simulation/operations/msg.go
```diff
@@ -15,69 +16,53 @@ import (
...
 	// get a pricefeed handler
 	handler := pricefeed.NewHandler(keeper)
 
 	return func(r *rand.Rand, app *baseapp.BaseApp, ctx sdk.Context, accs []simulation.Account) (
 		simulation.OperationMsg, []simulation.FutureOperation, error) {
+		genPrices.Do(func() {
+			// generate a random walk for each asset exactly once, with observations equal to the number of blocks in the sim
+			for _, m := range keeper.GetMarkets(ctx) {
+				startPrice := getStartPrice(m.MarketID)
+				// allow prices to fluctuate from 10x GAINZ to 100x REKT
```
jmahess, line 37: 😅😅😅

x/pricefeed/simulation/operations/msg.go
```diff
@@ -98,6 +83,51 @@ func SimulateMsgUpdatePrices(keeper keeper.Keeper) simulation.Operation {
...
+		return sdk.MustNewDecFromStr("15")
+	case "xrp:usd":
+		return sdk.MustNewDecFromStr("0.25")
+	}
+	return sdk.MustNewDecFromStr("100")
+}
+
+func getIncrement(marketID string) (increment sdk.Dec) {
+	startPrice := getStartPrice(marketID)
+	divisor := sdk.MustNewDecFromStr("20")
```
jmahess, line 100: Increasing `divisor` decreases volatility. E.g. setting to `100` instead of `20` will result in less volatility. Nothing to update just noting as an FYI

x/pricefeed/simulation/operations/msg.go
```diff
@@ -15,69 +16,53 @@ import (
...
-
 // SimulateMsgUpdatePrices updates the prices of various assets by randomly varying them based on current price
-func SimulateMsgUpdatePrices(keeper keeper.Keeper) simulation.Operation {
+func SimulateMsgUpdatePrices(keeper keeper.Keeper, blocks int) simulation.Operation {
 	// get a pricefeed handler
 	handler := pricefeed.NewHandler(keeper)
 
 	return func(r *rand.Rand, app *baseapp.BaseApp, ctx sdk.Context, accs []simulation.Account) (
 		simulation.OperationMsg, []simulation.FutureOperation, error) {
+		genPrices.Do(func() {
```
jmahess, line 33: Good strategy to use `sync.Once` to ensure only run one time and then memoize the aseet prices for fast lookup by `pickNewRandomPrice`

x/pricefeed/simulation/operations/msg.go
```diff
@@ -98,6 +83,51 @@ func SimulateMsgUpdatePrices(keeper keeper.Keeper) simulation.Operation {
...
+func pickNewRandomPrice(marketID string, blockHeight int) (newPrice sdk.Dec) {
+	switch marketID {
+	case "btc:usd":
+		return btcPrices[blockHeight-1]
+	case "bnb:usd":
+		return bnbPrices[blockHeight-1]
+	case "xrp:usd":
+		return xrpPrices[blockHeight-1]
+	}
+	panic("invalid price request")
```
jmahess, line 128: Ah yes good to include panic here instead of just returning a default price as it was before

PR #468, opened 2020-04-24
[HOTFIX] Add incentive params to testnet-6000's genesis file 

denalimarsh:
`make install` with the existing testnet-6000 genesis file gives a consensus failure. This PR fixes the issue by adding incentive module params to the file.


denalimarsh:
@karzak are there any specific rewards that you'd like to include in the `rewards: []` param?

karzak:
Something similar to:
```
types.NewParams(
			true, types.Rewards{types.NewReward(true, "bnb", c("ukava", 50000000000), time.Hour*7*24, time.Hour*24*365, time.Hour*7*24)},
		),
```
So paying out 50000 KAVA per week to USDX minters

PR #470, opened 2020-04-27
fix CLI tests

fedekunze:
closes: #469


Makefile
```diff
@@ -135,7 +135,7 @@ test:
 test-rest:
 	rest_test/./run_all_tests_from_make.sh
```
rhuairahrighairidh, line 137:
```suggestion
# Run cli integration tests
# `-p 4` to use 4 cores, `-tags cli_test` to tell go not to ignore the cli package
# These tests use the `kvd` or `kvcli` binaries in the build dir, or in `$BUILDDIR` if that env var is set.
```

rhuairahrighairidh: To keep things tidy, do you want to delete `cli_test/README.md` and add a short description to the makefile.

PR #472, opened 2020-04-28
Add kava modules to TestAppImportExport, fix bugs that were causing test to fail

karzak:


x/cdp/keeper/cdp.go
```diff
@@ -272,13 +270,8 @@ func (k Keeper) IndexCdpByOwner(ctx sdk.Context, cdp types.CDP) {
...
 	}
-	for _, id := range cdpIDs {
-		if id == cdp.ID {
-			return
-		}
-		cdpIDs = append(cdpIDs, cdp.ID)
-		store.Set(cdp.Owner, k.cdc.MustMarshalBinaryLengthPrefixed(cdpIDs))
-	}
+	cdpIDs = append(cdpIDs, cdp.ID)
+	store.Set(cdp.Owner, k.cdc.MustMarshalBinaryLengthPrefixed(cdpIDs))
```
rhuairahrighairidh, line 274: does the loop still need to be there to avoid adding duplicate IDs into the slice? eg:
```go
for _, id := range cdpIDs {
    if id == cdp.ID {
        return // do nothing as the slice already contains the id
    }
cdpIDs = append(cdpIDs, cdp.ID)
store.Set(cdp.Owner, k.cdc.MustMarshalBinaryLengthPrefixed(cdpIDs))
```

karzak: Given that this function is only called by AddCDP and we have already verified that a CDP for this owner and denom does not exist, this check was purely redundant. Ie, if an address has an XRP CDP and calls 'AddCDP' with XRP as the denom, it will fail before this point because the CDP already exists.

PR #482, opened 2020-05-01
[R4R] Enable bep3 claim txs from new addresses

denalimarsh:
**Goal**: remove claiming kava via faucet step from the bep3 user flow.

**Approach**: add an account keeper to bep3 module and automatically register an account in state for the atomic swap's recipient address.

**Details**: automatic registration occurs if and only if the atomic swap is incoming (i.e. sent by deputy) and the recipient's address is not found in state. If the recipient's address has an invalid bech32 encoding, the swap is rejected and the address is not registered. Registered accounts are able to claim atomic swaps. After receiving their claimed bnb, the user can send bnb to other addresses in state.

### DIFF
--- a/x/bep3/types/expected_keepers.go
+++ b/x/bep3/types/expected_keepers.go
@@ -6,7 +6,7 @@ import (
 	supplyexported "github.com/cosmos/cosmos-sdk/x/supply/exported"
 )
 
-// SupplyKeeper defines the expected supply keeper
+// SupplyKeeper defines the expected supply keeper (noalias)
 type SupplyKeeper interface {
 	GetModuleAddress(name string) sdk.AccAddress
 	GetModuleAccount(ctx sdk.Context, moduleName string) supplyexported.ModuleAccountI
@@ -19,7 +19,7 @@ type SupplyKeeper interface {
 	MintCoins(ctx sdk.Context, name string, amt sdk.Coins) error
 }
 
-// AccountKeeper defines the expected account keeper
+// AccountKeeper defines the expected account keeper (noalias)
 type AccountKeeper interface {
 	GetAccount(ctx sdk.Context, addr sdk.AccAddress) authexported.Account
 	NewAccountWithAddress(ctx sdk.Context, addr sdk.AccAddress) authexported.Account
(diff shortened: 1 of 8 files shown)


x/bep3/types/expected_keepers.go
```diff
@@ -2,10 +2,11 @@ package types
 
 import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	authexported "github.com/cosmos/cosmos-sdk/x/auth/exported"
 	supplyexported "github.com/cosmos/cosmos-sdk/x/supply/exported"
 )
 
-// SupplyKeeper defines the expected supply Keeper
+// SupplyKeeper defines the expected supply keeper
```
karzak, line 9: can be a separate PR, but we should follow the expected keepers pattern (see CDP, which was just updated, for exmaple).

Basically, add (noalias) to each type in `types/expected_keepers.go` and then in `module.go` change `NewAppModule` to take `types.AccountKeeper` and `types.SupplyKeeper` as args.

denalimarsh: Updated to expected keepers pattern in https://github.com/Kava-Labs/kava/pull/482/commits/dcceda48dc241fb7fae7cedea70bb3c67a0c0b69, https://github.com/Kava-Labs/kava/pull/482/commits/f1ef515f20f11020b84830a62cd0ff09e681b5a9

denalimarsh:
This PR is **blocked** by https://github.com/Kava-Labs/kava/pull/483. Conflicts will need to be resolved once https://github.com/Kava-Labs/kava/pull/483 is merged.

denalimarsh:
> Only question is: are zero gas fees required for 0 balance addresses to claim swaps? From the cdp review, having non-zero gas fees mitigates some DoS methods.

I'm not sure, we'll need to do some testing on gas fees system-wide before the next testnet.

> also nit: A comment explaining the purpose of registering the account would help future maitainers.

Added in https://github.com/Kava-Labs/kava/pull/482/commits/b70c52cf78c088cc07cca22f56e194f3b5799bb7.

PR #483, opened 2020-05-01
[R4R] Update bep3 module account permissions

denalimarsh:
**Goal**: Remove unused Burner, Minter bep3 macc permissions.


x/bep3/types/expected_keepers.go
```diff
@@ -2,18 +2,12 @@ package types
 
 import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
-	supplyexported "github.com/cosmos/cosmos-sdk/x/supply/exported"
 )
 
 // SupplyKeeper defines the expected supply Keeper
 type SupplyKeeper interface {
 	GetModuleAddress(name string) sdk.AccAddress
-	GetModuleAccount(ctx sdk.Context, moduleName string) supplyexported.ModuleAccountI
```
karzak, line 11: Add this back and update `InitGenesis` to call `GetModuleAccount`.  `GetModuleAddress` doesn't actually check that the account exists. You can remove that check from `keeper.go`

denalimarsh: Updated in https://github.com/Kava-Labs/kava/pull/483/commits/55d4762fe072417f36790d791e1cc41eeacfcf56. Noticed that `GetModuleAddress` isn't used in the bep3, cdp, or kavadist modules but it is defined in each module's `type SupplyKeeper interface`.

PR #488, opened 2020-05-06
x/bep3: genesis validation

fedekunze:


x/bep3/types/swap_test.go
```diff
@@ -44,95 +44,205 @@ func (suite *AtomicSwapTestSuite) SetupTest() {
...
 			},
 			true,
 		},
+		{
+			"invalid amount",
+			types.AtomicSwap{
+				Amount: sdk.Coins{sdk.Coin{Denom: "BNB", Amount: sdk.NewInt(10)}},
+			},
+			false,
+		},
```
denalimarsh, line 76: Fails even with a valid amount due to `the length of random number hash should be 32`. Recommend checking that the returned error matches expected error.

x/bep3/genesis_test.go
```diff
@@ -278,19 +278,14 @@ func (suite *GenesisTestSuite) TestGenesisState() {
 	}
 
 	for _, tc := range testCases {
-		suite.SetupTest()
 		if tc.expectPass {
-			suite.Run(tc.name, func() {
```
denalimarsh, line 283: We added `suite.Run(tc.name, func() ... )` because the test suite was reusing state and swallowing panics. Tested a few cases and the panics were caught due to genesis param validation, but want @karzak to take a look before approving.

denalimarsh: It also looks like tests are now failing because `ClosedBlock` can't be 0. For example, `name: "atomic swap asset type is unsupported"` actually fails due to `failed to validate bep3 genesis state: closed block cannot be 0`.

Can resolve by updating `ClosedBlock` to 1 for each test case, but I recommend checking the error message so we catch similar issues in the future. I can do it throughout the BEP3 module if you'd like as it's a legacy issue and adding the expected error msg to each test case is pretty tedious.

karzak: 👍 to checking the error message to catch test failing for reasons other than the intended reason. Can be in a follow up PR.

I had the same thoughts about `suite.Run, but indeed the tests are catching things properly here. Not sure if it was use of `Suite.Nil` versus `suite.Require()` methods that was causing the problem previously. Seems fine to merge as is though.

x/bep3/types/swap_test.go
fedekunze: I double-checked and the error matches the message

x/bep3/genesis_test.go
fedekunze: yeah, that's why I changed it to require. Otherwise, it reuses the state. We can do a test cleanup in another PR since it's out of scope from this one

denalimarsh: Created issue https://github.com/Kava-Labs/kava/issues/498 for the test cleanup

denalimarsh:
> This is fine to merge. Let's create an issue to check the specific panic in 'genesis_test.go` so we don't have regressions that slip through.

Created https://github.com/Kava-Labs/kava/issues/499 for this.

PR #497, opened 2020-05-11
Auction audit revisions

rhuairahrighairidh:
This pr applies various fixes from the review:
- tidied up events
- paniced in more places

When to panicing roughly follows stackoverflow advice: https://stackoverflow.com/a/26737099/2247240
Anything that is obviously a progammer's error result in panic (like passing -3 to a square root func). Error values are used for expected/likely errors (such as http request failing). I left errors in for things like not finding an auction in the database even when it should be there, and not having enough coins when they should be there. They felt less obviously a programmer error.
All panics in processing a tx/msg are caught anyway. And any error raised during begin blocker results in a panic. So the choice between them is not super consequential.

The coin splitting math in `splitIntIntoWeightedBuckets` (used when paying out collateral auctions) was checked over as well. Interestingly adding a check to ensure the num weighted addresses passed to `StartCollateralAuction` is > 0 brought up a bug cdp in `AuctionCollateral`.


x/auction/keeper/auctions.go
```diff
@@ -336,16 +341,20 @@ func (k Keeper) PlaceReverseBidCollateral(ctx sdk.Context, a types.CollateralAuc
...
-		err = k.supplyKeeper.SendCoinsFromModuleToAccount(ctx, types.ModuleName, a.LotReturns.Addresses[i], sdk.NewCoins(payout))
-		if err != nil {
-			return a, err
+		// if the payout amount is 0, don't send 0 coins
+		if payout.IsPositive() {
+			err = k.supplyKeeper.SendCoinsFromModuleToAccount(ctx, types.ModuleName, a.LotReturns.Addresses[i], sdk.NewCoins(payout))
+			if err != nil {
+				return a, err
+			}
 		}
```
fedekunze, line 358:
```suggestion
		if !payout.IsPositive() {
			break
		}
		err = k.supplyKeeper.SendCoinsFromModuleToAccount(ctx, types.ModuleName, a.LotReturns.Addresses[i], sdk.NewCoins(payout))
		if err != nil {
			return a, err
		}
```

PR #500, opened 2020-05-11
x/auction: types validation

fedekunze:


x/auction/types/auctions.go
```diff
@@ -66,15 +68,35 @@ func (a BaseAuction) GetType() string { return "base" }
 
 // Validate verifies that the auction end time is before max end time
 func (a BaseAuction) Validate() error {
-	if a.EndTime.After(a.MaxEndTime) {
-		return fmt.Errorf("MaxEndTime < EndTime (%s < %s)", a.MaxEndTime, a.EndTime)
+	// ID can be 0 for surplus, Debt and collateral auctions
```
fedekunze, line 71: ID is not used for any but the base auction type. Should we consider removing it? @karzak

x/auction/types/auctions.go
```diff
@@ -66,15 +68,35 @@ func (a BaseAuction) GetType() string { return "base" }
...
+	if strings.TrimSpace(a.Initiator) == "" {
+		return errors.New("auction initiator cannot be blank")
 	}
 	if !a.Lot.IsValid() {
 		return fmt.Errorf("invalid lot: %s", a.Lot)
 	}
+	// NOTE: bidder can be nil for Surplus and Collateral auctions
+	if a.Bidder != nil && a.Bidder.Empty() {
```
karzak, line 79: This check is failing on 
```
go test ./app -run TestAppImportExport -Enabled -Commit -NumBlocks=10 -BlockSize=200 -Seed 4 -v -timeout 24h
```

pretty sure it needs to be 
```suggestion
	if !a.Bidder.Equals(sdk.AccAddress{}) && a.Bidder.Empty() {
```

x/auction/types/auctions.go
```diff
@@ -66,15 +68,35 @@ func (a BaseAuction) GetType() string { return "base" }
...
+	if a.EndTime.IsZero() || a.MaxEndTime.IsZero() {
+		return errors.New("end time cannot be zero")
+	}
+	if a.EndTime.After(a.MaxEndTime) {
+		return fmt.Errorf("MaxEndTime < EndTime (%s < %s)", a.MaxEndTime, a.EndTime)
+	}
+	if a.HasReceivedBids && !a.Bid.IsPositive() {
+		return fmt.Errorf("cannot have a zero value bid when auction has a true HasReceivedBids flag value")
+	}
+	if !a.HasReceivedBids && a.Bid.IsPositive() {
```
karzak, line 97: This check is failing on sims. `auction/simulation/genesis.go` needs to be updated to set `HasReceivedBids` to be true for the auction that is created at genesis.

x/auction/types/auctions.go
```diff
@@ -66,15 +68,35 @@ func (a BaseAuction) GetType() string { return "base" }
...
+		return errors.New("auction initiator cannot be blank")
 	}
 	if !a.Lot.IsValid() {
 		return fmt.Errorf("invalid lot: %s", a.Lot)
 	}
+	// NOTE: bidder can be nil for Surplus and Collateral auctions
+	if a.Bidder != nil && a.Bidder.Empty() {
+		return errors.New("auction bidder cannot be empty")
+	}
+	if a.Bidder != nil && len(a.Bidder) != sdk.AddrLen {
```
karzak, line 82: ditto

```suggestion
	if !a.Bidder.Equals(sdk.AccAddress{}) && len(a.Bidder) != sdk.AddrLen {
```

x/auction/types/auctions.go
fedekunze: the ID will always be 0 for the named actions though

karzak: Yes, the ID is set by the keeper when it is first stored.

x/auction/types/auctions.go
```diff
@@ -173,9 +193,10 @@ func NewDebtAuction(buyerModAccName string, bid sdk.Coin, initialLot sdk.Coin, e
 			Lot:             initialLot,
 			Bidder:          supply.NewModuleAddress(buyerModAccName), // send proceeds from the first bid to the buyer.
 			Bid:             bid,                                      // amount that the buyer is buying - doesn't change over course of auction
-			HasReceivedBids: false,                                    // new auctions don't have any bids
+			HasReceivedBids: true,                                     // new auctions don't have any bids
```
fedekunze, line 196: ^^

karzak: This should be false.

The problem is here https://github.com/Kava-Labs/kava/blob/9932169a6786e2a58ef9bc5d9c4195246c9be8b0/x/auction/simulation/genesis.go#L74

It should create an auction and then manually set `HasReceivedBids` to `true`

fedekunze: If I set it to false the auction won't be valid because of the following tho:

```go
if !a.HasReceivedBids && a.Bid.IsPositive() {
		return fmt.Errorf("cannot have a positive value bid when auction has a false HasReceivedBids flag value")
	}
```

x/auction/types/auctions.go
```diff
@@ -292,18 +314,28 @@ func (wa WeightedAddresses) Validate() error {
...
+		if wa.Weights[i].IsNegative() {
+			return fmt.Errorf("weight %d contains a negative amount: %s", i, wa.Weights[i])
+		}
+		totalWeight = totalWeight.Add(wa.Weights[i])
 	}
+
 	if !totalWeight.IsPositive() {
 		return fmt.Errorf("total weight must be positive")
 	}
+
```
rhuairahrighairidh, line 339: Could the `WeighteAddresses.Validate` function be used here? These checks could be moved into that function

x/auction/types/msg.go
```diff
@@ -34,9 +35,15 @@ func (msg MsgPlaceBid) Type() string { return "place_bid" }
 
 // ValidateBasic does a simple validation check that doesn't require access to state.
 func (msg MsgPlaceBid) ValidateBasic() error {
+	if msg.AuctionID == 0 {
+		return errors.New("auction id cannot be zero")
+	}
```
rhuairahrighairidh, line 40: What was the reason for disallowing 0 as an ID?
The default auctionID is 1 but you could set it in genesis to be 0, then the first auction created would have ID 0.

x/auction/types/auctions.go
```diff
@@ -66,15 +68,32 @@ func (a BaseAuction) GetType() string { return "base" }
...
+	}
+	if a.EndTime.After(a.MaxEndTime) {
+		return fmt.Errorf("MaxEndTime < EndTime (%s < %s)", a.MaxEndTime, a.EndTime)
+	}
+	if a.HasReceivedBids && !a.Bid.IsPositive() {
+		return fmt.Errorf("cannot have a zero value bid when auction has a true HasReceivedBids flag value")
+	}
+	if !a.HasReceivedBids && a.Bid.IsPositive() {
+		return fmt.Errorf("cannot have a positive value bid when auction has a false HasReceivedBids flag value")
+	}
```
rhuairahrighairidh, line 96: These checks aren't valid for all the auction types, I would probably remove them (though you could do different checks for each auction type if you realy wanted).

Bad naming on my part, but "bidding" on an auction doesn't necessarily mean the "Bid" value is increasing:

- SurplusAuction:    Lot stays constant, Bid increases on each MsgPlaceBid
- DebtAuction:        Bid stays constant, Lot decreases on each MsgPlaceBid
- CollateralAuction: Lot stays constant, while Bid increases up to a max, at which point Bid stays constant and Lot decreases

x/auction/types/auctions.go
rhuairahrighairidh: ^ That check isn't quite right for DebtAuctions. See my comment above.

x/auction/types/msg.go
fedekunze: mostly because 0 is the default value for integers (i.e uninitialized) we should avoid using 0 as a valid ID for that reason.

x/auction/types/auctions.go
fedekunze: this is the `Validate()` function

fedekunze:
I think I addressed all comments 👍

PR #509, opened 2020-05-13
[R4R] Update secure RNG to generate bytes directly

denalimarsh:
BEP3 recently updated RNG was incorrectly failing to validate some numbers generated on bnbchain. This PR fixes the issue such that secret generation and validation work 100% of the time between Kava and bnbchain by directly generating hex-encodable bytes instead of using big.Int as an intermediary. The go-sdk has been updated with [this feature](https://github.com/Kava-Labs/go-sdk/pull/13) as well.

The function could return `[32]bytes` instead (or be renamed, as we're technically no longer generating a number) but I'm in favor of just merging a solution that works given our time constraints.

### DIFF
--- a/x/bep3/client/cli/query.go
+++ b/x/bep3/client/cli/query.go
@@ -82,10 +82,10 @@ func QueryCalcRandomNumberHashCmd(queryRoute string, cdc *codec.Codec) *cobra.Co
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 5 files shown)


denalimarsh:
Playground test for any future viewers: https://play.golang.org/p/VkY0fe8vvSg

Issue #506, opened 2020-05-12
Enforce minimum swap amount greater than bep3-deputy's fixed fee


denalimarsh:
Creating an atomic swap with an amount less than the bep3-deputy's fixed fee is allowed, but the updated bep3-deputy will not relay the swap. At the current fixed fee of 1000, a swap of 100bnb is incorrectly accepted by Kava.

**Example**
Create an atomic swap of 100bnb with cmd
```
kvcli tx bep3 create kava1l0xsq2z7gqd7yly0g40y5836g0appumark77ny bnb1urfermcg92dwq36572cx4xg84wpk3lfpksr5g7 bnb1uky3me9ggqypmrsvxk7ur6hqkzq7zmv4ed4ng7 now 100bnb 360 --from validator
```

bep3-deputy logs
```
2020-05-12 14:08:09 ERROR OtherSendHTLT reject swap for wrong out_amount,
other_chain_swap_id=44bf4f69bd5635ef9c6542f1123754f59fd03fdb047a2c2e5218ef32e32efc48,
random_number_hash=d555351c36ad7934c1261520a1e5b9e45d34b46ea6f01b5f8df3031134efd17e,
out_amount=-900
```

denalimarsh:
This case was discovered by manually submitting incoming swaps, something that is impossible as all incoming swaps will have been validated by bnbchain prior to deputy relay. However, we should confirm that the fixed fee is not an issue with outgoing swaps as well.

PR #511, opened 2020-05-13
[R4R] Add BnbDeputyFixedFee param to BEP3 module

denalimarsh:
Resolves https://github.com/Kava-Labs/kava/issues/506

**Problem**: The bep3-deputy specifies an 'other_chain_fixed_fee', a fee paid to the deputy for outgoing swaps (Kava -> BnbChain). If a swap's amount is less than the other_chain_fixed_fee, the bep3-deputy will not relay it. This is usually not a problem as the fixed fee is very, very small (1000bnb, or 10**10-14BNB) but it means micro outgoing swaps would be unfulfillable. An inflexible solution is to set the other_chain_fixed_fee to 0 in the bep3-deputy config, but if the other_chain_fixed_fee is changed then the problem is reintroduced.

**Solution**: This PR adds a BnbDeputyFixedFee parameter that should be 1:1 with the bep3-deputy's config, similar to the BnbDeputyAddress. It was difficult to write specific test cases for the new param - since it's an uint64 we can't pass negative values to it.

**NOTE: ~~base branch is set to dm-fix-rng for ease of review and will be set to master after https://github.com/Kava-Labs/kava/pull/509 is merged.~~

### DIFF
--- a/x/bep3/keeper/swap.go
+++ b/x/bep3/keeper/swap.go
@@ -175,7 +175,7 @@ func (k Keeper) ClaimAtomicSwap(ctx sdk.Context, from sdk.AccAddress, swapID []b
 			sdk.NewAttribute(types.AttributeKeyRecipient, atomicSwap.Recipient.String()),
 			sdk.NewAttribute(types.AttributeKeyAtomicSwapID, hex.EncodeToString(atomicSwap.GetSwapID())),
 			sdk.NewAttribute(types.AttributeKeyRandomNumberHash, hex.EncodeToString(atomicSwap.RandomNumberHash)),
-			sdk.NewAttribute(types.AttributeKeyRandomNumber, string(randomNumber)),
+			sdk.NewAttribute(types.AttributeKeyRandomNumber, hex.EncodeToString(randomNumber)),
 		),
 	)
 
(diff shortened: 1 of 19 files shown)


x/bep3/types/params.go
```diff
@@ -158,6 +169,20 @@ func validateBnbDeputyAddressParam(i interface{}) error {
...
 }
 
+func validateBnbDeputyFixedFeeParam(i interface{}) error {
+	bnbDeputyFixedFee, ok := i.(uint64)
+	if !ok {
+		return fmt.Errorf("invalid parameter type: %T", i)
+	}
+
+	// Sanity check (impossible for an unsigned integer to be negative)
+	if bnbDeputyFixedFee < 0 {
```
karzak, line 179: nit: this is unnecessary, a better check might be
```suggestion
	if bnbDeputyFixedFee >= uint64(0) - uint64(1) {
```

or just omit it

x/bep3/keeper/swap.go
```diff
@@ -64,11 +64,15 @@ func (k Keeper) CreateAtomicSwap(ctx sdk.Context, randomNumberHash []byte, times
 		}
 		err = k.IncrementIncomingAssetSupply(ctx, amount[0])
 	case types.Outgoing:
+		// Amount in outgoing swaps must be greater than the deputy's fixed fee. Incoming swaps
+		// have already had their fees collected by the deputy during the relay process.
```
karzak, line 68: nit: move comment about Incoming swaps to that case

x/bep3/types/params.go
denalimarsh: The proposed solution gives error `constant -1 overflows uint64`. Omitted the check entirely as the uint64 datatype already guarantees a non-negative number.

x/bep3/keeper/swap.go
denalimarsh: updated in latest

PR #512, opened 2020-05-15
[R4R] Change AssetSupply increment from !GTE to LT

denalimarsh:
This PR addresses confusion around asset supply incrementing.


x/bep3/keeper/asset.go
```diff
@@ -15,7 +15,7 @@ func (k Keeper) IncrementCurrentAssetSupply(ctx sdk.Context, coin sdk.Coin) erro
 	}
 
 	// Resulting current supply must be under asset's limit
```
fedekunze, line 17: nit, can you update the comments too?

PR #514, opened 2020-05-18
x/pricefeed: GenesisState validation

### LINKED ISSUE #502: Missing pricefeed GenesisState validation
Validate genesis `PostedPrices`

fedekunze:
closes #502

### DIFF
--- a/x/pricefeed/types/market.go
+++ b/x/pricefeed/types/market.go
@@ -146,7 +146,7 @@ func (pps PostedPrices) Validate() error {
 		if err := pp.Validate(); err != nil {
 			return err
 		}
-		seenPrices[pp.MarketID] = true
+		seenPrices[pp.MarketID+pp.OracleAddress.String()] = true
 	}
 
 	return nil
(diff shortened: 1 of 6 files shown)


x/pricefeed/types/market.go
```diff
@@ -72,9 +114,44 @@ func NewPostedPrice(marketID string, oracle sdk.AccAddress, price sdk.Dec, expir
...
+
 // PostedPrices type for an array of PostedPrice
 type PostedPrices []PostedPrice
 
+// Validate checks if all the posted prices are valid and there are no duplicated
+// entries.
+func (pps PostedPrices) Validate() error {
+	seenPrices := make(map[string]bool)
+	for _, pp := range pps {
+		if seenPrices[pp.MarketID] {
```
karzak, line 142: This should check if the `marketID` and `OracleAddress` is duplicated. It's fine (and expected) to have multiple posted prices when you export genesis (1 from each oracle for each market).

fedekunze: something like this?

changing `seenPrices := make(map[string]string)` and then 
```go
if oracleAddr, ok := seenPrices[pp.MarketID]l ok && oracleAddr == pp.OracleAddress.String()
```

PR #518, opened 2020-05-20
v0.8 Migration Scripts

rhuairahrighairidh:
Changes in this PR:
- replaces `kvd migrate` with cusom cmd to migrate an exported genesis.json between kava v0.3 and v0.8
- adds `kvd write-params` cmd to add suggested parameters for `kava-3` to a genesis.json
- small change to cdp `InitGenesis` that aligns behaviour with kavadist and incentive
- removed a couple of dependencies from `go.mod` that weren't being used
- super minor reorg of `contrib/testnet-6000` folder

Notes:
- Theres a write up on how migration works in `migration/doc.go`, and draft validator upgrade steps in `contrib/kava-3/`.
- All migration stuff is under a new top level folder `migration`. It includes a lot of copy pasted types from the sdk.
- I created a cmd for adding the kava-3 params to a genesis file, it seemed the safest option compared to a bash or python script. If we want to cut the release before params are finalilzed we could pull this cmd out into another repo.
- We have our own versions of the sdk module migrations because the originals are out of date
- The sdk supports migrating a gensis.json from any past version to the current. I think we shouldn't support this for kava as it adds a big mainenance burden and doesn't seem very useful. Migration code is really error prone - it involves a lot of copy and pasting types, manual checking, and will break silently as dependencies are updated. I expect this current migration code to be deleted and rewritten for kava's next release.
- kava v0.3 used sdk commit 18de63 and tendermint v0.32
- kava v0.8 uses sdk v0.38 and tendermint v0.33
- Ran into a problem unmarshalling large multisig public keys in sdk v.0.38. They're encoded into a long bech32 string in genesis.json, but on `InitGenesis` decoding fails as there's a check that the bech32 string isn't longer than 90 characters. Got around this problem by forking the bech32 library (from the commit used by tendermint) and increasing the limit to 2048 characters.
  - enigma also ran into this problem: https://github.com/cosmos/cosmos-sdk/pull/6237
  - not a problem in future sdk versions as public keys are no longer marshalled to bech32 https://github.com/cosmos/cosmos-sdk/pull/6054/files

non blocking TODOs
- check what evidences.max_age_duration should be


migrate/cmd.go
```diff
@@ -0,0 +1,83 @@
...
+const (
+	flagGenesisTime = "genesis-time"
+	flagChainID     = "chain-id"
+)
+
+// MigrateGenesisCmd returns a command to execute genesis state migration.
+func MigrateGenesisCmd(_ *server.Context, cdc *codec.Codec) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "migrate [genesis-file]",
+		Short:   "Migrate genesis from kava v0.3 to v0.8",
```
alexanderbez, line 27: So the main difference between kava and how the SDK does migration, is that the SDK supports multiple versions at a time and Kava supports only a single point-to-point migration?

migrate/doc.go
```diff
@@ -0,0 +1,37 @@
+/*
+Migrate handles translating the state of the blockchain between software versions.
+
+For example, as modules are changed over time the structure of the data they store changes. The data structure must be
+migrated to the new structure of the newer versions.
+
+There are two types of migration:
+- **genesis migration** that converts an exported genesis file from the old software to a new genesis file for the new software
+- **upgrade module migration** that converts data in the database itself from the old version to the new
```
alexanderbez, line 9: I'm not familiar with this type of upgrade. How does this even happen?


EDIT: I've realized you mean live upgrades here. I would recommend you use that nomenclature instead. Module migration isn't a thing really.

migrate/v0_8/sdk/auth/v18de63/auth_codec.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"github.com/cosmos/cosmos-sdk/codec"
+)
+
+// RegisterCodec registers concrete types on the codec
+func RegisterCodec(cdc *codec.Codec) {
+	cdc.RegisterInterface((*GenesisAccount)(nil), nil)
+	cdc.RegisterInterface((*Account)(nil), nil)
+	cdc.RegisterConcrete(&BaseAccount{}, "cosmos-sdk/Account", nil)
+	//cdc.RegisterConcrete(StdTx{}, "cosmos-sdk/StdTx", nil) // not needed for genesis migration
```
alexanderbez, line 12: remove this line if its not needed.

migrate/v0_8/sdk/auth/v18de63/auth_exported.go
```diff
@@ -0,0 +1,43 @@
...
+// This allows account types to be copy and pasted into this package without all their methods.
+
+// Account is an interface used to store coins at a given address within state.
+// It presumes a notion of sequence numbers for replay protection,
+// a notion of account numbers for replay protection for previously pruned accounts,
+// and a pubkey for authentication purposes.
+//
+// Many complex conditions can be used in the concrete struct which implements Account.
+type Account interface {
+	// GetAddress() sdk.AccAddress
```
alexanderbez, line 13: Why do we need these commented out bits?

x/validator-vesting/legacy/v0_3/types.go
```diff
@@ -0,0 +1,259 @@
...
+	*authvestingtypes.PeriodicVestingAccount
+	ValidatorAddress       sdk.ConsAddress       `json:"validator_address" yaml:"validator_address"`
+	ReturnAddress          sdk.AccAddress        `json:"return_address" yaml:"return_address"`
+	SigningThreshold       int64                 `json:"signing_threshold" yaml:"signing_threshold"`
+	CurrentPeriodProgress  CurrentPeriodProgress `json:"current_period_progress" yaml:"current_period_progress"`
+	VestingPeriodProgress  []VestingProgress     `json:"vesting_period_progress" yaml:"vesting_period_progress"`
+	DebtAfterFailedVesting sdk.Coins             `json:"debt_after_failed_vesting" yaml:"debt_after_failed_vesting"`
+}
+
+// // TotalCirculatingSupply represents the total circulating supply of Kava
```
alexanderbez, line 73: ditto

migrate/cmd.go
rhuairahrighairidh: yeah. I started copying how the sdk does migrations, but I think migrating between multiple versions would almost never be used by us or others. That and it would add a big maintenance burden to make sure all the old types are still compatible after upgrading dependencies.

migrate/doc.go
rhuairahrighairidh: cool, live upgrade is a better name

x/cdp/genesis.go
```diff
@@ -80,7 +80,10 @@ func InitGenesis(ctx sdk.Context, k Keeper, pk types.PricefeedKeeper, sk types.S
 	k.SetNextCdpID(ctx, gs.StartingCdpID)
 	k.SetDebtDenom(ctx, gs.DebtDenom)
 	k.SetGovDenom(ctx, gs.GovDenom)
-	k.SetPreviousSavingsDistribution(ctx, gs.PreviousDistributionTime)
+	// only set the previous block time if it's different than default
```
fedekunze, line 83: why is that?

x/cdp/keeper/savings_test.go
```diff
@@ -69,7 +69,7 @@ func (suite *SavingsTestSuite) TestGetSetPreviousDistributionTime() {
 	now := tmtime.Now()
 
 	_, f := suite.keeper.GetPreviousSavingsDistribution(suite.ctx)
-	suite.True(f)
+	suite.False(f) // distr time not set at genesis when the default genesis is used
```
fedekunze, line 72:
```suggestion
	suite.Require().False(f) // distr time not set at genesis when the default genesis is used
```

migrate/v0_8/tendermint/v0_32/genesis.go
```diff
@@ -0,0 +1,96 @@
...
+	ConsensusParams *ConsensusParams         `json:"consensus_params,omitempty"`
+	Validators      []types.GenesisValidator `json:"validators,omitempty"` // v0.33 GenesisValidator is backwards compatible with v0.32
+	AppHash         tmbytes.HexBytes         `json:"app_hash"`             // moved from `common` to `bytes` as they are the same between v0.32 and v0.33
+	AppState        json.RawMessage          `json:"app_state,omitempty"`
+}
+
+// ValidateAndComplete checks that all necessary fields are present
+// and fills in defaults for optional fields left empty
+func (genDoc *GenesisDoc) ValidateAndComplete() error {
+	if genDoc.ChainID == "" {
```
fedekunze, line 35: use `strings.TrimSpace`

contrib/kava-3/upgrade-instructions.md
```diff
@@ -0,0 +1,15 @@
+# Upgrade to kava-3
```
fedekunze, line 1: You can include this on the repo wiki 👍

contrib/kava-3/cmd.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/cosmos/cosmos-sdk/version"
+)
+
+const (
+	defaultChainID  = "kava-3"
+	flagGenesisTime = "genesis-time"
+	flagChainID     = "chain-id"
+)
+
+var defaultGenesisTime = ""
```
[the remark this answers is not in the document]
rhuairahrighairidh: I thought this approach might be less error prone, but yeah the init is a bit too weird

contrib/kava-3/params.go
```diff
@@ -0,0 +1,353 @@
...
+	))
+
+	appState[committee.ModuleName] = cdc.MustMarshalJSON(committee.NewGenesisState(
+		committee.DefaultNextProposalID,
+		[]committee.Committee{
+			committee.NewCommittee(
+				1,
+				"Kava Stability Committee",
+				[]sdk.AccAddress{
+					mustAccAddressFromBech32("kava1gru35up50ql2wxhegr880qy6ynl63ujlv8gum2"),
```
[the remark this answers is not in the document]
rhuairahrighairidh: yup, they're the addresses from the current gov vote. I added a comment to say so.

contrib/kava-3/upgrade-instructions.md
rhuairahrighairidh: yeah I think in the repo isn't the place for it

contrib/kava-3/upgrade-instructions.md
```diff
@@ -0,0 +1,15 @@
...
+1. `kvd export --for-zero-height > export-genesis.json`
+1. get new kava version:
+    1. in kava folder: `git pull`, `git checkout v0.8.0`, `make install`
+    1. check versions with `kvd version --long` and `kvcli version -long`
+1. `kvcli keys migrate`
+1. `kvd migrate export-genesis.json > migrated-genesis.json`
+1. `kvd write-params migrated-genesis.json --chain-id kava-3 --genesis-time 2020-06-01T14:00:00Z > genesis.json`
+1. check genesis file:
+    1. `kvd validate-genesis genesis.json` should say it's valid
+    1. `shasum -a 256 genesis.json` should output _
```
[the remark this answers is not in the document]
rhuairahrighairidh: I removed this file in favour of instructions being outside the repo

migrate/v0_8/tendermint/v0_32/genesis.go
rhuairahrighairidh: This file is copied in from v0.32 tendermint. My plan was to leave it as close as possible to the original to reduce chance of adding any bugs

karzak:
FYI, need a rebase now that #531 is merged

contrib/kava-3/params.go
```diff
@@ -0,0 +1,353 @@
...
+	genDoc.GenesisTime = genesisTime
+
+	// Add app params
+
+	var appState genutil.AppMap
+	if err := cdc.UnmarshalJSON(genDoc.AppState, &appState); err != nil {
+		return tmtypes.GenesisDoc{}, err
+	}
+
+	appState[auction.ModuleName] = cdc.MustMarshalJSON(auction.NewGenesisState(
```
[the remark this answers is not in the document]
rhuairahrighairidh: good idea

x/cdp/genesis.go
rhuairahrighairidh: If the previous block time in genesis.json is 0, then we don't set it here (in InitGenesis) so that it can be set in the begin blocker to the time of the first block.

I guess if previous block time is 0 in genesis.json we could just set it the ctx.BlockTime in InitGenesis. But I'm not sure the difference between the first block time and genesis time.

contrib/kava-3/params.go
```diff
@@ -0,0 +1,383 @@
...
+	genDoc.ChainID = chainID
+	genDoc.GenesisTime = genesisTime
+
+	// Add app params
+
+	var appState genutil.AppMap
+	if err := cdc.UnmarshalJSON(genDoc.AppState, &appState); err != nil {
+		return tmtypes.GenesisDoc{}, err
+	}
+
```
karzak, line 46: we need to add an `addAuthState` function that adds the bep3 deputy to state and populates it's account with the bnb that can be used for atomic swaps.

Issue #492, opened 2020-05-07
Update YAML marshaling/printing


denalimarsh:
nit, we should probably Marshal to YAML and print the string. We can do it in another PR for all types

_Originally posted by @fedekunze in https://github.com/Kava-Labs/kava-private/pull/7_

denalimarsh:
@rhuairahrighairidh noticed that the Auction module registers interfaces/types (Auction, GenesisAuction) on the codec while CDP, BEP3, and others only register msgs. To adhere to a global standard for marshaling/unmarshaling + `String` methods I'm pretty sure the other modules will need to follow the codec registration pattern in Auction, correct?

PR #519, opened 2020-05-21
x/incentive: genesis validation

fedekunze:


x/incentive/types/genesis.go
```diff
@@ -12,9 +15,35 @@ type GenesisClaimPeriodID struct {
...
+
 // GenesisClaimPeriodIDs array of GenesisClaimPeriodID
 type GenesisClaimPeriodIDs []GenesisClaimPeriodID
 
+// Validate checks if all the GenesisClaimPeriodIDs are valid and there are no duplicated
+// entries.
+func (gcps GenesisClaimPeriodIDs) Validate() error {
+	seenIDS := make(map[string]bool)
+	for _, gcp := range gcps {
+		if seenIDS[gcp.Denom+string(gcp.ID)] {
```
alexanderbez, line 34: nit: `gcp.Denom+string(gcp.ID)` is used twice, so I'd recommend storing this into a variable.

x/incentive/types/params.go
```diff
@@ -154,9 +124,47 @@ func (r Reward) String() string {
...
+// Validate performs a basic check of a reward fields.
+func (r Reward) Validate() error {
+	if !r.AvailableRewards.IsValid() {
+		return fmt.Errorf("invalid reward coins %s for %s", r.AvailableRewards, r.Denom)
+	}
+	if !r.AvailableRewards.IsPositive() {
+		return fmt.Errorf("reward amount must be positive, is %s for %s", r.AvailableRewards, r.Denom)
+	}
+	if r.Duration <= 0 {
+		return fmt.Errorf("reward duration must be positive, is %s for %s", r.Duration.String(), r.Denom)
```
alexanderbez, line 136: nit: you don't need to use `String()` when you pass the `%s` format

x/incentive/types/rewards.go
```diff
@@ -72,9 +104,51 @@ func NewClaimPeriod(denom string, id uint64, end time.Time, timeLock time.Durati
...
+	}
+	if cp.TimeLock == 0 {
+		return errors.New("claim period time lock cannot be 0")
+	}
+	return sdk.ValidateDenom(cp.Denom)
+}
+
+// String implements fmt.Stringer
+func (cp ClaimPeriod) String() string {
+	return fmt.Sprintf(`Claim Period:
```
alexanderbez, line 123: You can (should?) just use YAML here to make this cleaner.

x/incentive/types/rewards.go
```diff
@@ -72,9 +104,51 @@ func NewClaimPeriod(denom string, id uint64, end time.Time, timeLock time.Durati
...
+
 // ClaimPeriods array of ClaimPeriod
 type ClaimPeriods []ClaimPeriod
 
+// Validate checks if all the ClaimPeriods are valid and there are no duplicated
+// entries.
+func (cps ClaimPeriods) Validate() error {
+	seenPeriods := make(map[string]bool)
+	for _, cp := range cps {
+		if seenPeriods[cp.Denom+string(cp.ID)] {
```
alexanderbez, line 139: ditto

x/incentive/types/rewards.go
```diff
@@ -106,6 +194,24 @@ func (c Claim) String() string {
 // Claims array of Claim
 type Claims []Claim
 
+// Validate checks if all the claims are valid and there are no duplicated
+// entries.
+func (cs Claims) Validate() error {
+	seemClaims := make(map[string]bool)
+	for _, c := range cs {
+		if c.Owner != nil && seemClaims[c.Denom+string(c.ClaimPeriodID)+c.Owner.String()] {
```
alexanderbez, line 202: ditto

x/incentive/types/rewards.go
fedekunze: Agree, there's an issue for that #492

alexanderbez: But it's a code change you're adding here. It'd be one thing of the code already existed.

fedekunze: I just moved it. But I can update it to use yaml 👍

Issue #521, opened 2020-05-21
Pricefeed client/cli panics on postprice if expiry time is outside expected range 


denalimarsh:
`kvcli tx pricefeed postprice bnb:usd 1.001 99999999999999 --from validator`:

```
panic: Time.MarshalJSON: year outside of range [0,9999]

goroutine 1 [running]:
github.com/tendermint/go-amino.(*Codec).MustMarshalJSON(...)
        /Users/denali/go/pkg/mod/github.com/tendermint/go-amino@v0.15.1/amino.go:406
github.com/cosmos/cosmos-sdk/x/auth/client/utils.CompleteAndBroadcastTxCLI(0xc000bbdb80, 0x50da3e0, 0xc000357ae0, 0x0, 0x1, 0x30d40, 0x3ff0000000000000, 0x0, 0xc000f28550, 0x7, ...)
        /Users/denali/go/pkg/mod/github.com/cosmos/cosmos-sdk@v0.38.4/x/auth/client/utils/tx.go:83 +0xba1
github.com/cosmos/cosmos-sdk/x/auth/client/utils.GenerateOrBroadcastMsgs(0xc000e3b900, 0x14, 0x20, 0x50e0160, 0xc000e38f00, 0xc000f28550, 0x7, 0x0, 0x0, 0x50a3a80, ...)
        /Users/denali/go/pkg/mod/github.com/cosmos/cosmos-sdk@v0.38.4/x/auth/client/utils/tx.go:40 +0x15c
github.com/kava-labs/kava/x/pricefeed/client/cli.GetCmdPostPrice.func1(0xc000edb340, 0xc000e63270, 0x3, 0x5, 0x0, 0x0)
        /Users/denali/go/src/github.com/kava-labs/kava/x/pricefeed/client/cli/tx.go:66 +0x4e9
github.com/spf13/cobra.(*Command).execute(0xc000edb340, 0xc000e63220, 0x5, 0x5, 0xc000edb340, 0xc000e63220)
        /Users/denali/go/pkg/mod/github.com/spf13/cobra@v0.0.6/command.go:840 +0x460
github.com/spf13/cobra.(*Command).ExecuteC(0xc00016a840, 0x4490350, 0xc00016a840, 0x4d3ba1d)
        /Users/denali/go/pkg/mod/github.com/spf13/cobra@v0.0.6/command.go:945 +0x317
github.com/spf13/cobra.(*Command).Execute(...)
[...]

PR #532, opened 2020-05-29
Fix expiry parsing

alexanderbez:
closes: #521

### DIFF
--- a/x/pricefeed/client/cli/tx.go
+++ b/x/pricefeed/client/cli/tx.go
@@ -3,6 +3,7 @@ package cli
 import (
 	"bufio"
 	"fmt"
+	"strconv"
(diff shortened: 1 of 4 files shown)


x/pricefeed/client/cli/tx.go
```diff
@@ -52,17 +55,23 @@ func GetCmdPostPrice(cdc *codec.Codec) *cobra.Command {
 			if err != nil {
 				return err
 			}
-			expiryInt, ok := sdk.NewIntFromString(args[2])
```
alexanderbez, line 55: note to author: no need to use `NewIntFromString` when you don't need an `sdk.Int` -- better to use the stdlib.

PR #553, opened 2020-06-05
cleanup begin/end block

fedekunze:


x/auction/abci.go
```diff
@@ -4,7 +4,8 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
-// BeginBlocker runs at the start of every block.
+// BeginBlocker closes all expired auctions at the end of each block. It panics if
+// there's an error other than ErrAuctionNotFound.
```
karzak, line 8: Given that ErrAuctionNotFound shouldn't happen, any reason not to panic? cc @rhuairahrighairidh

x/cdp/keeper/savings.go
```diff
@@ -27,37 +28,46 @@ func (k Keeper) DistributeSavingsRate(ctx sdk.Context, debtDenom string) error {
 	modAccountCoins := k.getModuleAccountCoins(ctx, dp.Denom)
 	totalSupplyLessModAccounts := k.supplyKeeper.GetSupply(ctx).GetTotal().Sub(modAccountCoins)
 	surplusDistributed := sdk.ZeroInt()
+
 	var iterationErr error
+	// TODO: avoid iterating over all the accounts by keeping the storing stable coin
+	// holders' addresses separately.
```
karzak, line 34: +1 on leaving TODO here. I believe v0.39 of sdk will store balances such that you can iterate over only accounts that hold a balance of a particular coin type.

x/pricefeed/keeper/keeper.go
```diff
@@ -111,40 +119,38 @@ func (k Keeper) SetCurrentPrices(ctx sdk.Context, marketID string) error {
...
 		}
 	}
+
 	if len(notExpiredPrices) == 0 {
-		store := ctx.KVStore(k.key)
-		store.Set(
-			types.CurrentPriceKey(marketID), k.cdc.MustMarshalBinaryBare(types.CurrentPrice{}),
-		)
+		// TODO: why do we set an empty value?
+		k.setCurrentPrice(ctx, marketID, types.CurrentPrice{})
```
karzak, line 125: The current price stored will continue storing the most recent (expired) price if this in not set. This zero's out the current price stored value for that market and ensures that CDP methods that `GetCurrentPrice` will return error.

x/pricefeed/abci.go
```diff
@@ -10,19 +10,14 @@ import (
...
-				)
-				continue
-			}
+		if !market.Active {
+			continue
+		}
+
+		err := k.SetCurrentPrices(ctx, market.MarketID)
+		if err != nil {
+			// TODO: this should panic
```
karzak, line 19: If we want to panic here, we should check that the error is not `ErrNoValidPrice`. All oracles going down is bad, but shouldn't cause an application panic.

x/auction/keeper/auctions.go
```diff
@@ -530,29 +523,28 @@ func (k Keeper) PayoutCollateralAuction(ctx sdk.Context, a types.CollateralAucti
...
+		return nil
 	}
-	return nil
+
+	return k.supplyKeeper.SendCoinsFromModuleToModule(ctx, types.ModuleName, a.Initiator, sdk.NewCoins(a.CorrespondingDebt))
 }
 
 // CloseExpiredAuctions finds all auctions that are past (or at) their ending times and closes them, paying out to the highest bidder.
 func (k Keeper) CloseExpiredAuctions(ctx sdk.Context) error {
-	var expiredAuctions []uint64
```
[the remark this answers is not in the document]
fedekunze: I added a more comprehensive comment, let me know if it's better now

PR #567, opened 2020-06-08
Exhibit 13: Update GetTotalSurplus and GetTotalDebt to be consistent

nddeluca:
Changes:
  - Add explicit tests for GetTotalSurplus/Debt functions
  - Update GetTotalSurplus and GetTotalDebt to have consistent implementations

### DIFF
--- a/x/cdp/keeper/auctions_test.go
+++ b/x/cdp/keeper/auctions_test.go
@@ -111,7 +111,7 @@ func (suite *AuctionTestSuite) TestGetTotalSurplus() {
 	err = sk.MintCoins(suite.ctx, types.LiquidatorMacc, cs(c("debt", 500e6)))
 	suite.NoError(err)
 
-	// liquidator account still has 300e6 total usdx -- debt balance is ingored
+	// liquidator account still has 300e6 total usdx -- debt balance is ignored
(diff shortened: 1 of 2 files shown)


x/cdp/keeper/auctions_test.go
```diff
@@ -92,6 +92,66 @@ func (suite *AuctionTestSuite) TestDebtAuction() {
 	suite.Equal(cs(c("debt", 90000000000)), acc.GetCoins())
 }
 
+func (suite *AuctionTestSuite) TestGetTotalSurplus() {
+	sk := suite.app.GetSupplyKeeper()
+
+	// liquidator account has zero coins
+	suite.Equal(sdk.NewInt(0), suite.keeper.GetTotalSurplus(suite.ctx, types.LiquidatorMacc))
```
fedekunze, line 99: we should use a require context for the testing suite since the default is an assertion. Applies to all the new tests 👍 

```suggestion
	suite.Require().Equal(sdk.NewInt(0), suite.keeper.GetTotalSurplus(suite.ctx, types.LiquidatorMacc))
```

nddeluca: awesome, great tip 💯.  Added suite.Require() to rest of new tests

PR #574, opened 2020-06-10
Auction Invariant - Check that stored auctions end times after block time

nddeluca:
Updates if statement to match validation error.

Fixes exhibit 48.


nddeluca:
Closing, logic is correct and error message is wrong.

PR #581, opened 2020-06-11
feat: gentx for HubsValidator

hubsen1980:


karzak:
The kava mainnet launched in November - there is no gentx submission process ongoing. To run a validator on mainnet, see https://medium.com/kava-labs/validating-kava-mainnet-72fa1b6ea579

PR #585, opened 2020-06-12
Pricefeed price simulation refactor

nddeluca:
Refactors the pricefeed simulation to use PriceGenerator object that deterministically randomizes price movements.

### DIFF
--- a/x/pricefeed/simulation/operations.go
+++ b/x/pricefeed/simulation/operations.go
@@ -2,7 +2,6 @@ package simulation
 
 import (
 	"math/rand"
-	"sort"
 	"time"
 
 	"github.com/cosmos/cosmos-sdk/baseapp"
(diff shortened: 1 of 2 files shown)


nddeluca:
Thanks Denali! The goal here is to just make it deterministic, agree we need to improve the problem cases and add in more adversarial cases to the price changes. Added a card to the hopper. Going to mark this PR as ready for review

x/pricefeed/simulation/operations.go
```diff
@@ -198,3 +129,93 @@ func getExpiryTime(ctx sdk.Context) (t time.Time) {
 	// need to use the blocktime from the context as the context generates random start time when running simulations
 	return ctx.BlockTime().Add(time.Second * 1000000)
 }
+
+// PriceGenerator allows deterministic price generation in simulations
+type PriceGenerator struct {
```
fedekunze, line 134: I'd move this struct and its functions to a new file `x/pricefeed/simulation/types.go` to keep logic clean from type definitions 👍

karzak: 👍  to moving the type out

nddeluca: Great idea 💯

PR #586, opened 2020-06-12
[R4R] Restore supply endpoints

karzak:
Adds 
```
/vesting/circulatingsupply
/vesting/totalsupply
```

back to client (used by coin information sites like coin gecko).


denalimarsh:
@karzak looks like there's some duplicated code from https://github.com/Kava-Labs/kava/pull/584, cleaning it up would make reviewing these changes easier.

x/validator-vesting/client/rest/query.go
```diff
@@ -0,0 +1,102 @@
...
+
+	"github.com/cosmos/cosmos-sdk/client/context"
+	"github.com/cosmos/cosmos-sdk/types/rest"
+
+	"github.com/kava-labs/kava/x/validator-vesting/types"
+)
+
+func registerQueryRoutes(cliCtx context.CLIContext, r *mux.Router) {
+	r.HandleFunc("/vesting/circulatingsupply", getCirculatingSupplyHandlerFn(cliCtx)).Methods("GET")
+	r.HandleFunc("/vesting/totalsupply", getTotalSupplyHandlerFn(cliCtx)).Methods("GET")
```
denalimarsh, line 18: An explicit route variable would be clearer, declare const "vesting". May avoid some confusion as the other modules all use `/{ModuleName}/path/to/endpoint`, this is the only one that doesn't use the module name.

x/validator-vesting/client/cli/query.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/cosmos/cosmos-sdk/codec"
+
+	"github.com/kava-labs/kava/x/validator-vesting/types"
+)
+
+// GetQueryCmd returns the cli query commands for the kavadist module
+func GetQueryCmd(queryRoute string, cdc *codec.Codec) *cobra.Command {
+	valVestingQueryCmd := &cobra.Command{
+		Use:   types.ModuleName,
```
denalimarsh, line 18: Can we update `types.ModuleName` to `vesting` so it matches the REST api? It looks like it won't conflict with "vesting" in the latest version of cosmos-sdk, but not 100% sure here.

karzak: That would be a breaking change, so can't update it as part of the client release.

denalimarsh: Ah, my comment wasn't clear - I meant adding a new route variable to use instead of `types.ModuleName`

denalimarsh: resolved by https://github.com/Kava-Labs/kava/pull/586/commits/12a904214881a7ddf80ebd465c7f49674aaba68e

x/validator-vesting/client/rest/query.go
denalimarsh: resolved by https://github.com/Kava-Labs/kava/pull/586/commits/12a904214881a7ddf80ebd465c7f49674aaba68e

PR #591, opened 2020-06-17
Add raw params querier

rhuairahrighairidh:
Add a querier and cli method to retrieve raw param `[]byte` values.

This makes it easier to compare current param values with governance proposals and verify there are no mistakes. Should be useful if we need to create proposals quickly in emergency scenarios.

Ideally this would live in the sdk params module, but we don't have access, so I put it in the committee module as it has access to the params keeper.

Could add a rest endpoint, but I couldn't think of a use case, so left it just as a cli cmd.


denalimarsh:
Fix conflicts before merge. With each PR now containing a changelog entry we're unfortunately gonna have a lot of these conflicts in CHANGELOG.md.

karzak:
> Fix conflicts before merge. With each PR now containing a changelog entry we're unfortunately gonna have a lot of these conflicts in CHANGELOG.md.

To avoid this, the merging author/reviewer should update the changelog right before merging.

PR #593, opened 2020-06-18
Update BEP3 CLI swap & swaps command to use struct instead of string

nddeluca:
Previously, the `-o json` flag did not work for `kvcli q bep3 swap` and `kvcli q bep3 swaps` commands to due passing a string to `PrintOutput`.

This PR adds a decorating struct to pass to `PrintOutput` in order to support both json & yaml, and provide a similar output to the previous string format.

### DIFF
--- a/x/bep3/client/cli/query.go
+++ b/x/bep3/client/cli/query.go
@@ -26,6 +26,40 @@ const (
 	flagDirection  = "direction"
 )
 
+type FormattedAtomicSwap struct {
+	ID                  string `json:"ID" yaml:"ID"`
+	Amount              string `json:"amount"  yaml:"amount"`
+	RandomNumberHash    string `json:"random_number_hash"  yaml:"random_number_hash"`
+	ExpireHeight        uint64 `json:"expire_height"  yaml:"expire_height"`
+	Timestamp           int64  `json:"timestamp"  yaml:"timestamp"`
+	Sender              string `json:"sender"  yaml:"sender"`
+	Recipient           string `json:"recipient"  yaml:"recipient"`
+	SenderOtherChain    string `json:"sender_other_chain"  yaml:"sender_other_chain"`
+	RecipientOtherChain string `json:"recipient_other_chain"  yaml:"recipient_other_chain"`
+	ClosedBlock         int64  `json:"closed_block"  yaml:"closed_block"`
+	Status              string `json:"status"  yaml:"status"`
+	CrossChain          bool   `json:"cross_chain"  yaml:"cross_chain"`
+	Direction           string `json:"direction"  yaml:"direction"`
+}
+
+func NewFormattedAtomicSwap(swap types.AtomicSwap) FormattedAtomicSwap {
+	return FormattedAtomicSwap{
+		ID:                  swap.GetSwapID().String(),
+		Amount:              swap.Amount.String(),
+		RandomNumberHash:    hex.EncodeToString(swap.RandomNumberHash),
+		ExpireHeight:        swap.ExpireHeight,
+		Timestamp:           swap.Timestamp,
+		Sender:              swap.Sender.String(),
+		Recipient:           swap.Recipient.String(),
+		SenderOtherChain:    swap.SenderOtherChain,
+		RecipientOtherChain: swap.RecipientOtherChain,
+		ClosedBlock:         swap.ClosedBlock,
+		Status:              swap.Status.String(),
+		CrossChain:          swap.CrossChain,
+		Direction:           swap.Direction.String(),
+	}
+}
+
 // GetQueryCmd returns the cli query commands for this module
 func GetQueryCmd(queryRoute string, cdc *codec.Codec) *cobra.Command {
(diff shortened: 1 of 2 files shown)


x/bep3/client/cli/query.go
```diff
@@ -26,6 +26,40 @@ const (
 	flagDirection  = "direction"
 )
 
+type FormattedAtomicSwap struct {
```
nddeluca, line 29: Should this go in `x/bep3/types/swap.go`?  Is there a better existing pattern for decorating cli output?

x/bep3/client/cli/query.go
```diff
@@ -293,8 +327,14 @@ $ kvcli q bep3 swaps --page=2 --limit=100
...
 
+			formattedAtomicSwaps := []FormattedAtomicSwap{}
+
+			for _, swap := range matchingAtomicSwaps {
+				formattedAtomicSwaps = append(formattedAtomicSwaps, NewFormattedAtomicSwap(swap))
+			}
+
 			cliCtx = cliCtx.WithHeight(height)
-			return cliCtx.PrintOutput(matchingAtomicSwaps.String()) // nolint:errcheck
+			return cliCtx.PrintOutput(formattedAtomicSwaps)
```
nddeluca, line 337: I just matched the existing output of the CLI command, but it doesn't have parity with the REST API.

Are we OK with that?

Example:
CLI YAML
```
ID: FD6C531A2E0B8AF67030E59D0ED2424E3380950DB7444FC1CEB30D8369F83057
amount: 19999000bnb
random_number_hash: 1b24c223bb9e59385416dbcf71639efdb401dd9a14159c57ee704a8fd766b43b
expire_height: 72998
timestamp: 1592285967
sender: kava1r4v2zdhdalfj2ydazallqvrus9fkphmglhn6u6
recipient: kava1gzc54u4p67hh9r4m9vcml3ke9fc29tplvvaev3
sender_other_chain: bnb1nv2fnfc6xvvve2qvtzg2x5vp4h6wxt243d4ddl
recipient_other_chain: bnb1jh7uv2rm6339yue8k4mj9406k3509kr4wt5nxn
closed_block: 76556
status: Completed
cross_chain: true
direction: Incoming
```
CLI JSON
```
{
    "ID": "FD6C531A2E0B8AF67030E59D0ED2424E3380950DB7444FC1CEB30D8369F83057",
    "amount": "19999000bnb",
    "random_number_hash": "1b24c223bb9e59385416dbcf71639efdb401dd9a14159c57ee704a8fd766b43b",
    "expire_height": "72998",
    "timestamp": "1592285967",
    "sender": "kava1r4v2zdhdalfj2ydazallqvrus9fkphmglhn6u6",
    "recipient": "kava1gzc54u4p67hh9r4m9vcml3ke9fc29tplvvaev3",
    "sender_other_chain": "bnb1nv2fnfc6xvvve2qvtzg2x5vp4h6wxt243d4ddl",
    "recipient_other_chain": "bnb1jh7uv2rm6339yue8k4mj9406k3509kr4wt5nxn",
    "closed_block": "76556",
    "status": "Completed",
    "cross_chain": true,
    "direction": "Incoming"
}
```
REST JSON
```
{
      "amount": [
        {
          "denom": "bnb",
          "amount": "9999000"
        }
      ],
      "random_number_hash": "6AB6AC1A5DB9CB2EDF782D1D5D2076A9BBCCC663F8CB8D8CE11366778E77B74C",
      "expire_height": "59101",
      "timestamp": "1592193199",
      "sender": "kava1r4v2zdhdalfj2ydazallqvrus9fkphmglhn6u6",
      "recipient": "kava1qa5tpxdy9r636nt2ze76g8k5h6zce807z4tksg",
      "sender_other_chain": "bnb17sd249m4mtr66n3ux057rvjs0c09sqf5rvq39n",
      "recipient_other_chain": "bnb1jh7uv2rm6339yue8k4mj9406k3509kr4wt5nxn",
      "closed_block": "58982",
      "status": "Completed",
      "cross_chain": true,
     "direction": "Incoming"
}
```

nddeluca, line 337: I just matched the existing output of the CLI command, but it doesn't have parity with the REST API.

Are we OK with that?  Should we just return the existing AtomicSwap struct in the cli?

The biggest downside might be the `ID` isn't included and the random number hash is a yaml array instead of a hex string.

nddeluca: Makes sense, thanks for sharing thoughts on this.  I'm going to update the command and rest api to match.

x/bep3/client/cli/query.go
nddeluca: Agree, it seemed out of place outside of client.  The naming is tough.

Updating the cli and rest client to match will let me push the new interface into the `quierier`, which I think solves some of the ugliness of declaring the type in the client.

AugmentedAtomicSwap sounds better in that context, and likely better to stick with a single pattern.  Thanks for the example

nddeluca:
Closing in favor of #597

PR #594, opened 2020-06-18
Update `kvcli tx bep3 create -h` with valid height span

nddeluca:
Changes example height span to 270 from 360.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,8 +36,6 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 ## [Unreleased]
 
-[\#594](https://github.com/Kava-Labs/kava/pull/594) Update example CLI bep3 create help text with valid height span
-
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -36,6 +36,8 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 ## [Unreleased]
 
+[\#594](https://github.com/Kava-Labs/kava/pull/594) Update example CLI bep3 create help text with valid height span
```
denalimarsh, line 39: I don't think this warrants a changelog entry. It's good to start using the changelog as we've been lazy with it in the past, but I think we should guard against the pendulum swinging back the other way and cluttering it with entries for small changes.

Open to a discussion about this as I don't believe we have a standard for what warrants a changelog entry.

nddeluca: 👍 I was thinking similar as I was writing the changelog entry. Updated without the changelog addition.

PR #595, opened 2020-06-19
Add Addition CDP Client Methods

nddeluca:
- Query total usdx supply
- Query module account addresses


x/cdp/client/cli/query.go
```diff
@@ -225,3 +227,55 @@ func QueryParamsCmd(queryRoute string, cdc *codec.Codec) *cobra.Command {
 		},
 	}
 }
+
+func QueryTotalSupply(queryRoute string, cdc *codec.Codec) *cobra.Command {
```
fedekunze, line 231: you can already do this with the supply module 
`kavacli q supply total usdx`

nddeluca: 👍 Will remove the CDP specific query

nddeluca:
Closing to do some rework.  Addresses was the wrong direction, we want to return the full module account info

PR #597, opened 2020-06-19
Update BEB3 CLI Client to Support JSON

nddeluca:
- **(Breaking)** The YAML output matches the REST JSON now.  Some field representations have changed.
- The `id` field was added to the REST API Output.
  - This was already in the swagger definition, so no change there.
- JSON CLI output is now fixed and also matches the REST JSON.

### DIFF
--- a/x/bep3/types/swap.go
+++ b/x/bep3/types/swap.go
@@ -276,3 +276,5 @@ func NewAugmentedAtomicSwap(swap AtomicSwap) AugmentedAtomicSwap {
 		AtomicSwap: swap,
 	}
 }
+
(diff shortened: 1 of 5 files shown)


x/bep3/types/swap.go
```diff
@@ -264,3 +264,42 @@ func (direction SwapDirection) IsValid() bool {
 	}
 	return false
 }
+
+type AugmentedAtomicSwap struct {
```
nddeluca, line 268: I matched this naming with `AugmentedCDP` with idea that one pattern is better than two patterns of decorating output.

Alternatively, this could go be named `QueryAtomicSwapResponse` in `types/querier.go`.

PR #599, opened 2020-06-20
Issuance module

karzak:
- [x] Spec
- [x] Types and Keeper
- [x] Module
- [x] BeginBlock
- [x] Client
- [x] Tests
- [x] Sims

Note:  The v0.3 -> v0.8 migration tests will need to be updated with a custom app type for `v0.8`, then that type should be used in the migration tests. Out of scope for this PR.


x/issuance/spec/01_concepts.md
```diff
@@ -0,0 +1,7 @@
+<!--
+order: 1
+-->
+
+# Concepts
+
+The issuance mechanism in this module is designed to allow a trusted party to issue an asset on to the Kava blockchain. The issuer has sole discretion over the minting and redemption (burning) of the asset, as well as restricting access to the asset via transfer-restriction and asset seizure. The functionality of this module is similar to that of ERC-20 contracts for stablecoins that have a single issuer.
```
denalimarsh, line 7:
```suggestion
The issuance mechanism in this module is designed to allow a trusted party to issue an asset on to the Kava blockchain. The issuer has sole discretion over the minting and redemption (burning) of the asset, as well as restricting access to the asset via asset seizure. The functionality of this module is similar to that of ERC-20 contracts for stablecoins that have a single issuer.
```

Until we figure out a way to restrict transfers

x/issuance/spec/03_messages.md
```diff
@@ -0,0 +1,71 @@
...
+	BlockedAddress sdk.AccAddress `json:"blocked_address" yaml:"blocked_address"`
+}
+```
+
+## State Modifications
+
+* The address is added to the block list, which prevents the account from holding coins of that denom
+* Tokens are sent back to the issuer
+
+```
```
denalimarsh, line 54:
```suggestion
```

x/issuance/spec/05_params.md
```diff
@@ -0,0 +1,22 @@
+
```
denalimarsh, line 1:
```suggestion
```

x/issuance/spec/05_params.md
```diff
@@ -0,0 +1,22 @@
...
+order: 5
+-->
+
+# Parameters
+
+The issuance module has the following parameters:
+
+| Key        | Type           | Example       | Description                                 |
+|------------|----------------|---------------|---------------------------------------------|
+| Assets     | array (Asset)  | [{see below}] | array of assets created via issuance module |
```
denalimarsh, line 12:
```suggestion
| Assets     | array (Asset)  | `[{see below}]` | array of assets created via issuance module |
```
This may need backticks for proper display in docs, might be safe since it's inside brackets

x/issuance/spec/README.md
```diff
@@ -0,0 +1,20 @@
+<!--
+order: 0
+title: "Issuance Overview"
+parent:
+  title: "issuance"
```
denalimarsh, line 5: +1 for spec headers 💯

x/issuance/spec/README.md
```diff
@@ -0,0 +1,20 @@
...
+1. **[Concepts](01_concepts.md)**
+2. **[State](02_state.md)**
+3. **[Messages](03_messages.md)**
+4. **[Events](04_events.md)**
+5. **[Params](05_params.md)**
+6. **[BeginBlock](06_begin_block.md)**
+
+## Abstract
+
+`x/issuance` is an implementation of a Cosmos SDK Module that allows for an issuer to control the minting and burning of an asset. The issuer is a white-listed address, which may be a multisig address, that can mint and burn coins of a particular denomination. The issuer is the only entity that can mint or burn tokens of that asset type (ie, there is no outside inflation or deflation). Additionally, the issuer can place transfer-restrictions on addresses, which prevents an addresses from transferring and holding tokens with the issued denomination. Coins are minted at burned at the discretion of the issuer, using `MsgIssueTokens` and `MsgRedeemTokens` transaction types, respectively.
```
denalimarsh, line 20: "ie" -> "i.e."
"which prevents an addresses" -> "which prevents addresses"
"Coins are minted at burned" -> "Coins are minted and burned"

x/issuance/types/genesis_test.go
```diff
@@ -0,0 +1,142 @@
...
+				expectPass: false,
+				contains:   "invalid denom",
+			},
+		},
+		{
+			"duplicate asset",
+			args{
+				assets: types.Assets{
+					types.NewAsset(suite.addrs[0], "usdtoken", []sdk.AccAddress{suite.addrs[1]}, false),
+					types.NewAsset(suite.addrs[0], "usdtoken", []sdk.AccAddress{suite.addrs[1]}, false),
```
denalimarsh, line 117: add a test where only the denom is duplicated and the other fields are different.

x/issuance/client/cli/tx.go
```diff
@@ -0,0 +1,177 @@
...
+			if err != nil {
+				return err
+			}
+			var status bool
+			if args[1] == "true" {
+				status = true
+			} else if args[1] == "false" {
+				status = false
+			} else {
+				return fmt.Errorf(fmt.Sprintf("status must be true or flase, got %s", args[1]))
```
rhuairahrighairidh, line 166:
```suggestion
				return fmt.Errorf(fmt.Sprintf("status must be true or false, got %s", args[1]))
```

x/issuance/client/cli/tx.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+func getCmdRedeemTokens(cdc *codec.Codec) *cobra.Command {
+	return &cobra.Command{
+		Use:   "redeem [tokens]",
+		Short: "redeem tokens",
+		Long: strings.TrimSpace(
+			fmt.Sprintf(`The asset owner redeems (burns) tokens, removing them from the circulating supply
+
+			Example:
```
fedekunze, line 81: the `cobra.Command` has an `Example` field, I'd recommend using that instead 👍

x/issuance/client/cli/tx.go
```diff
@@ -0,0 +1,177 @@
...
+				return err
+			}
+			return utils.GenerateOrBroadcastMsgs(cliCtx, txBldr, []sdk.Msg{msg})
+		},
+	}
+}
+
+func getCmdPauseAsset(cdc *codec.Codec) *cobra.Command {
+	return &cobra.Command{
+		Use:   "pause [denom] [status]",
```
fedekunze, line 142: I don't understand the status of the denom being true or false. Maybe `toggle-status [denom]` or something like that?

x/issuance/genesis.go
```diff
@@ -0,0 +1,31 @@
...
+
+	"github.com/kava-labs/kava/x/issuance/keeper"
+	"github.com/kava-labs/kava/x/issuance/types"
+)
+
+// InitGenesis initializes the store state from a genesis state.
+func InitGenesis(ctx sdk.Context, k keeper.Keeper, supplyKeeper types.SupplyKeeper, gs types.GenesisState) {
+	if err := gs.Validate(); err != nil {
+		panic(fmt.Sprintf("failed to validate %s genesis state: %s", types.ModuleName, err))
+	}
```
fedekunze, line 16: imho there's no need for this if you have the genesis the validation function defined on module.go

x/issuance/keeper/issuance.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	// send to receiver
+	err = k.supplyKeeper.SendCoinsFromModuleToAccount(ctx, types.ModuleAccountName, receiver, sdk.NewCoins(tokens))
+	if err != nil {
+		return err
+	}
+	ctx.EventManager().EmitEvent(
+		sdk.NewEvent(
+			types.EventTypeIssue,
+			sdk.NewAttribute(types.AttributeKeyIssueAmount, fmt.Sprintf("%s", tokens)),
```
fedekunze, line 41:
```suggestion
			sdk.NewAttribute(types.AttributeKeyIssueAmount, tokens.String()),
```

x/issuance/keeper/issuance.go
```diff
@@ -0,0 +1,156 @@
...
+		return err
+	}
+	err = k.supplyKeeper.BurnCoins(ctx, types.ModuleAccountName, coins)
+	if err != nil {
+		return err
+	}
+	ctx.EventManager().EmitEvent(
+		sdk.NewEvent(
+			types.EventTypeRedeem,
+			sdk.NewAttribute(types.AttributeKeyRedeemAmount, fmt.Sprintf("%s", tokens)),
```
fedekunze, line 71:
```suggestion
			sdk.NewAttribute(types.AttributeKeyRedeemAmount, tokens.String()),
```

x/issuance/keeper/issuance.go
```diff
@@ -0,0 +1,156 @@
...
+		),
+	)
+	return nil
+}
+
+// BlockAddress adds an address to the blocked list
```
fedekunze, line 77: blocked list of? What does blocking mean in this context? preventing it from transferring?

x/issuance/keeper/issuance.go
```diff
@@ -0,0 +1,156 @@
...
+func (k Keeper) BlockAddress(ctx sdk.Context, denom string, owner, blockedAddress sdk.AccAddress) error {
+	asset, found := k.GetAsset(ctx, denom)
+	if !found {
+		return sdkerrors.Wrapf(types.ErrAssetNotFound, "denom: %s", denom)
+	}
+	if !owner.Equals(asset.Owner) {
+		return sdkerrors.Wrapf(types.ErrNotAuthorized, "owner: %s, address: %s", asset.Owner, owner)
+	}
+	if k.checkBlockedAddress(ctx, asset, blockedAddress) {
+		return nil
```
fedekunze, line 87: this should return an error if already blocked

x/issuance/keeper/issuance.go
```diff
@@ -0,0 +1,156 @@
...
+			types.EventTypeBlock,
+			sdk.NewAttribute(types.AttributeKeyBlock, fmt.Sprintf("%s", blockedAddress)),
+			sdk.NewAttribute(types.AttributeKeyDenom, asset.Denom),
+		),
+	)
+	return nil
+}
+
+// ChangePauseStatus pauses/un-pauses an asset
+func (k Keeper) ChangePauseStatus(ctx sdk.Context, owner sdk.AccAddress, denom string, status bool) error {
```
fedekunze, line 102: rename to ToggleAssetStatus?

x/issuance/keeper/issuance.go
```diff
@@ -0,0 +1,156 @@
...
+	if !found {
+		return sdkerrors.Wrapf(types.ErrAssetNotFound, "denom: %s", denom)
+	}
+	if !owner.Equals(asset.Owner) {
+		return sdkerrors.Wrapf(types.ErrNotAuthorized, "owner: %s, address: %s", asset.Owner, owner)
+	}
+	if asset.Paused == status {
+		return nil
+	}
+	asset.Paused = status
```
fedekunze, line 113:
```suggestion
	asset.Paused = !asset.Paused
```

x/issuance/keeper/issuance.go
```diff
@@ -0,0 +1,156 @@
...
+// SeizeCoinsFromBlockedAddresses checks blocked addresses for coins of the input denom and transfers them to the owner account
+func (k Keeper) SeizeCoinsFromBlockedAddresses(ctx sdk.Context, denom string) error {
+	asset, found := k.GetAsset(ctx, denom)
+	if !found {
+		return sdkerrors.Wrapf(types.ErrAssetNotFound, "denom: %s", denom)
+	}
+	for _, address := range asset.BlockedAddresses {
+		account := k.accountKeeper.GetAccount(ctx, address)
+		coinsAmount := account.GetCoins().AmountOf(denom)
+		if coinsAmount.IsPositive() {
```
fedekunze, line 134: use inverse `if !coinsAmount.IsPositive()`

x/issuance/types/errors.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+// DONTCOVER
+
+// Errors used by the issuance module
+var (
+	ErrAssetNotFound  = sdkerrors.Register(ModuleName, 1, "no asset with input denom found")
```
fedekunze, line 11: error codes should start from 2 as 1 is reserved for internal errors

x/issuance/types/genesis.go
```diff
@@ -0,0 +1,40 @@
+package types
+
+import "bytes"
+
+// GenesisState is the state that must be provided at genesis for the issuance module
+type GenesisState struct {
+	Params Params `json:"params" yaml:"params"`
```
fedekunze, line 7: blocked address are not persisted on genesis state?

x/issuance/types/params.go
```diff
@@ -0,0 +1,131 @@
...
+// Parameter keys and default values
+var (
+	KeyAssets         = []byte("Assets")
+	DefaultAssets     = Assets{}
+	ModuleAccountName = ModuleName
+)
+
+// Params governance parameters for the issuance module
+type Params struct {
+	Assets Assets `json:"assets" yaml:"assets"`
```
fedekunze, line 19: ok, so I think Assets should be governance params. They should be set to the issuance module store and defined on genesis. If a new asset needs to be added/removed you can create a custom gov proposal type for both operations and create a custom handler to add/delete assets from the store.

This way you have fine-grained control over the validation of the assets, eg validate that users cannot add and delete assets on the same param change proposal. Also, param change proposals don't allow you to validate the new value with the previously stored ones.

denalimarsh:
Created [an issue](https://github.com/cosmos/cosmos-sdk/issues/6518) on cosmos-sdk for enabling/disabling coin transfers of individual coins.

x/issuance/client/cli/tx.go
karzak: maybe `kvcli tx issuance set-pause-status usdtoken true/false`. Clearer?

x/issuance/keeper/issuance.go
karzak: The begin block seizes an coins of that denom held by that address and transfers them to owner

x/issuance/types/genesis.go
karzak: They are - blocked addresses is a property of asset, which is a property of params.

x/issuance/types/params.go
karzak: Assets already are goverance params because they are part of the `Params` for the module, so we can use `param-change` proposals to add/remove/edit assets. I think this is sufficient for governance for now, can expand if necessary as scope becomes better defined.

x/issuance/types/msg.go
```diff
@@ -0,0 +1,282 @@
...
+}
+
+// GetSigners returns the addresses of signers that must sign
+func (msg MsgUnblockAddress) GetSigners() []sdk.AccAddress {
+	return []sdk.AccAddress{msg.Sender}
+}
+
+// String implements fmt.Stringer
+func (msg MsgUnblockAddress) String() string {
+	return fmt.Sprintf(`Unblock Address:
```
fedekunze, line 225: nit we should use yaml

PR #606, opened 2020-07-01
Update developer tools

karzak:
Updates `kvd testnet` to allow for greater customization. Namely, uses `ukava` as the default staking denom and adds `--genesis-template` flag to allow creating local testnets with custom genesis state. 

Adds a guide in `contrib/README.md` for local testnet setup.


karzak:
Closing as this is better scoped outside this repo

PR #616, opened 2020-07-27
BEP3: support multiple assets

karzak:
* Refactors parameters so that all assets have their own distinct params
* Moves the asset supply tracking into the parameters
* Removes monkey-patch for updating supply limit
* Throws an error for if the deputy is both the sender and receiver of a swap (is there any reason to allow this?)

Build is currently breaking because v0.3 -> v0.8 migration break. We can discuss how we want to deal with that, but figured I would wait for review, as the PR is already quite large.

PR #623, opened 2020-08-10
Add rate limiting to bep3 assets

karzak:
For reviewers:
Please give final review to #616 before reviewing here. Any issues raised in that review and addressed here will be called out specifically. 

New Features that should be given thorough review. 
* Adds minting/burning from the bep3 module account, removing the need for the deputy address to hold coins. 
* Adds time-based rate limiting for incoming swaps. The time based limiting is optional and is set via params. The time passed for rate-limited assets is updated in the begin blocker. This begin-blocker logic also fixes the bug where assets added via governance would have no `AssetSupply`.

Issues from previous review fixed here:
* Bug where adding an asset via governance would have no `AssetSupply` in the keeper
* Add `AssetSupplies` back to `types/genesis_test`


x/committee/types/permissions.go
```diff
@@ -473,9 +473,10 @@ type AllowedAssetParam struct {
...
 
 func (aap AllowedAssetParam) Allows(current, incoming bep3types.AssetParam) bool {
+
 	allowed := ((aap.Denom == current.Denom) && (aap.Denom == incoming.Denom)) && // require denoms to be all equal
 		((current.CoinID == incoming.CoinID) || aap.CoinID) &&
-		(current.Limit.Equal(incoming.Limit) || aap.Limit) &&
+		(current.SupplyLimit.Equals(incoming.SupplyLimit) || aap.Limit) &&
 		((current.Active == incoming.Active) || aap.Active)
 	return allowed
 }
```
rhuairahrighairidh, line 482: Continuing from https://github.com/Kava-Labs/kava/pull/616#discussion_r468220235
I feel like being able to change MaxSwapAmount and MinBlockLock would be useful if it's not too much work. Gives more options for tuning the deputy.

x/committee/types/permissions.go
```diff
@@ -465,19 +465,25 @@ func (aaps AllowedAssetParams) Allows(current, incoming bep3types.AssetParams) b
...
-	Active bool   `json:"active" yaml:"active"`
+	Denom         string `json:"denom" yaml:"denom"`
+	CoinID        bool   `json:"coin_id" yaml:"coin_id"`
+	Limit         bool   `json:"limit" yaml:"limit"`
+	Active        bool   `json:"active" yaml:"active"`
+	MaxSwapAmount bool   `json:"max_swap_amount" yaml:"max_swap_amount"`
+	MinBlockLock  bool   `json:"min_block_lock" yaml:"min_block_lock"`
 }
 
+// Allows bep3 AssetParam parameters than can be changed by committee
```
denalimarsh, line 478:
```suggestion
// Allows bep3 AssetParam parameters that can be changed by committee
```

denalimarsh (approved):
LGTM

rhuairahrighairidh (approved):
looks good

migrate/v0_11/migrate.go
```diff
@@ -19,28 +22,22 @@ func MigrateBep3(oldGenState v0_9bep3.GenesisState) v0_11bep3.GenesisState {
 			CoinID:        asset.CoinID,
 			DeputyAddress: v0_9Params.BnbDeputyAddress,
 			FixedFee:      v0_9Params.BnbDeputyFixedFee,
-			MinSwapAmount: v0_9Params.MinAmount,
+			MinSwapAmount: sdk.OneInt(), // set min swap to one - prevents accounnts that hold zero bnb from creating spam txs
```
fedekunze, line 25:
```suggestion
			MinSwapAmount: sdk.OneInt(), // set min swap to one - prevents accounts that hold zero bnb from creating spam txs
```

x/bep3/genesis.go
```diff
@@ -20,6 +20,11 @@ func InitGenesis(ctx sdk.Context, keeper Keeper, supplyKeeper types.SupplyKeeper
 		panic(fmt.Sprintf("failed to validate %s genesis state: %s", ModuleName, err))
 	}
 
+	// only set the previous block time if it's different than default
```
fedekunze, line 23: should we set an empty time (`time.Time{}` or `time.Unix(0, 0)`) if it's the same as the default?

fedekunze: I guess I don't understand why do we need the condition here. A comment would help clarify that

x/bep3/keeper/asset.go
```diff
@@ -125,3 +143,39 @@ func (k Keeper) DecrementOutgoingAssetSupply(ctx sdk.Context, coin sdk.Coin) err
...
+	supply := types.NewAssetSupply(
+		sdk.NewCoin(denom, sdk.ZeroInt()), sdk.NewCoin(denom, sdk.ZeroInt()),
+		sdk.NewCoin(denom, sdk.ZeroInt()), sdk.NewCoin(denom, sdk.ZeroInt()), time.Duration(0))
+	k.SetAssetSupply(ctx, supply, denom)
+	return supply
+}
+
+// UpdateTimeBasedSupplyLimits updates the time based supply for each asset, resetting it if the current time window has elapsed.
+func (k Keeper) UpdateTimeBasedSupplyLimits(ctx sdk.Context) {
+	assets, _ := k.GetAssets(ctx)
```
fedekunze, line 158: what if the assets are not found?

x/bep3/keeper/swap.go
```diff
@@ -161,6 +160,16 @@ func (k Keeper) ClaimAtomicSwap(ctx sdk.Context, from sdk.AccAddress, swapID []b
 		if err != nil {
 			return err
 		}
+		// incoming case - coins should be MINTED, then sent to user
```
fedekunze, line 163: nit, I would move the logic of each case into a private function

x/bep3/genesis.go
karzak: 👍  this was unnecessary

PR #624, opened 2020-08-14
Update reward vesting length calculation

karzak:
Fixes issue where rewards are added to the first reward period, rather than creating a new reward period 1 year in the future.

### DIFF
--- a/x/incentive/keeper/payout_test.go
+++ b/x/incentive/keeper/payout_test.go
@@ -35,7 +35,7 @@ func (suite *KeeperTestSuite) setupChain() {
 	)
 	supplyKeeper := tApp.GetSupplyKeeper()
 	macc := supplyKeeper.GetModuleAccount(ctx, kavadist.ModuleName)
-	err := supplyKeeper.MintCoins(ctx, macc.GetName(), cs(c("ukava", 500)))
+	err := supplyKeeper.MintCoins(ctx, macc.GetName(), cs(c("ukava", 600)))
 	suite.Require().NoError(err)
 
 	// sets addrs[0] to be a periodic vesting account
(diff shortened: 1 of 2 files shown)


rhuairahrighairidh (changes requested):
I think there might be a bug when the current block time is in the middle of the schedule, and the length being added is small.

Line 199 `if lengthCounter < length {` seems suspicious. Should `lengthCounter` be compared to something like (block time - vesting start time) + length?

x/incentive/keeper/payout_test.go
```diff
@@ -99,178 +100,260 @@ func (suite *KeeperTestSuite) setupExpiredClaims() {
...
+					vesting.Period{Length: 5, Amount: cs(c("ukava", 5))},
+					vesting.Period{Length: 12, Amount: cs(c("ukava", 6))}},
+				expectedStartTime: 100,
+				expectedEndTime:   132,
+			},
+			errArgs: errArgs{
+				expectErr: true,
+				contains:  "insufficient funds",
+			},
 		},
```
rhuairahrighairidh, line 325: nit: I had a look at the code coverage, and there was a small section uncovered (ln 182 to 188), so this test case hits it.
```suggestion
		},
		{
			name: "add large period mid schedule",
			args: args{
				accArgs: accountArgs{
					periods: vesting.Periods{
						vesting.Period{Length: 5, Amount: cs(c("ukava", 5))},
						vesting.Period{Length: 5, Amount: cs(c("ukava", 5))},
						vesting.Period{Length: 5, Amount: cs(c("ukava", 5))},
						vesting.Period{Length: 5, Amount: cs(c("ukava", 5))}},
					origVestingCoins: cs(c("ukava", 20)),
					startTime:        100,
					endTime:          120,
				},
				period:              vesting.Period{Length: 50, Amount: cs(c("ukava", 6))},
				ctxTime:             time.Unix(110, 0),
				mintModAccountCoins: true,
				expectedPeriods: vesting.Periods{
					vesting.Period{Length: 5, Amount: cs(c("ukava", 5))},
					vesting.Period{Length: 5, Amount: cs(c("ukava", 5))},
					vesting.Period{Length: 5, Amount: cs(c("ukava", 5))},
					vesting.Period{Length: 5, Amount: cs(c("ukava", 5))},
					vesting.Period{Length: 40, Amount: cs(c("ukava", 6))}},
				expectedStartTime: 100,
				expectedEndTime:   160,
			},
			errArgs: errArgs{
				expectErr: false,
				contains:  "",
			},
		},
```

rhuairahrighairidh (approved):
looks good. I tried adding a zero length period (in case we set vesting to zero) and it looks like it works as well.

PR #626, opened 2020-08-17
Add Kava Ledger App usage guide to docs

denalimarsh:
Adds Kava Ledger App usage guide to docs.kava.io under new section 'Platform'.

If there's interest a 'Kava Web App' usage guide can be added to the Platform section in a future PR.

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -20,6 +20,15 @@ module.exports = {
           ]
         },
         {
+          title: "Platform",
+          children: [
+            {
+              title: "Kava Ledger App",
+              path: "/platform/ledger.html"
+            }
+          ]
+        },
+        {
           title: "Kava Tools",
           children: [
             {
--- a/docs/ledger.md
+++ b/docs/ledger.md
@@ -1,0 +1,77 @@
+<!---
+parent:
+  order: false
+--->
+# Kava Ledger App
+
+[Ledger Nano S](https://www.ledger.com/products/ledger-nano-s) is a hardware wallet that enables users to securely access and use cryptocurrencies. Hardware wallets are considered very secure for the storage of a user’s private keys in the blockchain world. Your digital assets are safe even when using an infected (or untrusted) PC.
+
+## Before you begin
+1. [Initialize](https://support.ledgerwallet.com/hc/en-us/articles/360000613793) your Ledger Nano S.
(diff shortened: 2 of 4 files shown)


docs/.vuepress/config.js
```diff
@@ -19,6 +19,15 @@ module.exports = {
             },
           ]
         },
+        {
+          title: "Platform",
```
fedekunze, line 23: should this be Guide or Tutorial instead?

docs/ledger.md
```diff
@@ -0,0 +1,77 @@
...
+3. Confirm your ledger address by clicking Show on device.
+4. View balances by clicking Balances.
+
+### Loading funds onto Kava
+1. From the Balances display, click Load.
+2. Close the Kava ledger app and open the Binance ledger app.
+3. Click Connect Binance App.
+4. On the Fund Account screen enter the amount and recipient.
+5. Click Fund.
+6. Review and sign the transaction on your ledger device.
```
fedekunze, line 38: From the requirements

> - How to receive crypto assets
>    - Document all the steps.
>    - Explicitly instruct to verify receive addresses on the device.
> - How to send crypto assets
>    - Document all the steps and form fields to fill.
>    - Explicitly instruct to verify transaction details on the device display, before confirming and signing it.


I think a separate section needs to be added in order to send + receive coins to an address (eg send KAVA or USDX). Can be via Lunie or other

docs/.vuepress/config.js
denalimarsh: The docs are primarily developer-facing, so a Guide/Tutorial section that doesn't include anything related to development feels off. Future PRs will expand this section with other platform related content.

denalimarsh: Open to opinions on this one

docs/ledger.md
denalimarsh: We can’t use Lunie’s web wallet to send crypto using the Kava ledger app, we'll need to add this functionality directly to the web app as even after we update the middleware Lunie isn't going to replace their own library with our fork.

denalimarsh:
Added blocked label as we need a section for sending/receiving crypto using the ledger app. We can't use existing apps such as Lunie's web wallet for this as they're pegged to a different version of the ledger app, so we need to add the functionality directly to our own web app.

@nddeluca @downing034

PR #629, opened 2020-08-19
Add collateral type to cdp

karzak:
Adds a 'Type' field to `CDP` and 'CollateralParam` structs. This enables one denom to have two different cdp types, for example, 'busd-a' and 'busd-b'. 

For messages, the denom of a CDP  is no longer implied by the coins that are being used as collateral. `CollateralType` is now a required field for cdp messages. For example,

```
kvcli tx cdp create 1000000000bnb 10000000usdx --from key
```
becomes
```
kvcli tx cdp create 1000000000bnb 10000000usdx bnb-a --from key
```

The store didn't undergo any changes, as the prefix for each collateral was used to index cdps by collateral denom, rather than the denom itself. Instead, getter methods that previously assumed each CollateralParam's 'Denom' was unique now assume each CollateralParam's 'Type' is unique. The params validation logic has been updated to reflect this change. 

Feedback wanted:
* I opted not to change the API of the current 'incentive' implementation, but its semantic meaning has changed. Previously, the `Reward.Denom` param referred to the denom of CDPs that rewards would be given for. Now, `Reward.Denom` would refer to the `Type` of CDPs that rewards would be given for. Would anyone argue in favor of changing the `Reward.Denom` field to be `Reward.CollateralType` or something similar?
* Similarly, I've left CDP method names such as 'IterateCDPsByDenom` as is, although again, Denom now refers to collateral type. 

Notes
* I'll include migration logic in a follow-on PR, as these changes should be given thorough review since they are breaking changes to the core CDP logic.


x/cdp/types/cdp.go
```diff
@@ -12,20 +12,22 @@ import (
 
 // CDP is the state of a single collateralized debt position.
 type CDP struct {
-	ID              uint64         `json:"id" yaml:"id"`                 // unique id for cdp
-	Owner           sdk.AccAddress `json:"owner" yaml:"owner"`           // Account that authorizes changes to the CDP
+	ID              uint64         `json:"id" yaml:"id"`       // unique id for cdp
+	Owner           sdk.AccAddress `json:"owner" yaml:"owner"` // Account that authorizes changes to the CDP
+	Type            string         `json:"type" yaml:"type"`
```
fedekunze, line 17: a comment here would be nice

x/incentive/types/genesis.go
```diff
@@ -20,7 +19,10 @@ func (gcp GenesisClaimPeriodID) Validate() error {
 	if gcp.ID == 0 {
 		return errors.New("genesis claim period id cannot be 0")
 	}
-	return sdk.ValidateDenom(gcp.Denom)
```
fedekunze, line 23: can you document this change? this could lead to invalid denoms panicking

x/incentive/types/msg.go
```diff
@@ -33,7 +36,10 @@ func (msg MsgClaimReward) ValidateBasic() error {
 	if msg.Sender.Empty() {
 		return sdkerrors.Wrap(sdkerrors.ErrInvalidAddress, "sender address cannot be empty")
 	}
-	return sdk.ValidateDenom(msg.Denom)
```
fedekunze, line 36: ditto

x/incentive/types/params.go
```diff
@@ -141,7 +142,10 @@ func (r Reward) Validate() error {
 	if r.ClaimDuration <= 0 {
 		return fmt.Errorf("claim duration must be positive, is %s for %s", r.ClaimDuration, r.Denom)
 	}
-	return sdk.ValidateDenom(r.Denom)
```
fedekunze, line 144: ditto

x/incentive/types/rewards.go
```diff
@@ -62,7 +63,10 @@ func (rp RewardPeriod) Validate() error {
 	if rp.ClaimTimeLock == 0 {
 		return errors.New("reward claim time lock cannot be 0")
 	}
-	return sdk.ValidateDenom(rp.Denom)
```
fedekunze, line 65: ditto

rhuairahrighairidh:
Thoughts on Type/Denom naming:
I reckon maker went with separate collateral denoms USDC-A and USDC-B because it was easiest to implement. But they're the same underlying collateral, just used in different types of CDP with different parameters.
Before, each of our CDP types had a unique collateral, so we used the denom to identify them. But now we’re using an arbitrary string to identify CDP types.
I’d vote to make a distinction between cdp type and collateral denoms..
eg
reward.Denom -> reward.CDPType
GetCdpByOwnerAndDenom -> GetCDPByOwnerAndType
CollateralParams -> CDPTypeParams (or CDPParams?)
k.GetCollateral -> k.GetCDPTypeParams
collateralType -> cdpType

PR #633, opened 2020-08-21
[WIP] Add cdp, incentive, committee migrations for v0.9 -> v0.11

karzak:
Migrates exported state from v0.9 (kava-3) to v0.11 (kava-4) for cdp, incentive and committee to account for changes introduced in #629. Includes tests and test data from kava-3. 

Removes code related to kava-2 -> kava-3 migration, as it implicitly assumed that the current state of `app` would not have breaking changes. Since this is no longer true, and we don't need to support kava-2 -> kava-3 migrations in the client for any reason I can think of, I removed that `kava-3` section of `contrib`. 

For reviewers:
Does this structure make sense (legacy v0.9, v0.11 types in the package, migrations at the top level)? 
Any additional tests needed for these modules? 

TODO: 
[ ] Include tests that verify migrations against current master branch, in additional to the legacy `v011`. Ideally, tests against master should have a way of being ignored once v0.11 is released and we've made breaking changes. 
[ ] Include CDP genesis state changes from #644 
[ ] Include incentive state changes from #655
[ ] Include auth.Accounts state changes from #625


rhuairahrighairidh:
Initial thoughts:
- I would vote to use the types from the module types folders rather than the copied types in legacy/v0_11 folders. In cosmos, they merged a few PRs that changed their types slightly, but the changes weren't copied over to the legacy folders, so the migrations siliently broke.
- I like the current tests that migrate a lot of data, then check it passes validation. Might be worth also having tests that migrate a small genesis state and check it's equal to hand migrated json. I remember making a lot of small mistakes on the previous migration that weren't caught by genesis validation.

karzak:
Closing in favor of smaller, individual PRs (662-667+)

Issue #639, opened 2020-08-27
Querier improvements for kava-4


denalimarsh:
Queries are organized by module and priority. I'm working on implementing these on branch `dm-queries`.
- [1] First priority: there is currently no way to get this data
- [2] Second priority: active consumers need this data (and are mitigating with hacky solutions)
- [3] Third priority: "nice to have" for consumers

CDP
- [x] [1] Query CDP by ID
- [x] [1] Query total amount of savings rate distributed USDX
- [x] [1] Query CDPs by owner address (without collateral type)
- [x] [2] Query total amount locked for each collateral type
- [ ] [2] Query total amount borrowed for each principal type (currently just USDX)
- [ ] [2] Query total amount of savings rate distributed USDX by collateral type
- [ ] [3] Query total amount of fees currently outstanding
- [x] [3] Query total amount of fees that have been repaid (i.e. protocol profit)

Auction
- [x] [1] Query for auctions associated with a specific address. In this case the lotReturnAddrs from StartCollateralAuction.
- [x] [3] Use events to query closed auctions. -> **develop a separate event storage/query service for our web app?**

BEP3
- [ ] [2] migrate calculate-swap-id cmd from kvtool - it requires less args
- [ ] [2] marshal YAML gives bytes formatted swap ID
- [ ] [2] add marshalYAML (or String?) methods to swap direction and status types so they show up as string not numbers in query output
- [ ] [3] Consider adding ExpectedOtherChainSwapID to the augmented AtomicSwap.
- [ ] [3] Improve "atomic swap ID not found error" for claim txs. Example: https://lcd-kava-3.cosmostation.io/txs/DEF1D9CBC7A7E9A71B09EE0BD6DAA7B5DB4C3DF6D0CA06D3758C5BCEA9E32679

Incentive
- [x] [1] Query ongoing claims
- [ ] [2] Get reward APY for an asset at the current period reward rate. Under the current architecture, this should be a CDP module query because we need the amount of collateral locked and the collateral’s system price, requiring the import of CDP and Pricefeed keeper.

denalimarsh:
@karzak @rhuairahrighairidh @nddeluca 

Feel free to modify the priority or add new queries to the list.

karzak:
re
> Query total amount locked for each collateral type

[...]

PR #644, opened 2020-09-01
Querier improvements: CDP and Auction priority 1 queries

denalimarsh:
Implements queries from https://github.com/Kava-Labs/kava/issues/639. Make sure to add `"savings_rate_distributed": "0"` to the CDP module's genesis state in your genesis.json file.

- [x] Query total amount of savings rate distributed
- `/cdp/savingsRateDist`
- [x] Query CDP by ID
- `cdp/cdps?id=1`
- [x] Query CDPs by owner address
- `cdp/cdps?owner=kava1l0xsq2z7gqd7yly0g40y5836g0appumark77ny`
- [x] Query CDPs by collateral type
- `cdp/cdps?collateral-type=bnb-a`
- [x] Query CDP by ratio
- `cdp/cdps?ratio=4.5`
- [x] Query for CollateralAuctions by the source CDP owner's address
- `auction/auctions?type=collateral&owner=kava1l0xsq2z7gqd7yly0g40y5836g0appumark77ny`

The new `cdps` querier uses params instead of appending to the query path.

### DIFF
--- a/x/auction/keeper/querier_test.go
+++ b/x/auction/keeper/querier_test.go
@@ -41,7 +41,7 @@ func (suite *QuerierTestSuite) SetupTest() {
 	tApp := app.NewTestApp()
 	ctx := tApp.NewContext(true, abci.Header{Height: 1, Time: tmtime.Now()})
 
-	_, addrs := app.GeneratePrivKeyAddressPairs(1)
+	_, addrs := app.GeneratePrivKeyAddressPairs(10)
 	buyer := addrs[0]
 	modName := cdp.LiquidatorMacc
 
@@ -64,8 +64,16 @@ func (suite *QuerierTestSuite) SetupTest() {
 	// Populate with auctions
 	randSrc := rand.New(rand.NewSource(int64(1234)))
 	for j := 0; j < TestAuctionCount; j++ {
-		lotAmount := simulation.RandIntBetween(randSrc, 10, 100)
-		id, err := suite.keeper.StartSurplusAuction(suite.ctx, modName, c("token1", int64(lotAmount)), "token2")
+		var id uint64
+		var err error
+		lotAmount := int64(simulation.RandIntBetween(randSrc, 10, 100))
+		ownerAddrIndex := simulation.RandIntBetween(randSrc, 1, 9)
+		if ownerAddrIndex%2 == 0 {
+			id, err = suite.keeper.StartSurplusAuction(suite.ctx, modName, c("token1", lotAmount), "token2")
+		} else {
+			id, err = suite.keeper.StartCollateralAuction(suite.ctx, modName, c("token1", lotAmount), c("usdx", int64(20)),
+				[]sdk.AccAddress{addrs[ownerAddrIndex]}, []sdk.Int{sdk.NewInt(lotAmount)}, c("debt", int64(10)))
+		}
 		suite.NoError(err)
 
 		auc, found := suite.keeper.GetAuction(suite.ctx, id)
--- a/x/auction/keeper/querier_test.go
+++ b/x/auction/keeper/querier_test.go
@@ -108,7 +108,7 @@ func (suite *QuerierTestSuite) TestQueryAuctions() {
 	query := abci.RequestQuery{
 		Path: strings.Join([]string{custom, types.QuerierRoute, types.QueryGetAuctions}, "/"),
 		Data: types.ModuleCdc.MustMarshalJSON(
-			types.NewQueryAllAuctionParams(int(1), int(TestAuctionCount), "", "", ""),
+			types.NewQueryAllAuctionParams(int(1), int(TestAuctionCount), "", "", "", nil),
 		),
 	}
 
--- a/x/cdp/client/cli/query.go
+++ b/x/cdp/client/cli/query.go
@@ -41,6 +41,7 @@ func GetQueryCmd(queryRoute string, cdc *codec.Codec) *cobra.Command {
 		QueryCdpDepositsCmd(queryRoute, cdc),
 		QueryParamsCmd(queryRoute, cdc),
 		QueryGetAccounts(queryRoute, cdc),
+		QueryGetSavingsRateDistributed(queryRoute, cdc),
 	)...)
 
 	return cdpQueryCmd
@@ -305,7 +306,7 @@ func QueryParamsCmd(queryRoute string, cdc *codec.Codec) *cobra.Command {
 	return &cobra.Command{
 		Use:   "params",
 		Short: "get the cdp module parameters",
-		Long:  "Get the current global cdp module parameters.",
+		Long:  "get the current global cdp module parameters.",
 		Args:  cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cliCtx := context.NewCLIContext().WithCodec(cdc)
@@ -325,11 +326,12 @@ func QueryParamsCmd(queryRoute string, cdc *codec.Codec) *cobra.Command {
 	}
 }
 
+// QueryGetAccounts queries CDP module accounts
 func QueryGetAccounts(queryRoute string, cdc *codec.Codec) *cobra.Command {
 	return &cobra.Command{
 		Use:   "accounts",
-		Short: "Get module accounts",
-		Long:  "Get cdp module account addresses",
+		Short: "get module accounts",
+		Long:  "get cdp module account addresses",
 		Args:  cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cliCtx := context.NewCLIContext().WithCodec(cdc)
@@ -345,6 +347,33 @@ func QueryGetAccounts(queryRoute string, cdc *codec.Codec) *cobra.Command {
 			var out []supply.ModuleAccount
 			if err := cdc.UnmarshalJSON(res, &out); err != nil {
 				return fmt.Errorf("failed to unmarshal accounts: %w", err)
+			}
+			return cliCtx.PrintOutput(out)
+		},
+	}
+}
+
+// QueryGetSavingsRateDistributed queries the total amount of savings rate distributed in USDX
+func QueryGetSavingsRateDistributed(queryRoute string, cdc *codec.Codec) *cobra.Command {
+	return &cobra.Command{
+		Use:   "savingsratedist",
+		Short: "get total amount of savings rate distributed in USDX",
+		Long:  "get total amount of savings rate distributed",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			cliCtx := context.NewCLIContext().WithCodec(cdc)
+
+			// Query
+			res, height, err := cliCtx.QueryWithData(fmt.Sprintf("custom/%s/%s", queryRoute, types.QueryGetSavingsRateDistributed), nil)
+			if err != nil {
+				return err
+			}
+			cliCtx = cliCtx.WithHeight(height)
+
+			// Decode and print results
+			var out sdk.Int
+			if err := cdc.UnmarshalJSON(res, &out); err != nil {
+				return fmt.Errorf("failed to unmarshal sdk.Int: %w", err)
 			}
 			return cliCtx.PrintOutput(out)
 		},
--- a/x/cdp/client/cli/query.go
+++ b/x/cdp/client/cli/query.go
@@ -23,6 +23,7 @@ const (
 	flagCollateralType = "collateral-type"
 	flagOwner          = "owner"
 	flagID             = "id"
+	flagRatio          = "ratio" // returns CDPs under the given collateralization ratio threshold
 )
 
 // GetQueryCmd returns the cli query commands for this module
@@ -35,9 +36,7 @@ func GetQueryCmd(queryRoute string, cdc *codec.Codec) *cobra.Command {
 
 	cdpQueryCmd.AddCommand(flags.GetCommands(
 		QueryCdpCmd(queryRoute, cdc),
-		QueryGetV2CdpsCmd(queryRoute, cdc),
-		QueryCdpsByCollateralTypeCmd(queryRoute, cdc),
-		QueryCdpsByCollateralTypeAndRatioCmd(queryRoute, cdc),
+		QueryGetCdpsCmd(queryRoute, cdc),
 		QueryCdpDepositsCmd(queryRoute, cdc),
 		QueryParamsCmd(queryRoute, cdc),
 		QueryGetAccounts(queryRoute, cdc),
@@ -90,23 +89,25 @@ $ %s query %s cdp kava15qdefkmwswysgg4qxgqpqr35k3m49pkx2jdfnw atom-a
 	}
 }
 
-// QueryGetV2CdpsCmd queries the cdps in the store
-func QueryGetV2CdpsCmd(queryRoute string, cdc *codec.Codec) *cobra.Command {
+// QueryGetCdpsCmd queries the cdps in the store
+func QueryGetCdpsCmd(queryRoute string, cdc *codec.Codec) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "v2cdps",
+		Use:   "cdps",
 		Short: "query cdps with optional filters",
 		Long: strings.TrimSpace(`Query for all paginated cdps that match optional filters:
 Example:
-$ kvcli q cdp v2cdps --collateral-type=bnb
-$ kvcli q cdp v2cdps --owner=kava1hatdq32u5x4wnxrtv5wzjzmq49sxgjgsj0mffm
-$ kvcli q cdp v2cdps --id=21
-$ kvcli q cdp v2cdps --page=2 --limit=100
+$ kvcli q cdp cdps --collateral-type=bnb
+$ kvcli q cdp cdps --owner=kava1hatdq32u5x4wnxrtv5wzjzmq49sxgjgsj0mffm
+$ kvcli q cdp cdps --id=21
+$ kvcli q cdp cdps --ratio=2.75
+$ kvcli q cdp cdps --page=2 --limit=100
 `,
 		),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			strCollateralType := viper.GetString(flagCollateralType)
 			strOwner := viper.GetString(flagOwner)
 			strID := viper.GetString(flagID)
+			strRatio := viper.GetString(flagRatio)
 			page := viper.GetInt(flags.FlagPage)
 			limit := viper.GetInt(flags.FlagLimit)
 
@@ -114,16 +115,13 @@ $ kvcli q cdp v2cdps --page=2 --limit=100
 				cdpCollateralType string
 				cdpOwner          sdk.AccAddress
 				cdpID             uint64
+				cdpRatio          sdk.Dec
 			)
 
-			params := types.NewQueryV2CdpsParams(page, limit, cdpCollateralType, cdpOwner, cdpID)
+			params := types.NewQueryCdpsParams(page, limit, cdpCollateralType, cdpOwner, cdpID, cdpRatio)
 
 			if len(strCollateralType) != 0 {
 				cdpCollateralType = strings.ToLower(strings.TrimSpace(strCollateralType))
-				err := sdk.ValidateDenom(cdpCollateralType)
-				if err != nil {
-					return err
-				}
 				params.CollateralType = cdpCollateralType
 			}
 
@@ -136,11 +134,20 @@ $ kvcli q cdp v2cdps --page=2 --limit=100
 			}
 
 			if len(strID) != 0 {
-				cdpID, err := strconv.Atoi(strID)
+				cdpID, err := strconv.ParseUint(strID, 10, 64)
 				if err != nil {
 					return fmt.Errorf("cannot parse cdp ID %s", strID)
 				}
-				params.ID = uint64(cdpID)
+				params.ID = cdpID
+			}
+
+			params.Ratio = sdk.ZeroDec()
+			if len(strRatio) != 0 {
+				cdpRatio, err := sdk.NewDecFromStr(strRatio)
+				if err != nil {
+					return fmt.Errorf("cannot parse cdp ratio %s", strRatio)
+				}
+				params.Ratio = cdpRatio
 			}
 
 			bz, err := cdc.MarshalJSON(params)
@@ -151,7 +158,7 @@ $ kvcli q cdp v2cdps --page=2 --limit=100
 			cliCtx := context.NewCLIContext().WithCodec(cdc)
 
 			// Query
-			res, height, err := cliCtx.QueryWithData(fmt.Sprintf("custom/%s/%s", queryRoute, types.QueryGetV2Cdps), bz)
+			res, height, err := cliCtx.QueryWithData(fmt.Sprintf("custom/%s/%s", queryRoute, types.QueryGetCdps), bz)
 			if err != nil {
 				return err
 			}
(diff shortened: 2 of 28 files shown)


x/cdp/keeper/querier.go
```diff
@@ -178,3 +183,85 @@ func queryGetAccounts(ctx sdk.Context, req abci.RequestQuery, keeper Keeper) ([]
...
+// query cdps in store and filter by request params
+func queryGetV2Cdps(ctx sdk.Context, req abci.RequestQuery, keeper Keeper) ([]byte, error) {
+	var params types.QueryV2CdpsParams
+	err := types.ModuleCdc.UnmarshalJSON(req.Data, &params)
+	if err != nil {
+		return nil, sdkerrors.Wrap(sdkerrors.ErrJSONUnmarshal, err.Error())
+	}
+
+	// Filter CDPs
+	unfilteredCDPs := keeper.GetAllCdps(ctx)
```
karzak, line 209: Putting all CDPs in memory and filtering is too inefficient as a default querying strategy.

* Check the query params supplied and apply the best keeper function. For example, if collateral type and ID is supplied, call `GetCDP`, if collateral type only is supplied, call `GetCDPsByCollateralType`. 
* I think we already have most of the methods we want, except for `GetCDPsByOwner` (ie, Owner is supplied in the query params, but not collateral type). Since I don't really want to re-write the store at this stage, we can add that method via calling `GetCdpIdsByOwner` then brute forcing it by calling `GetCDP` and trying each denom prefix.

x/cdp/types/querier.go
```diff
@@ -29,6 +31,26 @@ func NewQueryCdpsParams(collateralType string) QueryCdpsParams {
 	}
 }
 
+// QueryV2CdpsParams is the params for a filtered CDP query
+type QueryV2CdpsParams struct {
+	Page            int            `json:"page" yaml:"page"`
+	Limit           int            `json:"limit" yaml:"limit"`
+	CollateralDenom string         `json:"collateral_denom" yaml:"collateral_denom"`
```
karzak, line 38:
```suggestion
	CollateralType string         `json:"collateral_type" yaml:"collateral_type"`
```

x/cdp/types/querier.go
```diff
@@ -29,6 +31,26 @@ func NewQueryCdpsParams(collateralType string) QueryCdpsParams {
...
+	Owner           sdk.AccAddress `json:"owner" yaml:"owner"`
+	ID              uint64         `json:"id" yaml:"id"`
+}
+
+// NewQueryV2CdpsParams creates a new QueryV2CdpsParams
+func NewQueryV2CdpsParams(page, limit int, collateralDenom string, owner sdk.AccAddress, id uint64) QueryV2CdpsParams {
```
karzak, line 44: ditto `collateralType`

x/cdp/types/params.go
```diff
@@ -43,6 +44,7 @@ var (
 	DefaultDebtLot                      = sdk.NewInt(10000000000)
 	DefaultPreviousDistributionTime     = tmtime.Canonical(time.Unix(0, 0))
 	DefaultSavingsDistributionFrequency = time.Hour * 12
+	DefaultSavingsRateDistributed       = sdk.NewInt(0)
```
fedekunze, line 47: this is not defined on the params below

x/cdp/types/genesis.go
```diff
@@ -17,10 +17,12 @@ type GenesisState struct {
 	DebtDenom                string    `json:"debt_denom" yaml:"debt_denom"`
 	GovDenom                 string    `json:"gov_denom" yaml:"gov_denom"`
 	PreviousDistributionTime time.Time `json:"previous_distribution_time" yaml:"previous_distribution_time"`
+	SavingsRateDistributed   sdk.Int   `json:"savings_rate_distributed" yaml:"savings_rate_distributed"`
```
fedekunze, line 20: is this a param? should we remove the var definitions from that file? makes things more confusing

x/cdp/types/params.go
fedekunze: see comments

x/cdp/keeper/savings_test.go
```diff
@@ -76,7 +91,21 @@ func (suite *SavingsTestSuite) TestGetSetPreviousDistributionTime() {
 	pdt, f := suite.keeper.GetPreviousSavingsDistribution(suite.ctx)
 	suite.True(f)
 	suite.Equal(now, pdt)
+}
+
+func (suite *SavingsTestSuite) TestGetSetSavingsRateDistributed() {
+	// Savings rate dist set to 0 when the default genesis is used
+	preSavingsRateDistAmount := suite.keeper.GetSavingsRateDistributed(suite.ctx)
+	suite.True(preSavingsRateDistAmount.Equal(sdk.ZeroInt()))
```
fedekunze, line 99: nit use the default variable instead

x/cdp/keeper/querier.go
```diff
@@ -17,6 +18,8 @@ func NewQuerier(keeper Keeper) sdk.Querier {
 		switch path[0] {
 		case types.QueryGetCdp:
 			return queryGetCdp(ctx, req, keeper)
+		case types.QueryGetV2Cdps:
+			return queryGetV2Cdps(ctx, req, keeper)
 		case types.QueryGetCdps:
```
fedekunze, line 23: are you planning on deprecating this query?

x/cdp/keeper/keeper.go
```diff
@@ -110,3 +110,23 @@ func (k Keeper) IterateCdpsByCollateralRatio(ctx sdk.Context, collateralType str
...
+	store.Set(types.SavingsRateDistributedKey, bz)
+}
+
+// GetSavingsRateDistributed gets the SavingsRateDistributed from the store
+func (k Keeper) GetSavingsRateDistributed(ctx sdk.Context) sdk.Int {
+	var savingsRateDistributed sdk.Int
+	store := prefix.NewStore(ctx.KVStore(k.key), types.SavingsRateDistributedKey)
+	bz := store.Get(types.SavingsRateDistributedKey)
+	if bz == nil {
+		return savingsRateDistributed
```
fedekunze, line 127: this will return an uninitialized `Int`, better use `sdk.ZeroInt` if you are not returning a boolean or an error

x/cdp/keeper/keeper.go
```diff
@@ -110,3 +110,23 @@ func (k Keeper) IterateCdpsByCollateralRatio(ctx sdk.Context, collateralType str
 
 	}
 }
+
+// SetSavingsRateDistributed sets the SavingsRateDistributed in the store
+func (k Keeper) SetSavingsRateDistributed(ctx sdk.Context, totalDistributed sdk.Int) {
+	store := prefix.NewStore(ctx.KVStore(k.key), types.SavingsRateDistributedKey)
+	bz := k.cdc.MustMarshalBinaryLengthPrefixed(totalDistributed)
+	store.Set(types.SavingsRateDistributedKey, bz)
```
fedekunze, line 118: this is actually storing the key under `types.SavingsRateDistributedKey + types.SavingsRateDistributedKey` (i.e duplicated). Use `[]byte{}` instead
```suggestion
	store.Set([]byte{}, bz)
```

x/cdp/keeper/keeper.go
```diff
@@ -110,3 +110,23 @@ func (k Keeper) IterateCdpsByCollateralRatio(ctx sdk.Context, collateralType str
...
+}
+
+// GetSavingsRateDistributed gets the SavingsRateDistributed from the store
+func (k Keeper) GetSavingsRateDistributed(ctx sdk.Context) sdk.Int {
+	var savingsRateDistributed sdk.Int
+	store := prefix.NewStore(ctx.KVStore(k.key), types.SavingsRateDistributedKey)
+	bz := store.Get(types.SavingsRateDistributedKey)
```
fedekunze, line 125:
```suggestion
	bz := store.Get([]byte{})
```

x/cdp/client/rest/query.go
```diff
@@ -16,10 +18,12 @@ import (
 func registerQueryRoutes(cliCtx context.CLIContext, r *mux.Router) {
 	r.HandleFunc("/cdp/accounts", getAccountsHandlerFn(cliCtx)).Methods("GET")
 	r.HandleFunc("/cdp/parameters", getParamsHandlerFn(cliCtx)).Methods("GET")
+	r.HandleFunc("/cdp/savingsratedist", getSavingsRateDistributedHandler(cliCtx)).Methods("GET")
```
fedekunze, line 21: the naming should be consistent with the rest of the endpoints (camelCase)

x/cdp/client/rest/query.go
```diff
@@ -16,10 +18,12 @@ import (
...
 	r.HandleFunc(fmt.Sprintf("/cdp/cdps/cdp/{%s}/{%s}", types.RestOwner, types.RestCollateralType), queryCdpHandlerFn(cliCtx)).Methods("GET")
 	r.HandleFunc(fmt.Sprintf("/cdp/cdps/collateralType/{%s}", types.RestCollateralType), queryCdpsHandlerFn(cliCtx)).Methods("GET")
 	r.HandleFunc(fmt.Sprintf("/cdp/cdps/ratio/{%s}/{%s}", types.RestCollateralType, types.RestRatio), queryCdpsByRatioHandlerFn(cliCtx)).Methods("GET")
 	r.HandleFunc(fmt.Sprintf("/cdp/cdps/cdp/deposits/{%s}/{%s}", types.RestOwner, types.RestCollateralType), queryCdpDepositsHandlerFn(cliCtx)).Methods("GET")
```
fedekunze, line 25: not sure if these are actually necessary or if you can filter the results from the v2Cdps query

x/cdp/client/cli/query.go
```diff
@@ -253,3 +352,30 @@ func QueryGetAccounts(queryRoute string, cdc *codec.Codec) *cobra.Command {
 		},
 	}
 }
+
+// QueryGetSavingsRateDistributed queries the total amount of savings rate distributed in USDX
+func QueryGetSavingsRateDistributed(queryRoute string, cdc *codec.Codec) *cobra.Command {
+	return &cobra.Command{
+		Use:   "savingsratedist",
```
fedekunze, line 359: use separator `-` or `_`?

x/cdp/types/genesis.go
denalimarsh: It's not a param. I agree that having the definitions here is confusing and should probably be updated. But as this PR is scoped to the querier I just went with the existing module layout on this one.

x/cdp/types/params.go
denalimarsh: Responded below

x/cdp/keeper/querier.go
denalimarsh: Ongoing discussion at https://github.com/Kava-Labs/kava/pull/644#discussion_r481130316. The original idea was to maintain existing endpoints as to limit breaking changes for kava-4, but that's more relevant on the REST API as it's the primary point of data consumption.

x/cdp/client/rest/query.go
denalimarsh: These are legacy endpoints. The goal of v2Cdps is to replace these queries - but we'll likely want to maintain the legacy endpoints for kava-4 to limit breaking changes.

x/cdp/keeper/querier.go
denalimarsh: Looks like `keeper_bench_test.go` can likely be used to get some tangible efficiency metrics. This in-memory filtering strategy is something I picked up from the [gov module](https://github.com/cosmos/cosmos-sdk/blob/master/x/gov/client/rest/query.go), which probably has a lighter store than our CDP module.

We're also using filtering in the `auction/auctions` and `bep3/swaps` queries - if this method proves too inefficient we should consider patching bep3 as well. Auction is probably fine, but there can easily be as many swaps in memory as cdps given that we store swaps in longterm storage before deletion.

karzak: +1 to investigating bep3 filter method - wouldn't be surprised if we were storing thousands of swaps eventually - wouldn't want to load them all either. Agree that auctions is probably not a concern at the moment.

x/cdp/client/cli/query.go
denalimarsh: https://github.com/Kava-Labs/kava/pull/644/commits/3a59e79d6cf41cda60c1f5a6819f028d30b6f47a

x/cdp/client/rest/query.go
denalimarsh: https://github.com/Kava-Labs/kava/pull/644/commits/3a59e79d6cf41cda60c1f5a6819f028d30b6f47a

x/cdp/keeper/keeper.go
denalimarsh: https://github.com/Kava-Labs/kava/pull/644/commits/2ffcdfa8feb3fb5916bc0eb63d91669d1f65f6ba

x/cdp/keeper/keeper.go
denalimarsh: https://github.com/Kava-Labs/kava/pull/644/commits/2ffcdfa8feb3fb5916bc0eb63d91669d1f65f6ba

x/cdp/keeper/keeper.go
denalimarsh: https://github.com/Kava-Labs/kava/pull/644/commits/2ffcdfa8feb3fb5916bc0eb63d91669d1f65f6ba

x/cdp/keeper/keeper.go
```diff
@@ -110,3 +110,23 @@ func (k Keeper) IterateCdpsByCollateralRatio(ctx sdk.Context, collateralType str
...
+// SetSavingsRateDistributed sets the SavingsRateDistributed in the store
+func (k Keeper) SetSavingsRateDistributed(ctx sdk.Context, totalDistributed sdk.Int) {
+	store := prefix.NewStore(ctx.KVStore(k.key), types.SavingsRateDistributedKey)
+	bz := k.cdc.MustMarshalBinaryLengthPrefixed(totalDistributed)
+	store.Set(types.SavingsRateDistributedKey, bz)
+}
+
+// GetSavingsRateDistributed gets the SavingsRateDistributed from the store
+func (k Keeper) GetSavingsRateDistributed(ctx sdk.Context) sdk.Int {
+	var savingsRateDistributed sdk.Int
```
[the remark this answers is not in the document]
denalimarsh: https://github.com/Kava-Labs/kava/pull/644/commits/2ffcdfa8feb3fb5916bc0eb63d91669d1f65f6ba

x/cdp/keeper/savings_test.go
denalimarsh: https://github.com/Kava-Labs/kava/pull/644/commits/c682f2d721de29f4d36fb5c378f1a0fa99393970

x/auction/keeper/querier_test.go
```diff
@@ -108,7 +108,7 @@ func (suite *QuerierTestSuite) TestQueryAuctions() {
 	query := abci.RequestQuery{
 		Path: strings.Join([]string{custom, types.QuerierRoute, types.QueryGetAuctions}, "/"),
 		Data: types.ModuleCdc.MustMarshalJSON(
-			types.NewQueryAllAuctionParams(int(1), int(TestAuctionCount), "", "", ""),
+			types.NewQueryAllAuctionParams(int(1), int(TestAuctionCount), "", "", "", nil),
```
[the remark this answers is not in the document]
denalimarsh: https://github.com/Kava-Labs/kava/pull/644/commits/c682f2d721de29f4d36fb5c378f1a0fa99393970

x/cdp/keeper/savings.go
```diff
@@ -32,6 +32,11 @@ func (k Keeper) DistributeSavingsRate(ctx sdk.Context, debtDenom string) error {
 	totalSurplus := sdk.NewDecFromInt(surplusToDistribute)
 	totalSupply := sdk.NewDecFromInt(totalSupplyLessModAccounts.AmountOf(debtDenom))
 
+	// Update total savings rate distributed by surplus to distribute
+	currDistributed := k.GetSavingsRateDistributed(ctx)
+	newTotalDistributed := currDistributed.Add(surplusToDistribute)
+	k.SetSavingsRateDistributed(ctx, newTotalDistributed)
```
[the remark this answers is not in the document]
denalimarsh: Good catch, updated in https://github.com/Kava-Labs/kava/pull/644/commits/211f25559397c555a315c44c04e485e48a5f1ebd

x/cdp/keeper/savings.go
```diff
@@ -32,6 +32,11 @@ func (k Keeper) DistributeSavingsRate(ctx sdk.Context, debtDenom string) error {
 	totalSurplus := sdk.NewDecFromInt(surplusToDistribute)
 	totalSupply := sdk.NewDecFromInt(totalSupplyLessModAccounts.AmountOf(debtDenom))
 
+	// Update total savings rate distributed by surplus to distribute
+	currDistributed := k.GetSavingsRateDistributed(ctx)
```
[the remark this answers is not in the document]
denalimarsh: https://github.com/Kava-Labs/kava/pull/644/commits/211f25559397c555a315c44c04e485e48a5f1ebd

x/cdp/types/querier.go
denalimarsh: https://github.com/Kava-Labs/kava/pull/644/commits/343a24261740510649211726d1038f84107840ef

x/cdp/types/querier.go
denalimarsh: https://github.com/Kava-Labs/kava/pull/644/commits/343a24261740510649211726d1038f84107840ef

x/cdp/keeper/querier.go
denalimarsh: Removed in https://github.com/Kava-Labs/kava/pull/644/commits/6f8355c26f39aa6e0908214bd67621e02cd0f760

x/cdp/client/cli/query.go
```diff
@@ -79,6 +90,93 @@ $ %s query %s cdp kava15qdefkmwswysgg4qxgqpqr35k3m49pkx2jdfnw atom-a
 	}
 }
 
+// QueryGetV2CdpsCmd queries the cdps in the store
+func QueryGetV2CdpsCmd(queryRoute string, cdc *codec.Codec) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "v2cdps",
```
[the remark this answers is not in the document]
denalimarsh: Removed `v2cdps` and added ratio as a query param in https://github.com/Kava-Labs/kava/pull/644/commits/6f8355c26f39aa6e0908214bd67621e02cd0f760. Also fixed an issue where `Type` wasn't being passed from CDP -> AugmentedCDP and AugmentedCDP was printing the collateral denom instead.

I'm going to spend some more time digging into CDP.Type. It's important throughout the CDP module and feels like a potential weak point at the moment given the huge surface area. I think we should be wary of conflating CDP.Type with CDP.Collateral.Denom in tests e.g. https://github.com/Kava-Labs/kava/pull/644/commits/7fbe1919f3833df990a48a7e5c0486403a5ebbbe.

x/cdp/keeper/querier.go
denalimarsh: Halved the memory usage by converting to AugmentedCDP in the filter loop (as a byproduct of ratio param implementation requirements), but that's still only a constant O(n) improvement. So still need to address this one.

fedekunze (approved):

x/cdp/keeper/querier.go
denalimarsh: @karzak added the proposed CDP query result filtering in https://github.com/Kava-Labs/kava/pull/644/commits/1c6d5d52ed5f6505b9bc81c8e05943090a6ca8ac. The BEP3 module doesn't have any of the analogous keeper methods, so updating the querier would require significant store modifications. The query filtering is still done in linear time.

x/cdp/keeper/querier_test.go
```diff
@@ -303,6 +303,17 @@ func (suite *QuerierTestSuite) TestQueryAccounts() {
...
+	bz, err := suite.querier(ctx, []string{types.QueryGetSavingsRateDistributed}, abci.RequestQuery{})
+	suite.Nil(err)
+	suite.NotNil(bz)
+
+	var distAmount sdk.Int
+	suite.Nil(types.ModuleCdc.UnmarshalJSON(bz, &distAmount))
+	suite.True(sdk.ZeroInt().Equal(distAmount))
+}
+
 func TestQuerierTestSuite(t *testing.T) {
```
karzak, line 317: Given complexity of the matching/filtering cdp function, we should have a test for `QueryGetCdps`. Either a new test in the querier suite or even a unit test of the `filterCDPs` and `findItersection` functions

karzak (changes requested):
Looks directionally correct, I think a few further optimizations for the new CDPs query would be good and a basic test.

x/cdp/keeper/querier.go
```diff
@@ -178,3 +183,150 @@ func queryGetAccounts(ctx sdk.Context, req abci.RequestQuery, keeper Keeper) ([]
...
+
+	return bz, nil
+}
+
+// filterCDPs queries the store for all CDPs that match query params
+func filterCDPs(ctx sdk.Context, k Keeper, params types.QueryCdpsParams) types.AugmentedCDPs {
+	var matchCollateralType, matchOwner, matchID, matchRatio types.CDPs
+
+	// match cdp collateral denom (if supplied)
+	if len(params.CollateralType) > 0 {
```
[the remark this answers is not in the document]
denalimarsh: The query `cdp/cdps?owner="kava123abc..."&collateral-type="bnb-a"` should return exactly 1 CDP. The approach in https://github.com/Kava-Labs/kava/pull/644/commits/8493285a182aa8e29102cac883aae40f66d0352e addresses this revision while maintaining consistency of query results.

x/cdp/keeper/querier_test.go
denalimarsh: Added tests for `FindIntersection`, `FilterCDPs`, and `QueryGetCDPs` in https://github.com/Kava-Labs/kava/pull/644/commits/2f1cde9217553fff9a071300d1b0092291a36024.

Had to capitalize some functions so I could import them into the test suite. Wonder what the 'proper' approach to module/module test organization would be here, exposing internal functions just for the test suite feels off.

karzak: Yeah I've run into this as well. There didn't seem to be a consensus best answer, so I've generally resorted to making functions public for purposes of testing, which feels bad.

karzak (approved):

x/cdp/keeper/querier_test.go
rhuairahrighairidh: Anything that requires a keeper/app to be setup needs to be exported. If I remember correctly the test files need be in a `*_test` package to avoid import cycles.
But anything that doesn't require a keeper/app can live in a normal test file. Tests for `findIntersection` and `filterCDP`could live in a `filter_test.go` or something with `package keeper`. Though not important

x/cdp/keeper/querier.go
```diff
@@ -178,3 +183,159 @@ func queryGetAccounts(ctx sdk.Context, req abci.RequestQuery, keeper Keeper) ([]
...
+			commonCDPs = matchID
+		}
+	}
+	if len(matchRatio) > 0 {
+		if len(commonCDPs) > 0 {
+			commonCDPs = FindIntersection(commonCDPs, matchRatio)
+		} else {
+			commonCDPs = matchRatio
+		}
+	}
```
rhuairahrighairidh, line 302: This logic produces slightly confusing results like:

`kvcli q cdp cdps --collateral-type nonsense --ratio 2.0 ` returns all 2.0 ratio cdps, rather than no cdps.
`kvcli q cdp cdps --owner <address> --collateral-type btc` returns all of a user's cdps when they don't have a btc cdp, rather than no cdps

I think whether the `matchX` results are included in the final intersection should be decided by whether the query param was 0 (or "" etc), rather than whether the length of results was 0.

rhuairahrighairidh (changes requested):
Overall looks good, these queries will be pretty helpful. Left a couple of requests

x/cdp/types/genesis.go
```diff
@@ -65,6 +69,10 @@ func (gs GenesisState) Validate() error {
 		return fmt.Errorf("previous distribution time not set")
 	}
 
+	if gs.SavingsRateDistributed.IsNegative() {
+		return fmt.Errorf("savings rate distributed invalid")
+	}
```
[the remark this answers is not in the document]
denalimarsh: Added the validation strategy we're using for other sdk.Ints in https://github.com/Kava-Labs/kava/pull/644/commits/40253c06c6d4c005adda69312269e2e5692f7dc4.

x/cdp/keeper/querier.go
```diff
@@ -178,3 +183,159 @@ func queryGetAccounts(ctx sdk.Context, req abci.RequestQuery, keeper Keeper) ([]
...
+		for _, denom := range denoms {
+			cdp, found := k.GetCDP(ctx, denom, params.ID)
+			if found {
+				matchID = append(matchID, cdp)
+			}
+		}
+	}
+
+	// match cdp ratio (if supplied)
+	if params.Ratio.GT(sdk.ZeroDec()) {
```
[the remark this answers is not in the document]
denalimarsh: Defaulted the param to sdk.ZeroDec() when it's initially parsed in both the CLI and REST in https://github.com/Kava-Labs/kava/pull/644/commits/0bd8491f9fd9d4566078fdc87cc1ca62b63d2249.

x/cdp/keeper/querier.go
denalimarsh: Agreed, updated in https://github.com/Kava-Labs/kava/pull/644/commits/5a1f7e84a928fffd729cd2fa9716d05fe1c319ae

x/cdp/types/genesis.go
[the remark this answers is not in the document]
rhuairahrighairidh: This doesn't actually fix the panic. But looking into this issue a  bit more, it occurs in a lot of places in our code but doesn't cause immediate problems. So probably worth punting to a future PR.
The sdk doesn't make it easy to deal with this issue. For example `sdk.Dec` has a `IsNil` method, but `sdk.Int` doesn't.

x/cdp/keeper/querier.go
rhuairahrighairidh: Looks like the second example above still produces the same results

rhuairahrighairidh (changes requested):
looks like there's a remaining issue with filtering, but otherwise good

x/cdp/types/genesis.go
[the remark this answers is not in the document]
denalimarsh: We can open a PR for adding `IsNil` to `sdk.Int` upstream in the Cosmos SDK, but yeah probably won't be available in our codebase for many months.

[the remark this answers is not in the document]
denalimarsh: Governance can prevent malicious param change proposals, but it seems like a ubiquitous attack vector given how common `sdk.Int` is

x/cdp/keeper/querier.go
denalimarsh: Addresed in https://github.com/Kava-Labs/kava/pull/644/commits/d01f9b8e81dfa3be585543e23428d77ba1485e66

denalimarsh:
@karzak changes look good

rhuairahrighairidh (approved):
tested a bit locally, looks good

PR #645, opened 2020-09-01
[R4R] address potential panic in validator vesting

karzak:
In preparation for kava-4, as well as the fact that the failure case for validator vesting is going to be hit, I reviewed the code that handles failed vesting periods. Of note, it panics in the case where an undelegation fails.

```
k.stakingKeeper.IterateDelegations(ctx, vv.Address, func(index int64, d stakingexported.DelegationI) (stop bool) {
			_, err := k.stakingKeeper.Undelegate(ctx, d.GetDelegatorAddr(), d.GetValidatorAddr(), d.GetShares())
			if err != nil {
				panic(err)
			}
			return false
		})
```
 
I think it's possible for undelegation to fail due to invalid shares amount (caused by rounding?), which we have seen on mainnet: https://kava.mintscan.io/txs/6E2DF48C550CD8674B4C1049A55F9434B6B9D02E75B2998FE63F33953C30C56B

Because of thus, I believe there is a small but nonzero chance of a panic when a validator vesting account fails a vesting period and its tokens are unbonded. 

Options
* Leave as is
* Silently ignore error
* Emit error event
* Something else?

### DIFF
--- a/x/validator-vesting/keeper/keeper.go
+++ b/x/validator-vesting/keeper/keeper.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/cosmos/cosmos-sdk/codec"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	stakingexported "github.com/cosmos/cosmos-sdk/x/staking/exported"
 
 	"github.com/tendermint/tendermint/libs/log"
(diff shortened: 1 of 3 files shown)


rhuairahrighairidh:
Good catch, looking at the Undelegate function there could be a few cases that cause it to error. Given this isn't super critical functionality I think emiting an event (or logging) and continuing might be the best option.

denalimarsh (approved):
LGTM 💯

rhuairahrighairidh (approved):

PR #649, opened 2020-09-12
[R4R] feat: Fetch auctions from historical state 

karzak:
Adds a fallback to the query auction by ID method which searches for the height of the most recent (ie, the last bid) bid message and returns the state of the auction at that height. Addresses one of the items in #639 

Tested successfully on a `kava-3` node

### DIFF
--- a/x/auction/client/common/query.go
+++ b/x/auction/client/common/query.go
@@ -80,7 +80,7 @@ func QueryAuctionByID(cliCtx context.CLIContext, cdc *codec.Codec, queryRoute st
 	}
 
 	queryCLIContext := cliCtx.WithHeight(maxHeight)
-	res, height, err = queryCLIContext.QueryWithData(fmt.Sprintf("custom/%s/%s", queryRoute, types.QueryNextAuctionID), nil)
+	res, height, err = queryCLIContext.QueryWithData(fmt.Sprintf("custom/%s/%s", queryRoute, types.QueryGetAuction), bz)
 	if err != nil {
 		return nil, 0, err
 	}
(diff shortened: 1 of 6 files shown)


denalimarsh (approved):
Does the historical block query go back to block 0? Are all blocks being iterated over, or all events?

x/auction/client/common/query.go
```diff
@@ -0,0 +1,92 @@
...
+	if auctionID >= nextAuctionID {
+		return nil, 0, sdkerrors.Wrapf(types.ErrAuctionNotFound, "%d", auctionID)
+	}
+
+	events := []string{
+		fmt.Sprintf("%s.%s='%s'", sdk.EventTypeMessage, sdk.AttributeKeyAction, "place_bid"),
+		fmt.Sprintf("%s.%s='%s'", types.EventTypeAuctionBid, types.AttributeKeyAuctionID, []byte(fmt.Sprintf("%d", auctionID))),
+	}
+
+	searchResult, err := utils.QueryTxsByEvents(cliCtx, events, defaultPage, defaultLimit)
```
rhuairahrighairidh, line 59: Does `QueryTxsByEvents` limit the returned results to `defaultLimit`?
If not a comment might be nice noting all txs are returned. If results are limited then this will miss auctions once we get more than 30 place bid msgs.

karzak: Updated the limit to 100 and added a comment that the results will not capture the ending auction state if there are >100 bids.

rhuairahrighairidh (approved):
looks good

PR #650, opened 2020-09-12
[R4R] fix link-checker: ignore appstore links

karzak:
link checker is complaining about frontier's link in the appstore - just added it to the ignore list.


denalimarsh (approved):
Thanks for this 💯

PR #651, opened 2020-09-13
[R4R] feat: payout liquid claims without creating vesting account

karzak:
Adds a check for incentive reward payouts where the vesting length is zero, avoiding creating a vesting account in that case.


denalimarsh (commented):
Looks good to me. Could you update `SendTimeLockedCoinsToAccount`'s function description to reflect the fact that not all recipients have their account type automatically converted?

PR #655, opened 2020-09-17
[R4R] feat: add variable length lockups for incentive rewards

karzak:


x/incentive/handler.go
```diff
@@ -30,7 +32,7 @@ func handleMsgClaimReward(ctx sdk.Context, k keeper.Keeper, msg types.MsgClaimRe
 	}
 
 	for _, claim := range claims {
-		err := k.PayoutClaim(ctx, claim.Owner, claim.CollateralType, claim.ClaimPeriodID)
+		err := k.PayoutClaim(ctx, claim.Owner, claim.CollateralType, claim.ClaimPeriodID, types.MultiplierName(strings.ToLower(msg.MultiplierName)))
```
rhuairahrighairidh, line 35: nit: The multipler name is parsed here and in types.msgs. You could move the parsing to a func like `NewMultiplerNameFromString`

x/incentive/types/params.go
```diff
@@ -177,3 +184,77 @@ func (rs Rewards) String() string {
 	}
 	return out
 }
+
+// Multiplier amount the claim rewards get increased by, along with how long the claim rewards are locked
+type Multiplier struct {
+	Name         MultiplierName `json:"name" yaml:"name"`
+	MonthsLockup int            `json:"months_lockup" yaml:"months_lockup"`
```
rhuairahrighairidh, line 191: `int` could introduce some indeterminsm over using `sdk.Int` or `int64`.
I think if someone's running a node on a super old 32 bit computer, golang uses int32 as the underlying type for int. Whereas on modern machines it uses int64.

rhuairahrighairidh (changes requested):
Left some initial comments, otherwise looks good.

We might need to be careful with multipler factors so we don't run out of funds in the incentive account. If everyone chooses high lockup with high multiple then funds could be used up faster than we're expecting. But even if that happens claim msgs will just fail rather than anything worse.

x/incentive/keeper/payout.go
```diff
@@ -24,7 +24,20 @@ func (k Keeper) PayoutClaim(ctx sdk.Context, addr sdk.AccAddress, collateralType
...
+	if !found {
+		return sdkerrors.Wrapf(types.ErrInvalidMultiplier, string(multiplierName))
+	}
+
+	rewardAmount := sdk.NewDecFromInt(claim.Reward.Amount).Mul(multiplier.Factor).RoundInt()
+	if rewardAmount.IsZero() {
+		return types.ErrZeroClaim
+	}
+	rewardCoin := sdk.NewCoin(claim.Reward.Denom, rewardAmount)
+	length := ctx.BlockTime().AddDate(0, multiplier.MonthsLockup, 0).Unix()
```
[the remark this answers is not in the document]
karzak: good catch, this should be `ctx.BlockTime().AddDate(0, multiplier.MonthsLockup, 0).Unix() - ctx.BlockTime().Unix()`

rhuairahrighairidh (approved):
looks good

PR #657, opened 2020-09-18
Fix swagger parsing errors

nddeluca:
There are some indentation and key errors.  This PR fixed those so swagger-ui succeeds in parsing the yaml.

Current rpc.kava.io:
Screen Shot 2020-09-18 at 12 00 09 PM


denalimarsh (approved):
LGTM!

PR #658, opened 2020-09-21
harvest v1

karzak:


rhuairahrighairidh (approved):
I think validator vesting accounts can send to module accounts. We could add a a blacklist check as is done in some of the other modules, but low severity.

PR #662, opened 2020-09-25
Bep3 migration

karzak:


migrate/v0_11/migrate.go
```diff
@@ -57,11 +57,92 @@ func MigrateBep3(oldGenState v0_9bep3.GenesisState) v0_11bep3.GenesisState {
...
+		0,
+		v0_11bep3.SupplyLimit{
+			Limit:          sdk.NewInt(10000000000), // 100 BTC limit at launch
+			TimeLimited:    false,
+			TimePeriod:     time.Duration(0),
+			TimeBasedLimit: sdk.ZeroInt()},
+		true,
+		mustAccAddressFromBech32("kava14qsmvzprqvhwmgql9fr0u3zv9n2qla8zhnm5pc"),
+		v0_9Params.BnbDeputyFixedFee,
+		sdk.OneInt(),
```
denalimarsh, line 73: The initial BEP3 module’s minimum fees should match the deputy min fees, which we've agreed will vary for different asset types. Here the minimum is set to `sdk.OneInt()` for all of the assets which will allow users to create swaps that the deputy may not relay.

**Scenario**: Deputy minimum fee for btcb is set to 10. The BEP3 module accepts a swap of 9 btcb but the Deputy will never relay it. The minimum swap amount for an asset should be equal to (or ideally greater than) the fixed deputy fee for that asset.

This is also an issue [here](https://github.com/Kava-Labs/kava/blob/master/migrate/v0_11/migrate.go#L25). For existing BEP3 assets (just bnb) lets set the new MinSwapAmount to the same value as the existing MinSwapAmount (1000 bnb).

migrate/v0_11/migrate.go
```diff
@@ -57,11 +57,92 @@ func MigrateBep3(oldGenState v0_9bep3.GenesisState) v0_11bep3.GenesisState {
...
+		"btcb",
+		0,
+		v0_11bep3.SupplyLimit{
+			Limit:          sdk.NewInt(10000000000), // 100 BTC limit at launch
+			TimeLimited:    false,
+			TimePeriod:     time.Duration(0),
+			TimeBasedLimit: sdk.ZeroInt()},
+		true,
+		mustAccAddressFromBech32("kava14qsmvzprqvhwmgql9fr0u3zv9n2qla8zhnm5pc"),
+		v0_9Params.BnbDeputyFixedFee,
```
denalimarsh, line 72: The BnbDeputyFixedFee amount should vary by asset type in order to make them the same value. Let's use the existing 1000 bnb as the standard.

migrate/v0_11/migrate.go
```diff
@@ -57,11 +57,92 @@ func MigrateBep3(oldGenState v0_9bep3.GenesisState) v0_11bep3.GenesisState {
...
+		144,
+		v0_11bep3.SupplyLimit{
+			Limit:          sdk.NewInt(1000000000000), // 1,000,000 XRP limit at launch
+			TimeLimited:    false,
+			TimePeriod:     time.Duration(0),
+			TimeBasedLimit: sdk.ZeroInt()},
+		true,
+		mustAccAddressFromBech32("kava1c0ju5vnwgpgxnrktfnkccuth9xqc68dcdpzpas"),
+		v0_9Params.BnbDeputyFixedFee,
+		sdk.OneInt(),
```
denalimarsh, line 97: ditto

migrate/v0_11/migrate.go
```diff
@@ -57,11 +57,92 @@ func MigrateBep3(oldGenState v0_9bep3.GenesisState) v0_11bep3.GenesisState {
...
+		sdk.NewCoin("btcb", sdk.ZeroInt()),
+		sdk.NewCoin("btcb", sdk.ZeroInt()),
+		time.Duration(0))
+	assetParams = append(assetParams, btcbAssetParam)
+	assetSupplies = append(assetSupplies, btcbAssetSupply)
+	xrpbAssetParam := v0_11bep3.NewAssetParam(
+		"xrpb",
+		144,
+		v0_11bep3.SupplyLimit{
+			Limit:          sdk.NewInt(1000000000000), // 1,000,000 XRP limit at launch
```
denalimarsh, line 90: 1 million XRP is worth $240,000 while 100 BTC is worth ~$1 million USD. This variance is very likely intentional (we're picking round numbers for asset amounts that are generally in the same USD ballpark?) but wanted to mention the difference.

migrate/v0_11/migrate.go
```diff
@@ -57,11 +57,92 @@ func MigrateBep3(oldGenState v0_9bep3.GenesisState) v0_11bep3.GenesisState {
...
+		727, // note - no official SLIP 44 ID
+		v0_11bep3.SupplyLimit{
+			Limit:          sdk.NewInt(10000000000000), // 100,000 BUSD limit at launch
+			TimeLimited:    false,
+			TimePeriod:     time.Duration(0),
+			TimeBasedLimit: sdk.ZeroInt()},
+		true,
+		mustAccAddressFromBech32("kava1hh4x3a4suu5zyaeauvmv7ypf7w9llwlfufjmuu"),
+		v0_9Params.BnbDeputyFixedFee,
+		sdk.OneInt(),
```
denalimarsh, line 121: ditto

denalimarsh (changes requested):
Looks generally good, but some attention should be paid to each asset's `FixedFee` and `MinSwapAmount`. `FixedFee` must match the Deputy's config file exactly and `MinSwapAmount` must be greater than the asset's `FixedFee`. Technically the min swap amount and fixed fee can be equal, but this allows swaps that will have exactly 0 coins sent to the recipient.

migrate/v0_11/migrate.go
karzak: Yes, this was intentional. The stability committee can adjust limits pretty rapidly, so these values are less important to get exactly correct.

migrate/v0_11/migrate.go
karzak: Good catch. I updated the fees for each asset to be ~$0.0025 and set the min swap amount to be one greater than the fee. Also of note, XRP is 8 decimals on binance chain, so updated to reflect that as well.

denalimarsh: Oh, this answers my question over in https://github.com/Kava-Labs/kava/pull/663 about XRP conversion factor.

migrate/v0_11/migrate.go
```diff
@@ -84,18 +84,18 @@ func MigrateBep3(oldGenState v0_9bep3.GenesisState) v0_11bep3.GenesisState {
...
 			TimeLimited:    false,
 			TimePeriod:     time.Duration(0),
 			TimeBasedLimit: sdk.ZeroInt()},
 		true,
 		mustAccAddressFromBech32("kava1c0ju5vnwgpgxnrktfnkccuth9xqc68dcdpzpas"),
-		v0_9Params.BnbDeputyFixedFee,
-		sdk.OneInt(),
-		sdk.NewInt(100000000000),
+		sdk.NewInt(100000), // 0.001 XRP fee
+		sdk.NewInt(10001),
```
denalimarsh, line 97: Wait, isn't this swap minimum amount of `10001` one factor less than the minimum fee of `100000`?

denalimarsh (changes requested):
Good catch on the XRP decimals, I noticed it over in the CDP module but not here. I think XRP's minimum swap amount is off, but other than that this looks good.

denalimarsh (approved):

rhuairahrighairidh (commented):
Stepped through coins movements, put notes in a [google doc](https://docs.google.com/document/d/1kPjvfJ1Kd0lRwn1h8E86OuN3Q8x0HpAdeDMyVy6xm7U/edit#).
Conclusion is: in kava-3, coins are in the bep3 module account while there is a swap that has not been claimed or refunded yet. In kava-4 coins are in the module account while there is an _**outoing**_ swap that has not been claimed or refunded yet.

So ideally the migration should delete incoming coins from the module account.
However if the deputy is offline early enough, and we refund all expired swaps, there will be no incoming swaps at migration.

Users could still create outgoing swaps, but with no deputy they'll just sit there and expire eventually.

Looks like we could also set assets as inactive to stop anyone creating swaps. But I think this will trigger a panic on init genesis bep3/genesis.go ln38.

PR #663, opened 2020-09-25
Cdp migration

karzak:


migrate/v0_11/migrate.go
```diff
@@ -65,3 +67,53 @@ func MigrateBep3(oldGenState v0_9bep3.GenesisState) v0_11bep3.GenesisState {
...
+
+	return v0_11cdp.NewGenesisState(
+		newParams,
+		newCDPs,
+		newDeposits,
+		newStartingID,
+		oldGenState.DebtDenom,
+		oldGenState.GovDenom,
+		oldGenState.PreviousDistributionTime,
+		sdk.ZeroInt(),
```
denalimarsh, line 116: This is a TODO for me. The `SavingsRateDistributed` should be set to something reasonably reflective of the actual savings rate distributed. Do you have any thoughts on how to do this calculation @karzak @rhuairahrighairidh ? 

Given the available data, my initial thought is that the best of the bad options would be to do a TVL query at each block. This is very time-intensive, as we've gotta add up the amount locked in each CDP each time. Then apply the global fee per block percentage multiplier to get the total block fees; then add all the total block fees. 90% of that amount is the savings rate distributed because 4.5% of the 5% fee is redistributed as the savings rate.

denalimarsh: Or, you know, just look at event logs or something 🥇

migrate/v0_11/migrate.go
```diff
@@ -65,3 +67,53 @@ func MigrateBep3(oldGenState v0_9bep3.GenesisState) v0_11bep3.GenesisState {
...
+	var newCDPs v0_11cdp.CDPs
+	var newDeposits v0_11cdp.Deposits
+	var newCollateralParams v0_11cdp.CollateralParams
+	newStartingID := uint64(0)
+
+	for _, cdp := range oldGenState.CDPs {
+		newCDP := v0_11cdp.NewCDPWithFees(cdp.ID, cdp.Owner, cdp.Collateral, "bnb-a", cdp.Principal, cdp.AccumulatedFees, cdp.FeesUpdated)
+		newCDPs = append(newCDPs, newCDP)
+		if cdp.ID >= newStartingID {
+			newStartingID = cdp.ID + 1
```
denalimarsh, line 82: If CDP ID 1005 is the most recently created CDP on kava-3 but it has been closed or liquidated (making ID 1004 the largest seen ID) then on kava-4 the first new CDP will share ID 1005 with a closed CDP from kava-3. Could this cause issues when querying state across kava-3/kava-4?

If we can't check event records here then this may be the most reasonable implementation though.

migrate/v0_11/migrate.go
```diff
@@ -65,3 +67,53 @@ func MigrateBep3(oldGenState v0_9bep3.GenesisState) v0_11bep3.GenesisState {
...
+	for _, dep := range oldGenState.Deposits {
+		newDep := v0_11cdp.NewDeposit(dep.CdpID, dep.Depositor, dep.Amount)
+		newDeposits = append(newDeposits, newDep)
+	}
+
+	for _, cp := range oldGenState.Params.CollateralParams {
+		newCollateralParam := v0_11cdp.NewCollateralParam(cp.Denom, "bnb-a", cp.LiquidationRatio, cp.DebtLimit, cp.StabilityFee, cp.AuctionSize, cp.LiquidationPenalty, 0x01, cp.SpotMarketID, cp.LiquidationMarketID, cp.ConversionFactor)
+		newCollateralParams = append(newCollateralParams, newCollateralParam)
+	}
+	btcbCollateralParam := v0_11cdp.NewCollateralParam("btcb", "btcb-a", sdk.MustNewDecFromStr("1.5"), sdk.NewCoin("usdx", sdk.NewInt(100000000000)), sdk.MustNewDecFromStr("1.000000001547125958"), sdk.NewInt(100000000), sdk.MustNewDecFromStr("0.075000000000000000"), 0x02, "btc:usd", "btc:usd:30", sdk.NewInt(8))
```
denalimarsh, line 95: I think that the market denom should match the on-chain asset denom here (applies to `xrp:usd` as well). Theoretically, `btcb` and `btc` could experience price variation which could cause issues if a user can transfer in `btcb` worth $10,000 and use it to take out a loan on Kava at a `btc` valuation of $12,000.

We've priced `busd` at `busd:usd` market prices for this reason - if `busd `is off its peg and is worth $1.02 but we treat it as USD always worth $1 then it creates an illegitimate arbitrage opportunity for users. Shouldn't `btcb`/`xrpb` be treated the same way?

migrate/v0_11/migrate.go
```diff
@@ -65,3 +67,53 @@ func MigrateBep3(oldGenState v0_9bep3.GenesisState) v0_11bep3.GenesisState {
 		PreviousBlockTime: v0_11bep3.DefaultPreviousBlockTime,
 	}
 }
+
+// MigrateCDP migrates from a v0.9 (or v0.10) cdp genesis state to a v0.11 cdp genesis state
+func MigrateCDP(oldGenState v0_9cdp.GenesisState) v0_11cdp.GenesisState {
+	var newCDPs v0_11cdp.CDPs
+	var newDeposits v0_11cdp.Deposits
+	var newCollateralParams v0_11cdp.CollateralParams
+	newStartingID := uint64(0)
```
denalimarsh, line 76: Super minor, but the [DefaultCdpStartingID](https://github.com/Kava-Labs/kava/blob/master/x/cdp/types/params.go#L37) is 1 in `x/cdp/types/params.go` so this should probably start at 1 as well.

migrate/v0_11/migrate.go
```diff
@@ -65,3 +67,53 @@ func MigrateBep3(oldGenState v0_9bep3.GenesisState) v0_11bep3.GenesisState {
...
+
+	for _, cdp := range oldGenState.CDPs {
+		newCDP := v0_11cdp.NewCDPWithFees(cdp.ID, cdp.Owner, cdp.Collateral, "bnb-a", cdp.Principal, cdp.AccumulatedFees, cdp.FeesUpdated)
```
denalimarsh, line 79: This seems reasonable as block time picks up where it left off, but I haven't tested it yet. Have we confirmed that this results in fees being updated correctly without any minor math hiccups?

denalimarsh (changes requested):
Looks good, most comments are minor or related to previous conversations. Requesting revisions/comment responses on:
1. XRP decimals/conversion factor
2. Potential implications of using `btc:usd` market price instead of `btcb:usd`

migrate/v0_11/migrate.go
karzak: There is no market liquid enough to try and do individual price feeds for the various pegged assets, and thus we assume that btc:btcb ~= 1. Trying to overcome this risk with oracles is risky, because all of the markets where btcb is traded are very thin - thus manipulating the price of the oracle is pretty cheap. FWIW, Maker and Compound assume that pegged assets trade 1:1 as well. https://compound.finance/docs/prices (ie, BTC markets on coinbase pro are used to determine the price of WBTC).

migrate/v0_11/migrate.go
karzak: One solution: Query the balance of the `savings` module account each block. Whenever the balance decreases, that corresponds to a payout of the savings rate. Add the decrease to a global counter and you should get an approximation of the total distributed.

migrate/v0_11/migrate.go
karzak: Not sure why I wrote it this way - we have access to the value directly in `oldGenState.StartingCdpID`.

migrate/v0_11/migrate.go
karzak: Yes, I've confirmed that the fee calculation is preformed correctly when the chain restarts. We got a nice natural test of this in fact when the chain went down.

denalimarsh (approved):

migrate/v0_11/migrate.go
```diff
@@ -65,3 +67,50 @@ func MigrateBep3(oldGenState v0_9bep3.GenesisState) v0_11bep3.GenesisState {
...
+	busdaCollateralParam := v0_11cdp.NewCollateralParam("busd", "busd-a", sdk.MustNewDecFromStr("1.01"), sdk.NewCoin("usdx", sdk.NewInt(3000000000000)), sdk.OneDec(), sdk.NewInt(1000000000000), sdk.MustNewDecFromStr("0.075000000000000000"), 0x03, "busd:usd", "busd:usd:30", sdk.NewInt(8))
+	busdbCollateralParam := v0_11cdp.NewCollateralParam("busd", "busd-b", sdk.MustNewDecFromStr("1.1"), sdk.NewCoin("usdx", sdk.NewInt(1000000000000)), sdk.MustNewDecFromStr("1.000000012857214317"), sdk.NewInt(1000000000000), sdk.MustNewDecFromStr("0.075000000000000000"), 0x04, "busd:usd", "busd:usd:30", sdk.NewInt(8))
+	xrpbCollateralParam := v0_11cdp.NewCollateralParam("xrpb", "xrpb-a", sdk.MustNewDecFromStr("1.5"), sdk.NewCoin("usdx", sdk.NewInt(100000000000)), sdk.MustNewDecFromStr("1.000000001547125958"), sdk.NewInt(100000000000), sdk.MustNewDecFromStr("0.075000000000000000"), 0x05, "xrp:usd", "xrp:usd:30", sdk.NewInt(8))
```
rhuairahrighairidh, line 95: auction size is `1000.00000000` XRP which is $240. The others are ~$10k.

rhuairahrighairidh (changes requested):
looks good, picked over the json output and it seems fine, except the xrp auction amount

PR #665, opened 2020-09-25
Committee migration

karzak:


rhuairahrighairidh (approved):
looks good
Do you think we need some committee control over harvest? Currently there's none.
Could add this to allowed_params:
```json
{
  "key": "Active",
  "subspace":"harvest"
},
```

PR #666, opened 2020-09-25
Incentive migration

karzak:
Multipliers may still be updated before finalizing


migrate/v0_11/migrate.go
```diff
@@ -65,3 +67,46 @@ func MigrateBep3(oldGenState v0_9bep3.GenesisState) v0_11bep3.GenesisState {
...
+		newClaims = append(newClaims, newClaim)
+	}
+
+	for _, oldClaimPeriodID := range oldGenState.NextClaimPeriodIDs {
+		newClaimPeriodID := v0_11incentive.GenesisClaimPeriodID{
+			CollateralType: oldClaimPeriodID.Denom + "-a",
+			ID:             oldClaimPeriodID.ID,
+		}
+		newClaimPeriodIds = append(newClaimPeriodIds, newClaimPeriodID)
+	}
```
rhuairahrighairidh, line 109: There's a ukava claim_period_id in the testdata that gets converted to ukava-a. Do you know if that will cause problems? I can't think of any.

karzak: Goofy, but won't cause an issue - don't see a need to delete it.

PR #668, opened 2020-09-26
[R4R] fix: close expired auctions in init genesis

karzak:
If an auction expires during migration, auction invariants will panic when the chain tries to restart. Fix is to close expired auctions in init genesis.

### DIFF
--- a/x/auction/genesis.go
+++ b/x/auction/genesis.go
@@ -35,6 +35,10 @@ func InitGenesis(ctx sdk.Context, keeper Keeper, supplyKeeper types.SupplyKeeper
 	if !moduleAcc.GetCoins().IsEqual(totalAuctionCoins) {
 		panic(fmt.Sprintf("total auction coins (%s) do not equal (%s) module account (%s) ", moduleAcc.GetCoins(), ModuleName, totalAuctionCoins))
 	}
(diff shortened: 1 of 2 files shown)


denalimarsh:
If I understand correctly:
1. Auctions expiring during migration causes panic
2. We're closing all already expired Auctions before migration

Doesn't this mean that any active Auctions will not be closed before migration and could still expire during migration?

What's the current sequence of manually triggered CDP/Auction module events for the migration? We can mitigate this issue by pausing CDP liquidations before migration and waiting until all Auctions have expired before proceeding with the migration, but in that case, this code addition is redundant.

rhuairahrighairidh:
Looks like this will panic if a debt auction expires during migration.
When a debt auction closes, `supplykeeper.MintCoins` is called, which calls `supplykeeper.GetSupply` which will panic as supply's init genesis hasn't run yet.

rhuairahrighairidh (changes requested):
Looks like this will panic if a debt auction expires during migration.
When a debt auction closes, supplykeeper.MintCoins is called, which calls supplykeeper.GetSupply which will panic as  the current supply hasn't been set yet by supply's init genesis.

rhuairahrighairidh (approved):
looks good

PR #672, opened 2020-09-28
Harvest Migration

karzak:


denalimarsh (approved):
LGTM 👍

rhuairahrighairidh (changes requested):
Code looks good, looked over the output json and it looks good.
Only thing is the reward amounts might be too high, the exponential amounts go a bit nuts after a couple of years.

migrate/v0_11/migrate.go
```diff
@@ -65,3 +66,27 @@ func MigrateBep3(oldGenState v0_9bep3.GenesisState) v0_11bep3.GenesisState {
...
+func MigrateHarvest() v0_11harvest.GenesisState {
+	// total HARD per second for lps (week one): 633761
+	// HARD per second for delegators (week one): 1267522
+	incentiveGoLiveDate := time.Date(2020, 10, 16, 14, 0, 0, 0, time.UTC)
+	incentiveEndDate := time.Date(2024, 10, 16, 14, 0, 0, 0, time.UTC)
+	claimEndDate := time.Date(2025, 10, 16, 14, 0, 0, 0, time.UTC)
+	harvestGS := v0_11harvest.NewGenesisState(v0_11harvest.NewParams(
+		true,
+		v0_11harvest.DistributionSchedules{
+			v0_11harvest.NewDistributionSchedule(true, "usdx", incentiveGoLiveDate, incentiveEndDate, sdk.NewCoin("hard", sdk.NewInt(310543)), claimEndDate, v0_11harvest.Multipliers{v0_11harvest.NewMultiplier(v0_11harvest.Small, 1, sdk.MustNewDecFromStr("0.33")), v0_11harvest.NewMultiplier(v0_11harvest.Large, 12, sdk.OneDec())}),
```
rhuairahrighairidh, line 80: If we want 1 month locked tokens, I think we need to use `v0_11harvest.Medium` multiplier name.
In`keeper.claimLPReward`, it calculates the time to lock tokens for using `keeper.GetPeriodLength`. And this always returns 0 for multiplier name `Small`.

karzak: This is a bug. I would prefer to fix it rather than hard code that `small` always means liquid.

PR #673, opened 2020-09-29
[R4R] harvest fixes

### LINKED ISSUE #671: Found typos in harvest module spec
I found some typos in [x/harvest/spec/05_params.md](https://github.com/Kava-Labs/kava/blob/master/x/harvest/spec/05_params.md).

I think the description of `Start` should be `the time when the period will start`, not `the time when the period will end`.

karzak:
Includes:
* Params validation enforcing that rewards are always `hard` tokens
* `NewMsgWithdraw` returned the wrong message type 😬
* Enforce that rewards don't accumulate before the `Start` or after `End` date for each distribution schedule 
* Various CLI parsing issues

Closes #671

### DIFF
--- a/x/harvest/client/cli/query.go
+++ b/x/harvest/client/cli/query.go
@@ -114,7 +114,7 @@ func queryModAccountsCmd(queryRoute string, cdc *codec.Codec) *cobra.Command {
 }
 
 func queryDepositsCmd(queryRoute string, cdc *codec.Codec) *cobra.Command {
-	return &cobra.Command{
+	cmd := &cobra.Command{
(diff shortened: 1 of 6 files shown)


rhuairahrighairidh (approved):

denalimarsh (approved):
LGTM.

A thought: if we're always requiring rewards to be `hard` tokens it makes sense to remove the param and hardcode it. The current implementation enables different reward assets in the future by modifying param validation, but that requires a hard fork anyways so we could add the reward denom param back in then. Just thinking out loud, not recommending we change it at this point.

PR #676, opened 2020-10-01
kava-3 to kava-4 migration

karzak:
tested against exported state from kava-3 at height 1392000 (included in testdata as `kava-3-export.json`


rhuairahrighairidh (approved):
looks good, only nit is to use the v09 codec for unmarshalling the old genstates, but I don't see any incompatibilites where the v011 codec is curently used.

PR #678, opened 2020-10-02
kava 4 migration doc

karzak:


migrate/v0_11/migrate.md
```diff
@@ -0,0 +1,76 @@
+# Kava-4 Upgrade Instructions
+
```
rhuairahrighairidh, line 2:
```suggestion
The kava chain is expected to halt at 1200 UTC, and restart with new software at 1400 UTC October 15th. **Validators must stop their nodes and upgrade their software at this time, and not before.** Otherwise they will go offline.

See the upgrade procedure below for more details.
```

rhuairahrighairidh (approved):
looks good, haven't tested things out locally yet.
Only suggestion would be to include super clear instructions that this upgrade happens synchronously at a particular time, and validators can't upgrade before.

PR #682, opened 2020-10-06
feat: add function to initialize testApp with chain-id

karzak:
Useful when validating a specific genesis file - for example, checking that a testnet genesis file is valid before publishing it

### DIFF
--- a/app/test_common.go
+++ b/app/test_common.go
@@ -142,6 +142,34 @@ func (tApp TestApp) InitializeFromGenesisStatesWithTime(genTime time.Time, genes
 	return tApp
 }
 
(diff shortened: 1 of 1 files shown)


denalimarsh (approved):
utACK.

Why introduce a new function as opposed to adding a chainID param to the original? Is the chain ID accessible in the `GenesisStates…` so we could default to it?

karzak:
> utACK.
> 
> Why introduce a new function as opposed to adding a chainID param to the original? Is the chain ID accessible in the `GenesisStates…` so we could default to it?

The chain-id is only accessible in the genesis document. When writing end-to-end tests, you generally only have a genesis state and don't care what the chain-id is because there are no gentxs (that's the only time it matters).

PR #696, opened 2020-10-19
Fix CDP Withdraw Query - Exact Args Increased to 3

nddeluca:
The command requires 3 arguments, but currently validates for an exact match of 2.


rhuairahrighairidh (approved):
looks good

denalimarsh (approved):
LGTM

PR #698, opened 2020-10-19
Query total fees collected

denalimarsh:
Addresses a query from https://github.com/Kava-Labs/kava/issues/639 by adding CDP store key `FeesCollected` to track the total amount of fees repaid (CDP module revenue) and exposes an endpoint for querying the value.

```bash
kvcli q cdp fees usdx
```

### DIFF
--- a/x/cdp/keeper/draw.go
+++ b/x/cdp/keeper/draw.go
@@ -95,26 +95,25 @@ func (k Keeper) RepayPrincipal(ctx sdk.Context, owner sdk.AccAddress, collateral
 	if err != nil {
 		return err
 	}
+
 	// send the payment from the sender to the cpd module
-	err = k.supplyKeeper.SendCoinsFromAccountToModule(ctx, owner, types.ModuleName, sdk.NewCoins(feePayment.Add(principalPayment)))
+	totalRepayment := principalPayment.Add(feePayment)
+	err = k.supplyKeeper.SendCoinsFromAccountToModule(ctx, owner, types.ModuleName, sdk.NewCoins(totalRepayment))
 	if err != nil {
 		return err
 	}
 
 	// burn the payment coins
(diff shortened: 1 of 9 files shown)


x/cdp/keeper/draw.go
```diff
@@ -144,7 +143,10 @@ func (k Keeper) RepayPrincipal(ctx sdk.Context, owner sdk.AccAddress, collateral
 	cdp.AccumulatedFees = cdp.AccumulatedFees.Sub(feePayment)
 
 	// decrement the total principal for the input collateral type
-	k.DecrementTotalPrincipal(ctx, cdp.Type, feePayment.Add(principalPayment))
+	k.DecrementTotalPrincipal(ctx, cdp.Type, totalRepayment)
+
+	// increment total fees collected
+	k.IncrementFeesCollected(ctx, cdp.Principal.Denom, feePayment.Amount)
```
karzak, line 149: I don't think this approach to calculating 'fees collected' is correct. The way the code is written, it is definitionally true that feesAssessed = feesRepaid. Either the CDP owner will call `RepayPrincipal`, which always collects fees before principal, or the CDP will get liquidated, which attempts to raise funds equal to `fees + principal`. Calling `IncrementFeesCollected` in `RepayPrincipal` doesn't tell us how much fees were collected in total, just how much was repaid this way vs getting liquidated. 

I think the best approach is to call `IncrementFeesCollected` in `UpdateFeesForAllCdps` and perhaps change the function to `IncrementFeesAssessed`

karzak (changes requested):
Code looks good - see comment about how we're calculating this and let me know if I'm thinking about it correctly.

x/cdp/keeper/draw.go
denalimarsh: That makes sense, I'll make those changes.

denalimarsh: However, I do think it's valuable to be able to differentiate between outstanding fees and past fees paid (protocol profit) and this approach doesn't handle that differentiation.

PR #700, opened 2020-10-22
fix: patch keys parse command

### LINKED ISSUE #647: kvcli keys parse return cosmos bech32 addresses
using `kvcli keys parse <HEX>` will return bech32 addresess prefixed with cosmos prefixes (cosmos1 / comospub1)
As the prefix will also affect the checksum getting the correct kava address from the output is not trivial

karzak:
Closes #647


rhuairahrighairidh (approved):
looks good

PR #715, opened 2020-11-10
Harvest: borrows limited by global asset borrow limit

denalimarsh:
This PR implements functionality for [this task](https://app.asana.com/0/1156716699555540/1198999905425415/f).

Changes:
- Add `BorrowedCoins` to store + querier
- Increment `BorrowedCoins` balance after each successful borrow
- Validate that proposed borrows won't exceed the asset's limit

The param is set in each asset's MoneyMarket.BorrowLimit.MaximumLimitUSD.

### DIFF
--- a/x/harvest/keeper/borrow.go
+++ b/x/harvest/keeper/borrow.go
@@ -59,6 +59,10 @@ func (k Keeper) Borrow(ctx sdk.Context, borrower sdk.AccAddress, coins sdk.Coins
 
 // ValidateBorrow validates a borrow request against borrower and protocol requirements
 func (k Keeper) ValidateBorrow(ctx sdk.Context, borrower sdk.AccAddress, amount sdk.Coins) error {
+	if amount.IsZero() {
+		return types.ErrBorrowEmptyCoins
+	}
+
 	// Get the proposed borrow USD value
 	moneyMarketCache := map[string]types.MoneyMarket{}
 	proprosedBorrowUSDValue := sdk.ZeroDec()
--- a/x/harvest/keeper/borrow.go
+++ b/x/harvest/keeper/borrow.go
@@ -1,6 +1,7 @@
 package keeper
 
 import (
+	"fmt"
 	"strings"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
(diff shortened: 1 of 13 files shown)


x/harvest/types/params.go
```diff
@@ -223,22 +223,22 @@ type Multipliers []Multiplier
 
 // BorrowLimit enforces restrictions on a money market
 type BorrowLimit struct {
-	MaximumLimit sdk.Int `json:"maximum_limit" yaml:"maximum_limit"`
-	LoanToValue  sdk.Dec `json:"loan_to_value" yaml:"loan_to_value"`
+	MaximumLimitUSD sdk.Dec `json:"maximum_limit_usd" yaml:"maximum_limit_usd"`
```
karzak, line 226: This should just be in native units, not usd. Further, the max borrow limit should be an optional flag. Something like

```
type BorrowLimit struct {
        HasMaxLimit bool `json:"has_max_limit" yaml:"has_max_limit"`
	MaximumLimitUSD sdk.Int `json:"maximum_limit_usd" yaml:"maximum_limit_usd"`
	LoanToValue     sdk.Dec `json:"loan_to_value" yaml:"loan_to_value"`
}

File: x/harvest/keeper/borrow.go [hunk 2]
```diff
@@ -61,6 +71,22 @@ func (k Keeper) ValidateBorrow(ctx sdk.Context, borrower sdk.AccAddress, amount
 			return sdkerrors.Wrapf(types.ErrPriceNotFound, "no price found for market %s", moneyMarket.SpotMarketID)
 		}
 		coinUSDValue := sdk.NewDecFromInt(coin.Amount).Quo(sdk.NewDecFromInt(moneyMarket.ConversionFactor)).Mul(assetPriceInfo.Price)
+
+		// Validate the requested borrow value for the asset against the money market's global borrow limit
```
karzak, line 75: Refactor this validation logic based on `BorrowLimit` params comment

karzak (approved):
Approach looks correct and code looks clean. Just refactor the borrow limit to make it optional and native units.

The reason for native units is because we can measure the liquidity of a coin in native units, and that's what we're really concerned with. If the market can tolerate liquidations of 100k units, it doesn't matter if the USD value of those units fluctuate - so we wouldn't want borrow restrictions to happen just because the USD value of something increases.

x/harvest/keeper/borrow.go
denalimarsh: https://github.com/Kava-Labs/kava/pull/715/commits/777fc3e26d0db82666294056b9b923a1f0061f92

x/harvest/types/params.go
denalimarsh: Added in https://github.com/Kava-Labs/kava/pull/715/commits/777fc3e26d0db82666294056b9b923a1f0061f92

karzak (approved):

PR #717, opened 2020-11-12
Add authorized mempool

rhuairahrighairidh:
Add a (optional) filter on incoming txs to the mempool that rejects all not signed by a known list of addresses.

The list of addresses is the union of
- bep3 deputies
- pricefeed oracles
- addresses added in the config file `app.toml`

If this feature is enabled by a few validators, txs from authorized addresses will get into blocks even if the network is under high load.
Everytime one of these validators propose a block, it will contain only authorized txs, regardless of the size of the standard mempool.

To enable, add these lines to `app.toml`:
```toml
[mempool]
authorized-addresses = [
  "kava173w2zz287s36ewnnkf4mjansnthnnsz7rtrxqc",
]
enable-authentication = true
```

Implementation Notes:
- The antehandler used to be defined in `cosmos-sdk/x/auth/ante`. It has been moved to `app/ante`, and the new authorization filter added.
- This PR contains a slight refactor of the `app.NewApp` function. There were several arguments piling up so I gathered them into an `AppOption` struct. The default value `AppOptions{}` can be used in most cases.

TODO:
- merge in go1.15 test fixes
- decide where to merge this to (master or a non-breaking-release branch?)

Extensions
- The antehandler could track the size of the mempool, by counting `CheckTx` and `DeliverTx` requests. It's a bit hacky but would allow us to do things like only enable authentication when there is more than 5 txs in the mempool.


cmd/kvd/main.go
```diff
@@ -28,7 +29,11 @@ import (
 )
 
 // kvd custom flags
-const flagInvCheckPeriod = "inv-check-period"
+const (
+	flagInvCheckPeriod         = "inv-check-period"
+	configMempoolEnableAuth    = "mempool.enable_authentication"
```
karzak, line 34: Not sure how easy this is, but I would include a default config option in app.toml when `kvd init` is called for the first time. So, if installing from empty, `app.toml` would have 

```
[mempool]
enable_authentication = false
authorized_addresses = []
```

Additional nit:  Mempool options could also be added as flags to the `kvd start` command. It seems like that is the general pattern used, where options can be specified is `app.toml` or passed as flags to `kvd start`. Not a huge deal if we don't follow that pattern, but thought I would mention it.

karzak (commented):
Tested ACK. Functionality working as expected. Lefts a few comments/nits but nothing blocking.

app/ante/authorized.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	return addrs
+}
+
+// commonAddressesExist checks if there is any interection between two lists of addresses
+func commonAddressesExist(addresses1, addresses2 []sdk.AccAddress) bool {
+	for _, a1 := range addresses1 {
+		for _, a2 := range addresses2 {
+			if a1.Equals(a2) {
+				return false
```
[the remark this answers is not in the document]
rhuairahrighairidh: oops, I changed the name at some point and didn't switch around the returned values.

cmd/kvd/main.go
rhuairahrighairidh: Adding the stuff to `kvd init` is awkward. The config and flags in cosmos are a bit all over the place, and mostly of hard coded. However adding the additional flags is easy enough.
So I've added flags for the mempool options which at least give some visibility and documentation.

PR #734, opened 2020-12-07
Add 'InterestFactor' to CDP type

karzak:
Updates the CDP type to include an 'InterestFactor' field which represents the value of the global interest factor at the time the CDP was last updated (ie, when interest was last calculated and updated on the object).

This PR is in preparation of moving off of updating fees in the begin blocker. I didn't delete any fields from the CDP object, as I think they are still useful when querying. If you would argue to remove some fields as part of this refactor, comment on this PR. We will also do a larger review/audit once the `cdp-accumulators` branch is ready for merging into master.

### DIFF
--- a/migrate/v0_11/migrate.go
+++ b/migrate/v0_11/migrate.go
@@ -670,7 +670,7 @@ func MigrateCDP(oldGenState v0_9cdp.GenesisState) v0_11cdp.GenesisState {
 	newStartingID := oldGenState.StartingCdpID
 
 	for _, cdp := range oldGenState.CDPs {
-		newCDP := v0_11cdp.NewCDPWithFees(cdp.ID, cdp.Owner, cdp.Collateral, "bnb-a", cdp.Principal, cdp.AccumulatedFees, cdp.FeesUpdated)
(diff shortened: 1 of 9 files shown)


denalimarsh (commented):
Some thoughts:
- Will the accumulator include a global ‘interest factor’ for each collateral denom (similar to the Hard module)?
- In the Hard module we're using the term "BorrowIndex". Assuming the two operate as interest accumulators we should probably standardize the naming; I prefer "InterestFactor".
- Perhaps outside the scope here, but how are we approaching CDP liquidation in the new architecture?

karzak:
> * Will the accumulator include a global ‘interest factor’ for each collateral denom (similar to the Hard module)?
Yes
> * In the Hard module we're using the term "BorrowIndex". Assuming the two operate as interest accumulators we should probably standardize the naming; I prefer "InterestFactor".
+1 to standardizing on 'InterestFactor' - I think it's clearer. 
> * Perhaps outside the scope here, but how are we approaching CDP liquidation in the new architecture?
Similar approach to Hard - attempt to liquidate in begin blocker, add a message for liquidation than can be called by keepers

PR #735, opened 2020-12-08
Add cdp accumulator methods

karzak:
Addressed comments and added more robust tests.


x/cdp/keeper/interest.go
```diff
@@ -9,19 +9,80 @@ import (
...
+	}
+
+	borrowRateSpy := k.getFeeRate(ctx, ctype)
+	if borrowRateSpy.Equal(sdk.OneDec()) {
+		return nil
+	}
+	interestFactor := CalculateInterestFactor(borrowRateSpy, sdk.NewInt(timeElapsed))
+	interestAccumulated := (interestFactor.Mul(totalPrincipalPrior.ToDec())).RoundInt().Sub(totalPrincipalPrior)
+	if interestAccumulated.IsZero() {
+		// in the case accumulated interest rounds to zero, exit early without updating accrual time
```
rhuairahrighairidh, line 52: Accural time might need to be updated every block?

Example: The fee rate is set to 1.00 by governance for some time. So `PreviousAccuralTime` isn't set for a while. Then governance sets the fee rate back to normal (>1). Then this func will run using the old accural time but using the new fee rate. It would calculate fees accruing over the period when the rate was 1.00 when no fees should accrue.

x/cdp/keeper/interest.go
```diff
@@ -98,6 +159,21 @@ func (k Keeper) UpdateFeesForAllCdps(ctx sdk.Context, collateralType string) err
...
 }
 
+// CalculateFees returns the fees accumulated since fees were last calculated based on
+// the input amount of outstanding debt (principal) and the number of periods (seconds) that have passed
+func (k Keeper) CalculateFees(ctx sdk.Context, principal sdk.Coin, periods sdk.Int, collateralType string) sdk.Coin {
+	// how fees are calculated:
+	// feesAccumulated = (outstandingDebt * (feeRate^periods)) - outstandingDebt
+	// Note that since we can't do x^y using sdk.Decimal, we are converting to int and using RelativePow
+	feePerSecond := k.getFeeRate(ctx, collateralType)
+	scalar := sdk.NewInt(1000000000000000000)
```
denalimarsh, line 169: Use the `scalingFactor` var declared on line 13.

x/cdp/keeper/keeper.go
```diff
@@ -130,3 +131,41 @@ func (k Keeper) GetSavingsRateDistributed(ctx sdk.Context) sdk.Int {
...
+}
+
+// GetInterestFactor returns the current interest factor for an individual collateral type
+func (k Keeper) GetInterestFactor(ctx sdk.Context, ctype string) (sdk.Dec, bool) {
+	store := prefix.NewStore(ctx.KVStore(k.key), types.InterestFactorPrefix)
+	bz := store.Get([]byte(ctype))
+	if bz == nil {
+		return sdk.ZeroDec(), false
+	}
+	var iterestFactor sdk.Dec
```
denalimarsh, line 161:
```suggestion
	var interestFactor sdk.Dec
```

x/cdp/keeper/interest.go
```diff
@@ -9,19 +9,80 @@ import (
...
+
+	totalPrincipalPrior := k.GetTotalPrincipal(ctx, ctype, types.DefaultStableDenom)
+	if totalPrincipalPrior.IsZero() || totalPrincipalPrior.IsNegative() {
+		return nil
+	}
+
+	interestFactorPrior, foundInterestFactorPrior := k.GetInterestFactor(ctx, ctype)
+	if !foundInterestFactorPrior {
+		interestFactorPrior = sdk.OneDec()
+		k.SetInterestFactor(ctx, ctype, interestFactorPrior)
```
denalimarsh, line 39:
```suggestion
		k.SetInterestFactor(ctx, ctype, sdk.OneDec())
```

No need to set `interestFactorPrior` since we return at the end of this condition

rhuairahrighairidh (commented):
Overall direction looks good to me.

Left a possible error case in `AccumulateInterest`

denalimarsh (commented):
utACK, agree with the direction. There are some good approaches here that can be applied back over to the Hard module when we standardize accumulators.

x/cdp/keeper/interest.go
denalimarsh: Do param updates take effect before the BeginBlocker runs? If so, wouldn't any updates to the InterestFee param retroactively apply the new InterestFee rate to the last period? Such that when the fee rate is increased from 1.00 it'll still apply fees for a 1 block period for which the rate was 1.00 and when no fees should have accrued.

denalimarsh: Here I think we're trying to prevent the situation in which the 1-block accumulated interest rounds to 0 and no interest is applied over a long sequence of blocks during which the interest rate is > 1. Can we add a test that identifies the (probably low) interest rate/total principal amount thresholds at which this is a problem?

karzak: Resolved in 80651c3. Handles the case where interest is rounded to zero and doesn't update `PreviousAccuralTime`. In cases where the APY is 0%, does update `PreviousAccuralTime`.

x/cdp/keeper/interest.go
karzak: 80651c3

x/cdp/keeper/keeper.go
karzak: 80651c3

x/cdp/keeper/interest.go
karzak: 80651c3

denalimarsh (approved):
Changes LGTM.

PR #742, opened 2020-12-15
Hard: introduce LTV index

denalimarsh:
This PR introduces an LTV index to the Hard module. User interactions (deposit/borrow/repay) update their address' position in the LTV index. If the scaffolding looks good, adding the liquidation engine will be straightforward.

Notes:
- LTV index iteration is currently low -> high, should probably be high -> low

Discussion points:
- Highest LTV doesn't mean most likely to be liquidated. The user with the smallest difference between their (last known LTV ratio, weighted LTV liquidation threshold ratio) is most likely to get liquidated. This isn't reflected in the current approach.


denalimarsh:
Updating branch to master made the git history muddled, but the file diff is still correct

x/harvest/keeper/borrow.go
```diff
@@ -19,11 +19,31 @@ func (k Keeper) Borrow(ctx sdk.Context, borrower sdk.AccAddress, coins sdk.Coins
...
+	if !found {
+		// On user's first borrow, build borrow index list containing denoms and current global borrow index value
+		var borrowIndexes types.BorrowIndexes
+		for _, coin := range coins {
+			borrowIndexValue, _ := k.GetBorrowIndex(ctx, coin.Denom)
+			borrowIndex := types.NewBorrowIndexItem(coin.Denom, borrowIndexValue)
+			borrowIndexes = append(borrowIndexes, borrowIndex)
+		}
+		borrow := types.NewBorrow(borrower, sdk.Coins{}, borrowIndexes)
+		k.SetBorrow(ctx, borrow)
```
karzak, line 39: I believe `SetBorrow` is only safe to call after `SendCoinsFromModuleToAccount` or some equivalent check like a `validateBalance` function is called and `validateBorrow` has passed. As currently written, the borrow would get created in the store even if the user didn't have `coins` in their account or the LTV was invalid.

karzak (changes requested):
Scaffolding looks good, need to address the potentially unsafe call to `SetBarrow`.

x/harvest/keeper/borrow.go
denalimarsh: Fixed in https://github.com/Kava-Labs/kava/pull/742/commits/e32ec93b237a7a0f5a4f36110c5d72a6162f9cfe

x/harvest/keeper/liquidation.go
```diff
@@ -253,6 +253,78 @@ func (k Keeper) StartAuctions(ctx sdk.Context, borrower sdk.AccAddress, borrows,
...
+func (k Keeper) GetCurrentLTV(ctx sdk.Context, addr sdk.AccAddress) (sdk.Dec, error) {
+	// Fetch deposits and parse coin denoms
+	deposits := k.GetDepositsByUser(ctx, addr)
+	depositDenoms := []string{}
+	for _, deposit := range deposits {
+		depositDenoms = append(depositDenoms, deposit.Amount.Denom)
+	}
+
+	// Fetch borrow balances and parse coin denoms
+	borrowBalances := k.GetBorrowBalance(ctx, addr)
```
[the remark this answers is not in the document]
denalimarsh: Added in https://github.com/Kava-Labs/kava/pull/742/commits/e32ec93b237a7a0f5a4f36110c5d72a6162f9cfe

x/harvest/keeper/borrow.go
```diff
@@ -19,11 +19,20 @@ func (k Keeper) Borrow(ctx sdk.Context, borrower sdk.AccAddress, coins sdk.Coins
 		}
```
karzak, line 19: Just realized this, but we should check that the the money market exists in params before calling `k.SetBorrowIndex`, as a user could input a coin that doesn't exist and we would still store a borrow index for it.

denalimarsh: Good catch, updated in https://github.com/Kava-Labs/kava/pull/742/commits/75c2fcd81848211ff3f15d17f3b7437abc50f222

x/harvest/keeper/liquidation.go
```diff
@@ -253,6 +253,81 @@ func (k Keeper) StartAuctions(ctx sdk.Context, borrower sdk.AccAddress, borrows,
...
+// UpdateItemInLtvIndex updates the key a borrower's address is stored under in the LTV index
+func (k Keeper) UpdateItemInLtvIndex(ctx sdk.Context, prevLtv sdk.Dec, borrower sdk.AccAddress) error {
+	currLtv, err := k.GetCurrentLTV(ctx, borrower)
+	if err != nil {
+		return err
+	}
+
+	k.RemoveFromLtvIndex(ctx, prevLtv, borrower)
+	// If the user doesn't have any borrows their LTV is 0
+	if currLtv.GT(sdk.ZeroDec()) {
```
karzak, line 325: Question: Is it possible for a user to have a micro borrow where the ltv rounds to zero and thus we don't store it in the index? If that happened, could it cause a panic/unexpected error somewhere?

karzak (approved):
LGTM, just a question. If the `microloan` is a problem, we may want to introduce min borrow amount in a follow on PR.

x/harvest/keeper/liquidation.go
denalimarsh: Good point. I updated `GetCurrentLTV` with a bool that indicates if the LTV should be added to the index instead of using `sdk.ZeroDec()` as a bool for this conditional in https://github.com/Kava-Labs/kava/pull/742/commits/8c2bb779ad0a7b7efa681dd12753efa77ad5de54.

Additionally, I'm using the bool returned by the function when getting `prevLtv` to determine if the borrower should be removed from the index, because only existing items in the index should be removed. This would have failed silently without issues because attempting to remove a non-existent item from the index doesn't return an error, but probably still better to prevent it from happening.

karzak (approved):

PR #750, opened 2020-12-24
Add cdp liquidations by external keeper

karzak:
Features: 
* Adds a `MsgLiquidate` where an external party can cause a liquidation for a risky cdp. If successful, the keeper is rewarded a percentage of the collateral according to `KeeperRewardPercentage` parameter.
* Adds a `SynchronizeInterestForRiskyCDPs` function in the begin blocker that will update fees for the `n` riskiest cdps in the store (cdps with the lowest collateral:debt ratio). `n` is set according to the `CheckCollateralizationIndexCount` parameter. If any CDPs that are touched by `SynchronizeInterestForRiskyCDPs` go below the required collateralization ratio, they will be liquidated in the begin blocker.


x/cdp/keeper/seize.go
```diff
@@ -93,7 +115,49 @@ func (k Keeper) ApplyLiquidationPenalty(ctx sdk.Context, collateralType string,
...
+	}
+
+	deposits := k.GetDeposits(ctx, cdp.ID)
+	for _, dep := range deposits {
+		if dep.Amount.IsGTE(rewardCoin) {
+			dep.Amount = dep.Amount.Sub(rewardCoin)
+			k.SetDeposit(ctx, dep)
+			break
+		}
+	}
```
rhuairahrighairidh, line 161: If no deposit is greater than the `rewardCoin` then the `if` is never triggered and the reward is not subtracted.

It would leave cdp.Collateral not equal to the sum of deposits. In SeizeCollateral, auctions are started from deposits, rather than cdp.Collateral, so deposits would be auctioned without the reward subtracted. End result is too much collateral is sent out from the cdp module account.

Seems really unlikely to occur naturally, but could be forced by a user.

rhuairahrighairidh (changes requested):
Looks good, only request is to patch a super edge case in paying out liquidation rewards when there's many small deposits. Unlikely to be hit, but it would interfere with the accounting.

rhuairahrighairidh (approved):
looks good

PR #752, opened 2020-12-30
USDX incentive accumulators

karzak:
Blocked by #750

Refactors USDX incentives to use accumulators. Accomplishes this by:

* Introducing `hooks` on the CDP module, which are called after a cdp is created or before it is modified.
* Refactors `Claim` objects to be 1 object per USDX minter, rather than 1 per eligible reward type
* Whenever a CDP is opened, a `USDXMintingClaim` is created/updated that tracks the users rewards 
* Whenever a CDP is modified (directly after interest is synced, to be precise), the USDX minting rewards are synced. 
* When a user claims rewards, USDX minting rewards are synced, and set back to zero once they are claimed. 

The goal of this PR is also to generalize the `incentive` module such that it could be used for other incentives, such as hard v2 rewards.


rhuairahrighairidh (approved):
Looks good, couldn't find a way to break anything.
Scenarios I considered:
- multiple cdps per owner, opening and closing
- multiple cdp and incentive msgs within a single tx
- param changes
- chain halts and genesis export/imports

Looks like the reward calculation is slightly off the true value, but I think the difference is so small I'm leaving it as a nit.

A CDP's rewards per block is based on its debt in that block. But the debt gradually increases. The code uses the cdp debt for the current block to calculate rewards for all blocks since the claim was last synced. So the claimed rewards will be a bit too high. And there is a (small) incentive to rarely update a cdp to maximse the error.

I fixed the usdx incentive section in the optimization doc: https://docs.google.com/document/d/1JB2fGhNC5Mxulnklub_8Oq38KjyYf4uBL14Z_llFjKM
I think this is all required for a fix:
- multiply the reward factor by the cdp accumulator every bloc
  - ie `newRewardFactor = oldRewardFactor + cdpFactor*(rewards/totalPrincipal)`)
  - also make sure cdp begin block runs before incentive
- run `incentiveKeeper.SynchronizeReward` _before_ `cdpKeeper.SynchronizeInterest`. It needs to update the rewards using the old cdp principal.

Question: where does the incentive module account get its coins from?

nits:
- check `sdk.Int` and `sdk.Dec` params aren't nil in genesis validation to avoid the slightly confusing panics when a genesis file hasn't been setup correctly.
- might be useful to have an option in the querier that returns synced claims so users can see how much they could claim now
- `Active` params aren't hooked up to the module. Could cause mild confusion for future devs/governors.

karzak:
> Looks like the reward calculation is slightly off the true value, but I think the difference is so small I'm leaving it as a nit.
> 
> A CDP's rewards per block is based on its debt in that block. But the debt gradually increases. The code uses the cdp debt for the current block to calculate rewards for all blocks since the claim was last synced. So the claimed rewards will be a bit too high. And there is a (small) incentive to rarely update a cdp to maximse the error.
> 
> I fixed the usdx incentive section in the optimization doc: https://docs.google.com/document/d/1JB2fGhNC5Mxulnklub_8Oq38KjyYf4uBL14Z_llFjKM
> I think this is all required for a fix:
> 
> * multiply the reward factor by the cdp accumulator every bloc
>   
>   * ie `newRewardFactor = oldRewardFactor + cdpFactor*(rewards/totalPrincipal)`)
>   * also make sure cdp begin block runs before incentive
> * run `incentiveKeeper.SynchronizeReward` _before_ `cdpKeeper.SynchronizeInterest`. It needs to update the rewards using the old cdp principal.
> 

I'm not sure that ` cdpFactor*(rewards/totalPrincipal))` is necessary because the totalPrincipal returned by the cdp module already accounts for the accumulator. See https://github.com/Kava-Labs/kava/blob/c5b05af36d34b852bb4dc08e1219ae1a01ea34cb/x/cdp/keeper/interest.go#L89, where the total principal is scaled each block by the accumulator. 

Question: where does the incentive module account get its coins from?

The `kavadist` module account, which mints coins each blocks according to the inflation schedule we outlined in gov prop.

PR #753, opened 2021-01-03
[DRAFT] Implement USDX savings rate using bank hooks

karzak:
In order to re-integrate the USDX savings rate using accumulators, I've drafted an implementation that uses hooks on the bank module. The hook works as follows

- Before any transfer, if the transfer involves USDX, the bank module checks if the sender and/or receiver have accumulated any USDX savings rate. If they have, it transfers that amount to their account, prior to the send. This applies to any `MsgTransfer`, but also all other transfers of coins, because they all use the bank keeper (except `MsgMultiSend`, the implementation of which I've left out in this draft, but can be added with minimal effort).
- USDX savings rate is only paid out to non-module accounts
- When the USDX savings rate is paid out, the users claim object is synchronized with the global savings rate factor. 
- Note that when USDX is sent from the savings module account, the hook exits immediately, otherwise you would get an infinite loop (transfer calls bank hook, which transfers USDX from the savings module account, which calls bank hook, etc...) 

Status:
- Basic tests passing
- Successfully tested functionality using kvtool 

Remaining work
- Thorough review for security and edge-case concerns
- Implement MsgMultisend hook
- Improve tests
- Get original cosmos-sdk `x/bank` tests passing (need to remove use of `simapp`, as it's not compatible with forked bank module). 
- Implement message where users can claim their USDX savings? Currently, it happens automatically when a transfer involving usdx hits that account.

Biggest Open Questions:
- Is this the best strategy for implementation?
- Do bank hooks introduce any new security concerns?


rhuairahrighairidh (commented):
neat solution, some thoughts
- my main concern is if the additional complexity/maintenance cost is worth it for the savings rate feature. Triggering hooks in bank transfers adds more to think about when doing reviews. They're a slightly unintuitive thing for new people to learn about.
  - An alternative would be to force people to deposit their usdx somewhere to accumulate savings. But that doesn't seem ideal either.
- worth checking if the hooks cause slow downs in other modules, and doing some optimising if the do. `SyncUSDXSavingsRateBeforeTransfer` adds at least one extra account unmarshalling to every send.
- It would be nice to have the hooks on lower level bank functions (like Add/SubtractCoins) to avoid the possibility of a module bypassing the hooks by not using SendCoins. It could mess up the savings calculations. But I can't think how to do this without infinite loops.

karzak:
closing in favor of #764

PR #754, opened 2021-01-04
CDP begin blocker optimization 1

rhuairahrighairidh:
This PR makes non breaking changes to the cdp BeginBlocker to increase speed.

Running the go profiler on the begin blocker showed a large portion of time was spent unmarshaling the cdp params.
They were being unmarshalled twice in `k.CalculateCollateralToDebtRatio` and once in `k.CalculateFees`.
However the param values should never change during the begin blocker. This PR refactors the code to read the params at the start of the loop and then reference those unmarshalled values within the loop.

Benchmark results running locally (with 2000 cdps):

    BenchmarkBeginBlocker-4                2        1737453770 ns/op
    BenchmarkBeginBlocker-4                2         760821978 ns/op
    BenchmarkBeginBlocker-4                1        1028420038 ns/op
    BenchmarkBeginBlocker-4                1        1178868937 ns/op
    BenchmarkBeginBlocker-4                1        1043643995 ns/op

average: 1.149841744 s

    BenchmarkBeginBlocker-4               28          45595337 ns/op
    BenchmarkBeginBlocker-4               26          44801298 ns/op
    BenchmarkBeginBlocker-4               25          42354691 ns/op
    BenchmarkBeginBlocker-4               27          42985697 ns/op
    BenchmarkBeginBlocker-4               27          54006477 ns/op
    BenchmarkBeginBlocker-4               18          64251118 ns/op
    BenchmarkBeginBlocker-4               31          39956663 ns/op

average: 0.047707326s

Recent profiling on mainnet showed the cdp begin blocker took 2.2s, harvest 0.8s, other modules combined < 0.2s.

### DIFF
--- a/x/cdp/abci_test.go
+++ b/x/cdp/abci_test.go
@@ -56,10 +56,10 @@ func (suite *ModuleTestSuite) SetupTest() {
 	suite.addrs = addrs
 }
 
-func createCDPs(ctx sdk.Context, keeper cdp.Keeper, addrs []sdk.AccAddress) (liquidationTracker, error) {
+func createCDPs(ctx sdk.Context, keeper cdp.Keeper, addrs []sdk.AccAddress, numCDPs int) (liquidationTracker, error) {
 	tracker := liquidationTracker{}
 
-	for j := 0; j < 100; j++ {
+	for j := 0; j < numCDPs; j++ {
 		var collateral string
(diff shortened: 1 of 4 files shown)


karzak (approved):
tACK. Tested on production node and was working properly.

PR #755, opened 2021-01-05
Hard: Hard <-> Incentive module hooks

denalimarsh:
`kd-usdx-incentive-accumulators` doesn't include the latest Hard code (was still called Harvest) so instead of a messy rebase I just plugged the latest version of the Hard module. The relevant changes are in hard/types/expected_keepers.go, hard/types/hooks.go, hard/keeper/keeper.go, and hard/keeper/hooks.go.

- Doesn't compile due to migrations: `x/hard/legacy/v0_11` has been removed in master, there's just v0_12. What's the appropriate way to reconcile this?


denalimarsh:
Closed in favor of https://github.com/Kava-Labs/kava/pull/756

PR #758, opened 2021-01-07
Hard: LTV index refactor

denalimarsh:
Goal: Breakout Set/Update/Delete as separate functions for the LTV index.

The Hard LTV index is different than the CDP index as it's based upon two store objects (Borrow, Deposit) which can be updated individually as opposed to just one object (CDP), so the resulting implementation is not 1:1 with the goal. The methods added are:
- UpdateBorrowAndLtvIndex
- UpdateDepositAndLtvIndex
- DeleteDepositBorrowAndLtvIndex

Noticed some potential issues with attempted keeper liquidations that I created Asana cards for.

### DIFF
--- a/x/hard/keeper/liquidation.go
+++ b/x/hard/keeper/liquidation.go
@@ -22,10 +22,10 @@ func (k Keeper) AttemptIndexLiquidations(ctx sdk.Context) error {
 	borrowers := k.GetLtvIndexSlice(ctx, params.CheckLtvIndexCount)
 
 	for _, borrower := range borrowers {
-		_, err := k.AttemptKeeperLiquidation(ctx, sdk.AccAddress(types.LiquidatorAccount), borrower)
+		err := k.AttemptKeeperLiquidation(ctx, sdk.AccAddress(types.LiquidatorAccount), borrower)
 		if err != nil {
 			if !errors.Is(err, types.ErrBorrowNotLiquidatable) {
-				panic(err) // TODO: should this panic?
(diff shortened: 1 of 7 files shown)


x/hard/keeper/liquidation.go
```diff
@@ -22,7 +22,7 @@ func (k Keeper) AttemptIndexLiquidations(ctx sdk.Context) error {
 	borrowers := k.GetLtvIndexSlice(ctx, params.CheckLtvIndexCount)
 
 	for _, borrower := range borrowers {
-		_, err := k.AttemptKeeperLiquidation(ctx, sdk.AccAddress(types.LiquidatorAccount), borrower)
+		err := k.AttemptKeeperLiquidation(ctx, sdk.AccAddress(types.LiquidatorAccount), borrower)
 		if err != nil {
 			if !errors.Is(err, types.ErrBorrowNotLiquidatable) {
```
karzak, line 27:
```
!errors.Is(err, types.ErrBorrowNotLiquidatable) && !errors.Is(err, types.ErrBorrowNotFound) {
```

catch case where borrower doesn't exist (ie, that user has only supplied coins)

karzak (changes requested):
Changes look good. Added a fix to the issue you raised about index liquidations panicing if the user supplies but doesn't borrow.

x/hard/keeper/liquidation.go
denalimarsh: Fixed in https://github.com/Kava-Labs/kava/pull/758/commits/52d86dfca41c429246af6dc2356a96101f936e9f

Issue #727, opened 2020-12-03
Catch integer overflow panic in sdk.RelativePow


denalimarsh:
`CalculateInterestFactor` in x/hard/interest.go uses `sdk.RelativePow` which panics on integer overflow. We need to create a `SafeRelativePow` function that wraps the current one that acts as a panic handler and reverts if an overflow occurs.

From https://github.com/Kava-Labs/kava/pull/720#discussion_r534462134.

denalimarsh:
Related: https://github.com/Kava-Labs/kava/pull/737#issuecomment-745426776

PR #762, opened 2021-01-08
Hard: investigate sdk.RelativePow runtime panic 

denalimarsh:
### Goal
Determine if `sdk.RelativePow` in the Hard module can panic during runtime. This PR addresses issue https://github.com/Kava-Labs/kava/issues/727. This PR does **not** address the incorrect values returned by `sdk.RelativePow` at the upper-bound as described [here](https://github.com/Kava-Labs/kava/pull/737#issuecomment-745426776).

### Findings
Due to proper param validation, the use of `sdk.RelativePow` in the Hard module cannot panic during runtime. 

### Investigation
The CalculateBorrowInterestFactor function calls `sdk.RelativePow`, which can panic due to integer overflow. CalculateBorrowInterestFactor is only called by AccrueInterest, which in turn is only called by the begin blocker. Assuming a max block time is 30 seconds, `sdk.RelativePow` begins to overflow at a per second interest rate of ~1845%/second (see test in changed files).

The per second interest rate argument to CalculateBorrowInterestFactor is output by the APYToSPY function. The APYToSPY function takes `MoneyMarket.InterestRateModel.BaseRateAPY` as a parameter and panics itself if the `BaseRateAPY` argument is above 177. Inputting a `BaseRateAPY` of 177 to the APYToSPY function is valid and returns 1.000002441641340532, well within the valid range of CalculateBorrowInterestFactor. So the APYToSPY function is our new constraint.

The InterestRateModel param's Validate function limits the max `BaseRateAPY` to 1.0, so APYToSPY will never panic or output a result that causes CalculateBorrowInterestFactor's `sdk.RelativePow` to panic during runtime.

### DIFF
--- a/x/hard/keeper/interest_test.go
+++ b/x/hard/keeper/interest_test.go
@@ -289,14 +289,20 @@ func (suite *InterestTestSuite) TestCalculateBorrowInterestFactor() {
 				expectedValue:         sdk.MustNewDecFromStr("40628388.864535408465693310"),
 			},
 		},
-		// If we raise the per second interest rate too much we'll cause an integer overflow.
-		// For example, perSecondInterestRate: '1.000005555555' will cause a panic.
 		{
 			"1 year: highest interest rate",
 			args{
 				perSecondInterestRate: sdk.MustNewDecFromStr("1.000001555555"),
 				timeElapsed:           sdk.NewInt(oneYearInSeconds),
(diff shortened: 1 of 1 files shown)


rhuairahrighairidh (approved):
Sounds good, looks like it won't panic.

Those numbers that cause panics are really big, if they get that big something else has probably gone wrong.

Is there a doc for how the interest rate is calculated?

denalimarsh:
@rhuairahrighairidh I'll forward you the Interest Rate Data Model document.

PR #764, opened 2021-01-13
Remove savings rate

karzak:


rhuairahrighairidh (approved):
looks good

nit: remove the `savings` module account / permission. There's like $100 in it on mainnet, should be safe to throw into the liquidator account.

PR #765, opened 2021-01-14
Add usdx incentives calculation test

rhuairahrighairidh:
This PR adds an `incentive` test to compares cdp rewards calculated using previous style (updated every block) to the current style (using rewardFactors).

The difference is noticable if the interest rate on the cdp is high enough. The test uses 500% APR, and 10 super long blocks (in total covering 1/5th of a year to get to large accumulated fees).

The second commit adds the expected calculation fix and the numbers get much closer, but not exact. I expect it's due to rounding somewhere but haven't tracked it down.

Optional TODOs
- why did the exising tests still pass when the calculation was changed?
- use `Suite` in the new test to keep style the same

### DIFF
--- a/x/incentive/keeper/rewards_test.go
+++ b/x/incentive/keeper/rewards_test.go
@@ -230,7 +230,6 @@ func TestRewardCalculation(t *testing.T) {
 		NewCDPGenStateHighInterest(),
 		NewIncentiveGenState(initialTime, initialTime.Add(oneYear), rewardPeriod),
 	)
-	fmt.Println("RUN INIT GENESIS")
 
 	// Create a CDP
 	cdpKeeper := tApp.GetCDPKeeper()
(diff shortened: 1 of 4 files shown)


x/incentive/keeper/rewards_test.go
```diff
@@ -200,6 +203,102 @@ func (suite *KeeperTestSuite) TestSyncRewards() {
...
+	_, addrs := app.GeneratePrivKeyAddressPairs(5)
+	tApp := app.NewTestApp()
+	ctx := tApp.NewContext(true, abci.Header{Height: 1, Time: initialTime})
+	tApp.InitializeFromGenesisStates(
+		app.NewAuthGenState(addrs[:1], []sdk.Coins{cs(initialCollateral)}),
+		NewPricefeedGenStateMulti(),
+		NewCDPGenStateHighInterest(),
+		NewIncentiveGenState(initialTime, initialTime.Add(oneYear), rewardPeriod),
+	)
+	fmt.Println("RUN INIT GENESIS")
```
karzak, line 233:
```suggestion

```

karzak (approved):
LGTM

PR #766, opened 2021-01-15
Hard: delete borrow/deposit from store on 0 amount

denalimarsh:
This bug was introduced by the LTV refactor. The tests were comparing empty coins (nil) to empty borrow amount (nil) and passing, which is why they didn't catch it. Updated the tests to be more explicit.

- Should the user be removed from the LTV index if their borrow is deleted from the store? 
- Should the user be removed from the LTV index if their deposit is deleted from the store?

### DIFF
--- a/x/hard/keeper/liquidation.go
+++ b/x/hard/keeper/liquidation.go
@@ -270,9 +270,9 @@ func (k Keeper) UpdateBorrowAndLtvIndex(ctx sdk.Context, borrow types.Borrow, ne
 	k.RemoveFromLtvIndex(ctx, oldLtv, borrow.Borrower)
 	if borrow.Amount.Empty() {
 		k.DeleteBorrow(ctx, borrow)
-	} else {
-		k.SetBorrow(ctx, borrow)
+		return
 	}
+	k.SetBorrow(ctx, borrow)
 	k.InsertIntoLtvIndex(ctx, newLtv, borrow.Borrower)
 }
 
@@ -281,9 +281,9 @@ func (k Keeper) UpdateDepositAndLtvIndex(ctx sdk.Context, deposit types.Deposit,
 	k.RemoveFromLtvIndex(ctx, oldLtv, deposit.Depositor)
 	if deposit.Amount.Empty() {
 		k.DeleteDeposit(ctx, deposit)
-	} else {
-		k.SetDeposit(ctx, deposit)
+		return
 	}
+	k.SetDeposit(ctx, deposit)
 	k.InsertIntoLtvIndex(ctx, newLtv, deposit.Depositor)
 }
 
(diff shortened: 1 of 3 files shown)


karzak:
>Should the user be removed from the LTV index if their borrow is deleted from the store?
Should the user be removed from the LTV index if their deposit is deleted from the store?

Yes and yes

x/hard/keeper/liquidation.go
```diff
@@ -268,14 +268,22 @@ func (k Keeper) IsWithinValidLtvRange(ctx sdk.Context, deposit types.Deposit, bo
 // UpdateBorrowAndLtvIndex updates a borrow and its LTV index value in the store
 func (k Keeper) UpdateBorrowAndLtvIndex(ctx sdk.Context, borrow types.Borrow, newLtv, oldLtv sdk.Dec) {
 	k.RemoveFromLtvIndex(ctx, oldLtv, borrow.Borrower)
-	k.SetBorrow(ctx, borrow)
+	if borrow.Amount.Empty() {
+		k.DeleteBorrow(ctx, borrow)
```
karzak, line 272: return here so the LTV index is removed as well

x/hard/keeper/liquidation.go
```diff
@@ -268,14 +268,22 @@ func (k Keeper) IsWithinValidLtvRange(ctx sdk.Context, deposit types.Deposit, bo
...
+	}
 	k.InsertIntoLtvIndex(ctx, newLtv, borrow.Borrower)
 }
 
 // UpdateDepositAndLtvIndex updates a deposit and its LTV index value in the store
 func (k Keeper) UpdateDepositAndLtvIndex(ctx sdk.Context, deposit types.Deposit, newLtv, oldLtv sdk.Dec) {
 	k.RemoveFromLtvIndex(ctx, oldLtv, deposit.Depositor)
-	k.SetDeposit(ctx, deposit)
+	if deposit.Amount.Empty() {
+		k.DeleteDeposit(ctx, deposit)
```
karzak, line 283: ditto

karzak (changes requested):
Good to merge after LTV index change

x/hard/keeper/liquidation.go
denalimarsh: Updated in https://github.com/Kava-Labs/kava/pull/766/commits/317c587b45f2800c6888970af2bd56d589a009f1

x/hard/keeper/liquidation.go
denalimarsh: Updated in https://github.com/Kava-Labs/kava/pull/766/commits/317c587b45f2800c6888970af2bd56d589a009f1

PR #768, opened 2021-01-18
add query for all current prices

karzak:


denalimarsh (approved):
LGTM. Any other queries simple queries like this we can add to speed up the front-end's data fetching?

PR #769, opened 2021-01-18
[R4R] CDP module migrations from v0.11 -> v0.13

karzak:


rhuairahrighairidh (approved):
Changes in this PR look good so I've approved, but I caught a bug in `MigrateCDP`:

- `migrate.go` ln 49
  Bit of a gotcha but looping over a map returns results in random order. So the totalPrincipals in the migrated genesis will be not be the same for everyone running the migration. So genesis hashes won't be the same. The json sorting done later on won't catch it either, since it's a list. https://play.golang.org/p/JpQwrCt_lb-
We could add a determinism test (once all the migrations are done) to migrate several times and compare hashes.
- `migrate.go` ln 36
  question: in `previousAccumulationTime is updated, but the result is never used. Is it meant to be used?

rhuairahrighairidh (approved):
determinism fix looks good

PR #774, opened 2021-01-21
Minor docs updates

rhuairahrighairidh:
changes
- Fix the broken link to the kava-3 migration guide by removing it. The upgrade process shouldn't really be run by anyone after the upgrade. So keeping a migration guide around permanently could cause confusion. Open to other opinions though
- Update the README and the docs homepage to use latest v0.12.1 versions.


denalimarsh (approved):
Changes LGTM

PR #778, opened 2021-01-25
[R4R] fix: export harvest genesis state 

karzak:


denalimarsh (approved):
utACK. Good catch. Once the deposits/claims are exported we'll import the Deposits to the Hard module and Claims to the Incentive module, but it looks like that part of the migration is going to be in another PR.

PR #784, opened 2021-01-29
[R4R] add v0.13 legacy types for kava-5

karzak:
Adds legacy types for the kava modules (other than hard and incentive, which are addressed in separate PRs as part of migrations).


denalimarsh (approved):
LGTM

PR #786, opened 2021-02-01
[R4R] fix: match reward payouts to current harvest v1 payouts

karzak:
Pays out rewards (USDX or hard) on either the 1st or 15th of the month. Note: this is a back port of harvest v1 so that hard rewards claiming on kava-5 matches kava-4 harvest reward claiming.


x/incentive/keeper/payout.go
```diff
@@ -167,6 +184,28 @@ func (k Keeper) SendTimeLockedCoinsToBaseAccount(ctx sdk.Context, senderModule s
...
 
+// GetPeriodLength returns the length of the period based on the input blocktime and multiplier
+// note that pay dates are always the 1st or 15th of the month at 14:00UTC.
+func (k Keeper) GetPeriodLength(ctx sdk.Context, multiplier types.Multiplier) (int64, error) {
+
+	if multiplier.MonthsLockup == 0 {
+		return 0, nil
+	}
+	switch multiplier.Name {
+	case types.Small, types.Medium, types.Large:
```
denalimarsh, line 195: We can remove this switch statement since all cases execute the same logic and there's no default case specified.

denalimarsh (changes requested):
utACK. One small change then should be good to merge.

x/incentive/keeper/payout.go
karzak: added a default 1d3a2e9

PR #794, opened 2021-02-04
Hard Audit: fix duplicate supply/borrow index factors

denalimarsh:
- [x] Reset SupplyIndexFactor/BorrowIndexFactor if the underlying denom amount is 0 due to a withdraw/repay
- [x] Instead of appending new IndexFactors, update the values of existing IndexFactors

### DIFF
--- a/x/hard/keeper/repay.go
+++ b/x/hard/keeper/repay.go
@@ -43,6 +43,13 @@ func (k Keeper) Repay(ctx sdk.Context, sender sdk.AccAddress, coins sdk.Coins) e
 		return err
 	}
 
+	// If any coin denoms have been completely repaid reset the denom's borrow index factor
+	for _, coin := range payment {
+		if coin.Amount.Equal(borrow.Amount.AmountOf(coin.Denom)) {
+			borrow.Index = borrow.Index.SetInterestFactor(coin.Denom, sdk.ZeroDec())
+		}
+	}
+
 	// Update user's borrow in store
 	borrow.Amount = borrow.Amount.Sub(payment)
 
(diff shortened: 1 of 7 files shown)


x/hard/keeper/repay.go
```diff
@@ -43,6 +43,13 @@ func (k Keeper) Repay(ctx sdk.Context, sender sdk.AccAddress, coins sdk.Coins) e
 		return err
 	}
 
+	// If any coin denoms have been completely repaid reset the denom's borrow index factor
+	for _, coin := range payment {
+		if coin.Amount.Equal(borrow.Amount.AmountOf(coin.Denom)) {
+			borrow.Index = borrow.Index.SetInterestFactor(coin.Denom, sdk.ZeroDec())
+		}
```
rhuairahrighairidh, line 50: nit: I feel like it's slightly safer to remove the interest factor rather than set to zero.
If any other part of the code assumes the values in `borrow.Index` are correct it could run into issues since zero is not a valid factor value. But open to other opions.

rhuairahrighairidh (commented):
looks good

nit: the GetSet methods could also improve the readability of `SyncBorrow/SupplyIndex()`.

rhuairahrighairidh (approved):

x/hard/keeper/repay.go
karzak: Agree that removing the factor is best option. I don't think it has any security implications, but makes the state machine a little more explicit (no open borrow = no factor, as opposed to no open borrow = no factor or zero factor).

denalimarsh: Updated in https://github.com/Kava-Labs/kava/pull/794/commits/6ece947e2f0885f45b9235402b3644358cb4c6b4

PR #796, opened 2021-02-04
Remove unused Hard module accounts

denalimarsh:
The LPAccount and DelegatorAccount module accounts were still in the Hard module even though we've deprecated them, this PR removes them.

### DIFF
--- a/x/hard/legacy/v0_11/types.go
+++ b/x/hard/legacy/v0_11/types.go
@@ -18,12 +18,6 @@ const (
 	// ModuleName name that will be used throughout the module
 	ModuleName = "harvest"
 
-	// LPAccount LP distribution module account
-	LPAccount = "harvest_lp_distribution"
-
-	// DelegatorAccount delegator distribution module account
-	DelegatorAccount = "harvest_delegator_distribution"
-
 	// ModuleAccountName name of module account used to hold deposits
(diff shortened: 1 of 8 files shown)


x/hard/legacy/v0_11/types.go
```diff
@@ -18,12 +18,6 @@ const (
 	// ModuleName name that will be used throughout the module
 	ModuleName = "harvest"
 
-	// LPAccount LP distribution module account
-	LPAccount = "harvest_lp_distribution"
-
-	// DelegatorAccount delegator distribution module account
-	DelegatorAccount = "harvest_delegator_distribution"
-
```
rhuairahrighairidh, line 26: these shouldn't be removed. Anything in `legacy/` for old versions is meant to be a copy of the types from the old version, so shouldn't be modified

rhuairahrighairidh (changes requested):
minor change requested, otherwise looks good

x/hard/legacy/v0_11/types.go
denalimarsh: Addressed in https://github.com/Kava-Labs/kava/pull/796/commits/ce1d67ed48570c482c85f4ec2695f4c8d96607af

rhuairahrighairidh (approved):
looks good

PR #800, opened 2021-02-05
Hard Audit: remove LTV index from Hard module

denalimarsh:
- remove LTV index
- remove LTVIndexCount param


x/hard/keeper/borrow.go
```diff
@@ -22,12 +22,6 @@ func (k Keeper) Borrow(ctx sdk.Context, borrower sdk.AccAddress, coins sdk.Coins
...
 
-	// Get current stored LTV based on stored borrows/deposits
-	prevLtv, err := k.GetStoreLTV(ctx, borrower)
-	if err != nil {
-		return err
-	}
-
 	// Call incentive hook
 	existingBorrow, hasExistingBorrow := k.GetBorrow(ctx, borrower)
 	if hasExistingBorrow {
```
karzak, line 27: Given that we are updating the deposit interest, I think we should call `BeforeDepositModified` as well

x/hard/keeper/withdraw.go
```diff
@@ -9,12 +9,6 @@ import (
...
 func (k Keeper) Withdraw(ctx sdk.Context, depositor sdk.AccAddress, coins sdk.Coins) error {
-	// Get current stored LTV based on stored borrows/deposits
-	prevLtv, err := k.GetStoreLTV(ctx, depositor)
-	if err != nil {
-		return err
-	}
-
 	deposit, found := k.GetDeposit(ctx, depositor)
 	if !found {
 		return sdkerrors.Wrapf(types.ErrDepositNotFound, "no deposit found for %s", depositor)
```
karzak, line 14: Ditto. Since we call `k.SyncBorrowInterest(ctx, depositor)`, we should call `k.BeforeBorrowModified(ctx, borrow)` if one exists.

karzak (approved):
Only changes are unrelated to PR - can be done in a follow up PR for clarity. Otherwise, LGTM.

x/hard/keeper/withdraw.go
denalimarsh: Fixed in https://github.com/Kava-Labs/kava/pull/800/commits/0d9bcf911e3cf6f625c7e72b40023d811679b401

x/hard/keeper/borrow.go
denalimarsh: Fixed in https://github.com/Kava-Labs/kava/pull/800/commits/0d9bcf911e3cf6f625c7e72b40023d811679b401

PR #801, opened 2021-02-05
Hard Audit: enable users to repay another account's borrows

denalimarsh:
Enables users to repay someone else's borrow position.

Tagging @downing034 as a reviewer since this impacts the REST API (additional required argument `owner` for MsgRepay).


x/hard/client/cli/tx.go
```diff
@@ -119,12 +119,12 @@ func getCmdBorrow(cdc *codec.Codec) *cobra.Command {
 
 func getCmdRepay(cdc *codec.Codec) *cobra.Command {
 	return &cobra.Command{
-		Use:   "repay [1000000000ukava]",
+		Use:   "repay [amount] [owner]",
```
karzak, line 122: Might be nicer to have owner be an optional flag, so the default syntax is 

```
kvcli tx hard repay 100000000ukava --from <key>
```

And to pay a different address' loan

```
kvcli tx hard repay 100000000ukava --owner kava1hgcfsuwc889wtdmt8pjy7qffua9dd2tralu64j --from <key>
```

In the default case, `from` becomes `owner`.

x/hard/client/rest/tx.go
```diff
@@ -97,7 +97,7 @@ func postRepayHandlerFn(cliCtx context.CLIContext) http.HandlerFunc {
 			return
 		}
 
-		msg := types.NewMsgRepay(req.From, req.Amount)
+		msg := types.NewMsgRepay(req.From, req.Owner, req.Amount)
```
karzak, line 100: similar to CLI suggestion, could parse the request and if `Owner` is empty, use `From`.

karzak (changes requested):
Minor client suggestions, otherwise LGTM

x/hard/client/cli/tx.go
denalimarsh: Refactored to optional `owner` flag in https://github.com/Kava-Labs/kava/pull/801/commits/a5b1ab86e6250b010c6abb256649b549e041fe6a

x/hard/client/rest/tx.go
denalimarsh: Updated REST in https://github.com/Kava-Labs/kava/pull/801/commits/4d595c50d47153e9a2a2a4be72475a32f63884d1

denalimarsh:
Looks like master build is failing. This branch was passing CircleCI until master latest was merged in.

PR #807, opened 2021-02-08
fix: set reward indexes in init genesis

karzak:
Updates `InitGenesis` and `ExportGenesis` methods so that all reward indexes are set to zero on `InitGenesis` and reset to zero on `ExportGenesis`. 

Fixes the issue that rewards don't accumulate for the first deposit.


denalimarsh (approved):
tACK, fixes the problem. Thanks!

PR #811, opened 2021-02-09
fix: export synced deposits/borrows

karzak:
Synchronizes each deposit/borrow at `ExportGenesis` so that the balance is correct and the object's indexes match the global indexes. 

Synchronizes CDP interest at export as well.


denalimarsh (approved):
LGTM

PR #812, opened 2021-02-10
fix: renumber error messages

karzak:
Removes use of error code `1`, which causes the rawlog for failed transactions to read simply 'internal' rather than a descriptive error message


denalimarsh (approved):
Good catch. Looks like all the other modules already follow this standard.

PR #813, opened 2021-02-10
fix: collateral param permission

karzak:
FIxes issue in `Allows` method for `cdp.CollateralParams` which was comparing `Denom` instead of `Type`. Denom is no longer guaranteed to be unique (`busd-a` and `busd-b`) and this caused the `Allows` method to fail for collateral param changes.


rhuairahrighairidh (approved):
looks good

PR #819, opened 2021-02-11
Add bep3 migration

karzak:


denalimarsh:
Resetting chain height to 0 on the upgrade could introduce issues.

1. Open swaps have an `ExpireHeight` greater than the current block height. When the block height reaches the expiration height the begin blocker marks the swap as expired so it can't be claimed. If there are any open swaps when we migrate they will be claimable until the new chain's height reaches the swap's expected expiration height on the old chain, allowing potential double spends (refund on Binance Chain, then claim on Kava). Last time we paused all BEP3 swaps before migration, we should do the same for this upgrade.
2. Swaps are removed from long-term storage at block height `swap.ClosedBlock + 86400`. Any swaps in long-term storage at the migration won't get deleted for a very long time. Not an attack vector, but would bloat state.

Both issues could be solved by starting kava-5 at kava-4's current block height.

rhuairahrighairidh:
Feature Request:

Set `MinBlockLock` to 24475 and `MaxBlockLock` to 85665 for all `AssetParam`s

This corresponds to 48hrs and 7 days at 7.06s a block (kava-3's average block time)

rhuairahrighairidh (changes requested):
left a suggesting for handling open swaps, I think we can get away with not touching the supplies. If we do update the supplies, we'd also need to return coins to the user

rhuairahrighairidh (approved):
looks good

PR #820, opened 2021-02-11
[R4R] Hard Audit: Refresh borrow before repay

karzak:
Current implementation was fetching `borrow` from the store before syncing interest when a repayment was made. Additionally, the error value for the method `DecrementBorrowedCoins` wasn't being checked. It turns out, it was sometimes throwing an error because it didn't account for the fact that `payment` can be greater than `borrowedCoins`. Because interest accumulates every 7 seconds, while positions are only synced infrequently (when users interact with them), a position can accumulate a few micro-units more borrow interest. When that position is the only position,`DecrementBorrowedCoins` will error.

### DIFF
--- a/x/hard/keeper/repay.go
+++ b/x/hard/keeper/repay.go
@@ -9,16 +9,19 @@ import (
 
 // Repay borrowed funds
 func (k Keeper) Repay(ctx sdk.Context, sender, owner sdk.AccAddress, coins sdk.Coins) error {
-	// Sync borrow interest so loan is up-to-date
-	k.SyncBorrowInterest(ctx, owner)
 
-	// Check borrow exists here to avoid duplicating store read in ValidateRepay
 	borrow, found := k.GetBorrow(ctx, owner)
 	if !found {
 		return types.ErrBorrowNotFound
 	}
 	// Call incentive hook
 	k.BeforeBorrowModified(ctx, borrow)
+
+	// Sync borrow interest so loan is up-to-date
+	k.SyncBorrowInterest(ctx, owner)
+
+	// refresh borrow after syncing interest
+	borrow, _ = k.GetBorrow(ctx, owner)
 
 	// Validate that sender holds coins for repayment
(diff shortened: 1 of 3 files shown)


x/hard/keeper/repay.go
```diff
@@ -17,9 +20,6 @@ func (k Keeper) Repay(ctx sdk.Context, sender, owner sdk.AccAddress, coins sdk.C
 	// Call incentive hook
 	k.BeforeBorrowModified(ctx, borrow)
```
rhuairahrighairidh, line 21: this hook should be before SyncBorrowInterest, since modifications can be made

rhuairahrighairidh (changes requested):
Looks like the hook should called before any modifications to the borrow are made. Though I'm not aware of the implications of this in `incentive`.

The same issue also exists in `keeper/withdraw.go`. `Withdraw` reads in the deposit before syncing it, thereby using the old state in the calculations. And the error is not checked on `DecrementSuppliedCoins`.

x/hard/keeper/repay.go
denalimarsh, line 21: We intentionally changed the ordering in https://github.com/Kava-Labs/kava/pull/800#discussion_r571646085 and https://github.com/Kava-Labs/kava/pull/804 so that `BeforeBorrowModified` is called before `SyncBorrowInterest`. Bit of a chicken and egg problem here. Does modifying the borrow before calling the `BeforeBorrowModified` incentive hook cause any issues? We'll likely want to follow a similar pattern in borrow.go, repay.go, deposit.go, and withdraw.go.

denalimarsh (changes requested):
The approach looks good. Have a question about Incentive hook order.

x/hard/keeper/repay.go
karzak: This came up in CDP accumulator audit - it's correct to call the hook before interest is updated.

karzak:
closing in favor of #824

PR #821, opened 2021-02-12
fix: account for start time in calculating time elapsed

karzak:
Incentive rewards wasn't looking at reward period start time, so it would accumulate rewards even if they were not supposed to have started.


denalimarsh (commented):
LGTM

PR #822, opened 2021-02-12
Hard Audit: add minimum borrow USD value

denalimarsh:
Adds a minimum borrow USD value, default is $10.

Design decisions:
- `MinimumBorrowUSDValue` is a module-level param, it's not inside `MoneyMarkets`. Users can borrow/repay multiple coins per tx so instead of checking individual asset values we need to calculate the proposed resulting borrow's total USD value and compare it to an aggregate minimum USD limit. It also simplifies param management, as there's just one USD denominated value instead of maintaining several limits denominated in different currencies.

Optimizations:
- `ValidateBorrow` and `ValidateRepay` use a caching strategy to reduce param reads as they add computational overhead. Since `MoneyMarkets` are available in the store, we should just read them directly from there and not fetch `MoneyMarkets` from params at all. If we're in agreement about this change let's add an issue/card for it.


karzak (approved):
tACK 

Error message nice and clear 👍

PR #828, opened 2021-02-15
Hard Audit: skipped error check after borrow validation

denalimarsh:
Check error returned from `ValidateBorrow` for general errors as well after checking for module balance errors. The correct approach was already being done in `ValidateDeposit`, in this PR borrow is updated to match it.


rhuairahrighairidh (approved):
looks good

karzak (approved):

PR #829, opened 2021-02-15
fix: set interest last updated to previous accrual time, not block time

karzak:
Fixes import-export test failure.

Because `ctx.BlockTime` is not set during `ExportGenesis`, fees updated time for cdps was getting over-written. Fixes by setting it to the previous accrual time, which is the correct strategy anyway (ctx.BlockTime will usually equal previousAccrualTime, but not during export).


rhuairahrighairidh (approved):
looks good

PR #830, opened 2021-02-16
fix: allow ltv of zero

karzak:
Money markets can be "supply only" - ie supplying the asset is allowed but it doesn't increase your borrow power. This PR fixes current param validation to allow zero LTV.


denalimarsh (approved):
LGTM

rhuairahrighairidh (approved):
looks good

PR #833, opened 2021-02-17
Allocate Hard supply/borrow rewards to legacy suppliers/borrowers

denalimarsh:
When a user deposits or borrows an asset that doesn't have rewards, they should start earning rewards if they're added for the asset.

**General logic**
1. On first deposit/borrow, if the deposited/borrowed denom doesn't have rewards, still add the denom to the user's claim object but initialize it with empty reward indexes
2. Do the same on successive borrows/deposits
3. On sync, if the denom now has global rewards, sync the user's reward index from 0.0 to the current global reward index

### DIFF
--- a/x/incentive/keeper/rewards.go
+++ b/x/incentive/keeper/rewards.go
@@ -439,11 +439,13 @@ func (k Keeper) UpdateHardSupplyIndexDenoms(ctx sdk.Context, deposit hardtypes.D
 	for _, coin := range deposit.Amount {
 		_, foundUserRewardIndexes := claim.SupplyRewardIndexes.GetRewardIndex(coin.Denom)
 		if !foundUserRewardIndexes {
-			globalRewardIndexes, foundGlobalRewardIndexes := k.GetHardSupplyRewardIndexes(ctx, coin.Denom)
-			if !foundGlobalRewardIndexes {
-				continue // No rewards for this coin type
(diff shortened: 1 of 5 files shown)


x/incentive/keeper/rewards.go
```diff
@@ -373,8 +382,8 @@ func (k Keeper) SynchronizeHardBorrowReward(ctx sdk.Context, borrow hardtypes.Bo
 			continue
 		}
 
-		userRewardIndexes, foundUserRewardIndexes := claim.BorrowRewardIndexes.GetRewardIndex(coin.Denom)
-		if !foundUserRewardIndexes {
+		userMultiRewardIndex, foundUserMultiRewardIndex := claim.SupplyRewardIndexes.GetRewardIndex(coin.Denom)
```
karzak, line 385:
```suggestion
		userMultiRewardIndex, foundUserMultiRewardIndex := claim.BorrowRewardIndexes.GetRewardIndex(coin.Denom)
```

karzak (changes requested):
Fix Supply/Borrow switch, otherwise looks good. 

Tested with kvtool, only other issue I found I just turned into a PR (#834)

x/incentive/keeper/rewards.go
denalimarsh: Fixed in https://github.com/Kava-Labs/kava/pull/833/commits/8d16c3e8c3f3efdd601687310fa9e952306921ac

karzak (approved):

PR #834, opened 2021-02-17
Update querier to show synced rewards for legacy deposits/borrows

karzak:


denalimarsh (approved):
Ah forgot about the simulation methods, thanks for updating

PR #835, opened 2021-02-18
[R4R] fix: set indexes to zero

karzak:
Reward indexes weren't actually getting set to zero on export


denalimarsh (approved):
Ah, I should have caught this before. LGTM 👍

rhuairahrighairidh (approved):
looks good

PR #842, opened 2021-02-20
Hard Audit: don't convert directly from sdk.Int to uint64

denalimarsh:
Updates Hard and CDP modules' interest calculations to use big.Int as an intermediary data type when converting from sdk.Int to sdk.Uint.


karzak (approved):
Nice, didn't even have to implement a new method.

PR #843, opened 2021-02-20
Query Hard module reserves

denalimarsh:
Implements CLI and REST querier for Hard module's current reserves. The query is filterable by `denom` and always returns the `sdk.Coins` data type.


rhuairahrighairidh (approved):

karzak (approved):

PR #849, opened 2021-02-23
fix: enforce max number of liquidations in cdp begin blocker

karzak:
Current implementation would liquiate all cdps at a given ratio - which could result in more cdps being liquidated than the params suggest. Fixed by grabbing a slice of cdps that is limited by `collateral_params.check_collateralization_index_count`

### DIFF
--- a/x/cdp/keeper/seize.go
+++ b/x/cdp/keeper/seize.go
@@ -87,7 +87,7 @@ func (k Keeper) SeizeCollateral(ctx sdk.Context, cdp types.CDP) error {
 }
 
 // LiquidateCdps seizes collateral from all CDPs below the input liquidation ratio
-func (k Keeper) LiquidateCdps(ctx sdk.Context, marketID string, collateralType string, liquidationRatio sdk.Dec, slice sdk.Int) error {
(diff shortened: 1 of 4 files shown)


x/cdp/keeper/seize.go
```diff
@@ -87,7 +87,7 @@ func (k Keeper) SeizeCollateral(ctx sdk.Context, cdp types.CDP) error {
 }
 
 // LiquidateCdps seizes collateral from all CDPs below the input liquidation ratio
-func (k Keeper) LiquidateCdps(ctx sdk.Context, marketID string, collateralType string, liquidationRatio sdk.Dec) error {
+func (k Keeper) LiquidateCdps(ctx sdk.Context, marketID string, collateralType string, liquidationRatio sdk.Dec, slice sdk.Int) error {
```
denalimarsh, line 90:
```suggestion
func (k Keeper) LiquidateCdps(ctx sdk.Context, marketID string, collateralType string, liquidationRatio sdk.Dec, count sdk.Int) error {
```
Nit

denalimarsh (approved):
LGTM

rhuairahrighairidh (approved):
looks good

PR #853, opened 2021-02-24
fix: check correct time elpased condition

karzak:
Time elapsed was checking 

```
(start.After(previousAccrualTime)) 
```

- should be checking against the current block time


rhuairahrighairidh (approved):

denalimarsh (approved):
LGTM

PR #854, opened 2021-02-27
Fix: Add missing stability committee permissions

karzak:
Fixes:
1. removes old `harvest` param permissions
2. adds `MoneyMarkets` and `MinimumBorrowUSDValue` hard parameter keys to allowed stability committee permissions
3. ensures that all collateral types are added to stability committee `AllowedCollateralParams`


rhuairahrighairidh (approved):
haven't tested locally, but looks good

PR #859, opened 2021-03-02
Fix delegator rewards initialization

karzak:
Bug: When a new delegation was created, `InitializeHardDelegatorReward` was overriding the value of `DelegatorRewardIndexes` without first synchronizing the claim object. This resulted in lost HARD delegator rewards when new delegations were created for delegators that already had accumulated rewards.

Fix: Synchronize the claim object first, then update `DelegatorRewardIndexes`


rhuairahrighairidh (approved):
looks good

PR #863, opened 2021-03-04
kava-5 rollback migration

karzak:


rhuairahrighairidh (approved):
looks good
I tested against kava-4 exported state (height 1267330, exported for-zero-height) and diffed the migrated with original and it looks good.

denalimarsh (approved):

PR #865, opened 2021-03-07
fix: decrement keeper liquidation reward coins from hard total supply

karzak:
Decrements keeper reward coins from hard total supply.
Additionally, updates hard liquidation tests so that the initial module balance is deposited rather than minted - this mimics the protocol functionality.


rhuairahrighairidh (approved):
looks good

PR #867, opened 2021-03-08
Fix hard/committee begin blocker order

rhuairahrighairidh:
This PR verified and updated the money market deprecation steps.

As part of this the committee / hard begin blocker order was switched to ensure interest was accrued correctly around param changes. For small blocks this is minor, but long blocks, say due to a chain halt, it could be larger.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -432,8 +432,8 @@ func NewApp(logger log.Logger, db dbm.DB, traceStore io.Writer, appOpts AppOptio
 	// So it should be run before cdp.BeginBlocker which cancels out debt with stable and starts more auctions.
 	app.mm.SetOrderBeginBlockers(
 		upgrade.ModuleName, mint.ModuleName, distr.ModuleName, slashing.ModuleName,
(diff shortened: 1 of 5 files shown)


app/app.go
```diff
@@ -433,7 +433,7 @@ func NewApp(logger log.Logger, db dbm.DB, traceStore io.Writer, appOpts AppOptio
 	app.mm.SetOrderBeginBlockers(
 		upgrade.ModuleName, mint.ModuleName, distr.ModuleName, slashing.ModuleName,
 		validatorvesting.ModuleName, kavadist.ModuleName, auction.ModuleName, cdp.ModuleName,
-		bep3.ModuleName, hard.ModuleName, committee.ModuleName, issuance.ModuleName, incentive.ModuleName,
+		bep3.ModuleName, committee.ModuleName, hard.ModuleName, issuance.ModuleName, incentive.ModuleName,
```
karzak, line 436: Along similar lines - move `committee` in front of `cdp` begin blocker so that interest accumulation reflects any changes from proposals

karzak (changes requested):
Tiny nit - otherwise LGTM

PR #868, opened 2021-03-08
fix: set previous block time correctly on block one

karzak:


rhuairahrighairidh (approved):
looks good, though haven't tested locally

PR #885, opened 2021-03-19
Fix: delete incentive reward factors on full withdraw/repay

denalimarsh:
Fixes the recently reported bug where incentive reward indexes were not being deleted on Hard repay/withdraw.


x/incentive/keeper/rewards.go
```diff
@@ -873,3 +893,27 @@ func (k Keeper) SimulateUSDXMintingSynchronization(ctx sdk.Context, claim types.
 
 	return claim
 }
+
+// Set setDifference: A - B
+func setDifference(a, b []string) (diff []string) {
```
karzak, line 898: Worth adding unit test imo

rhuairahrighairidh (approved):
looks good, only nit is to add the test for `setDifference`

PR #886, opened 2021-03-22
Fix index sync when deposits/borrows are small

rhuairahrighairidh:
To repoduce bug:
- a user deposits 1ukava in hard
- wait a long duration _t_
- the user deposits 10^12ukava
- the user's claim will now show they've been rewarded as if they've had 10^12ukava deposited for time _t_.

Problem: the reward indexes in a user's claim were not being updated when new reward amounts rounded to zero.

Changes
- sync a claim's indexes even when the new rewards are zero
- panic if the new reward amount would be negative. This would be caused by a global index being < a claim's index. Global indexes should never decrease.


karzak (approved):
tACK

PR #889, opened 2021-03-23
Fix delegation claim syncing

rhuairahrighairidh:
We distribute rewards based on a user's total delegated tokens to bonded validators.
Whenever this value changes, we need to synchronize the incentive claim beforehand.
Previously we did not catch all cases where this value changed. This PR catches the rest of them.

The fix is added in the first commit: https://github.com/Kava-Labs/kava/pull/889/commits/d9f6e612a58a9da35f354e54c2e7c5a47f5c99d5
The other commits are tidying up and adding tests.

The fix was basically adding more staking hooks to catch the cases where:
- a validator changes between bonded/unbonding
- when slashing changes the total delegated tokens

We had to add a couple of arguments to `SynchronizeHardDelegatorRewards` to deal with a validator's bond/unbonding state changing right before the hook is called. We ideally want the hook called before the validator state changes, but instead we have to pretend that validator has the opposite state when summing up all the delegations.
For example, when a validator starts unbonding (drops out of the top 100), we sync claims and still include delegations to this validator even though it's now unbonding. This calculates rewards up to the current time.

The last commit https://github.com/Kava-Labs/kava/pull/889/commits/135b8dab9d8f9d9366fb640426ffa3c65cb04d9c isn't strictly necessary keeps the hook calling consistent.
We weren't calling the `AfterDeposit/BorrowModified` hooks when a hard deposit/borrow was liquidated and deleted.
As far as I can tell this caused no issues, but it's different to what happens when a deposit/borrow is deleted during withdraw/repay; where the hook is always called.

Previously after liquidation a claim would still have the old reward indexes. This isn't a problem as the whole deposit/borrow is deleted, so when it's created again (ie when user next depsits) the `AfterDeposit/BorrowCreated` hook is run which wipes out whatever indexes are currently in the claim. Now the reward indexes are just cleared after liquidation.

### DIFF
--- a/x/hard/client/cli/query.go
+++ b/x/hard/client/cli/query.go
@@ -38,11 +38,14 @@ func GetQueryCmd(queryRoute string, cdc *codec.Codec) *cobra.Command {
 		queryParamsCmd(queryRoute, cdc),
 		queryModAccountsCmd(queryRoute, cdc),
 		queryDepositsCmd(queryRoute, cdc),
+		queryUnsyncedDepositsCmd(queryRoute, cdc),
 		queryTotalDepositedCmd(queryRoute, cdc),
 		queryBorrowsCmd(queryRoute, cdc),
+		queryUnsyncedBorrowsCmd(queryRoute, cdc),
 		queryTotalBorrowedCmd(queryRoute, cdc),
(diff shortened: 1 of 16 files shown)


karzak (approved):
tACK

PR #894, opened 2021-03-25
Incentive queries: unsynced claims, global reward factors

denalimarsh:
Adds the following queries to the incentive module:
- Query unsynced claims: `kvcli q incentive rewards --unsynced true`
- Query global reward factors: `kvcli q incentive reward-factors`, `kvcli q incentive reward-factors --denom ukava`


karzak (approved):
tACK

PR #897, opened 2021-04-05
Call incentive hooks before CDPs are modified in begin blocker

rhuairahrighairidh:
In the cdp module, risky cdps are synced regularly to update their collateral ratio so they can be accurately liquidated (if below liquidation ratio).
This modifies the cdp, however the `BeforeCDPModified` hook was not called. This PR adds the hook call and a test.

### DIFF
--- a/x/incentive/keeper/cdp_test.go
+++ b/x/incentive/keeper/cdp_test.go
@@ -1,0 +1,67 @@
+package keeper_test
+
+import (
+	"time"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	abci "github.com/tendermint/tendermint/abci/types"
+
+	cdptypes "github.com/kava-labs/kava/x/cdp/types"
+	"github.com/kava-labs/kava/x/incentive/types"
+)
+
+func (suite *KeeperTestSuite) TestRiskyCDPsAccumulateRewards() {
+	suite.SetupWithGenState()
+	initialTime := suite.ctx.BlockTime()
+
+	// Setup incentive state
+	collateralType := "bnb-a"
+	rewardsPerSecond := c(types.USDXMintingRewardDenom, 1_000_000)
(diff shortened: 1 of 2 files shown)


x/incentive/keeper/cdp_test.go
```diff
@@ -0,0 +1,67 @@
...
+	_ = suite.app.BeginBlocker(suite.ctx, abci.RequestBeginBlock{}) // height and time in header are ignored by module begin blockers
+
+	_ = suite.app.EndBlocker(suite.ctx, abci.RequestEndBlock{})
+	suite.ctx = suite.ctx.WithBlockTime(suite.ctx.BlockTime().Add(10 * time.Minute))
+	_ = suite.app.BeginBlocker(suite.ctx, abci.RequestBeginBlock{})
+
+	// check cdp rewards
+	cdp, found := cdpKeeper.GetCdpByOwnerAndCollateralType(suite.ctx, suite.addrs[0], collateralType)
+	suite.Require().True(found)
+	suite.keeper.SynchronizeUSDXMintingReward(suite.ctx, cdp)
```
karzak, line 61: why does sync need to be called here... shouldn't it have been synced in the begin blocker?

karzak (commented):
Question on the test case, otherwise LGTM

x/incentive/keeper/cdp_test.go
rhuairahrighairidh: I added some comments to try and make this clearer.
The cdp syncs during the cdp begin blocker, but rewards accumulate globally during the incentive begin blocker, which runs after the cdp one. So at this point in the code the incentive BB has accumulated some rewards that haven't yet been added/synced to the claim.

rhuairahrighairidh:
I'd wait on merging this until #908 is merged.

PR #898, opened 2021-04-05
Update migration docs 

karzak:


denalimarsh (approved):
Edits LGTM. I double-checked the dates, halt time, and downgrade version in case of failure.

PR #899, opened 2021-04-19
Committee types

denalimarsh:
Refactor Committee to an interface that supports `MemberCommittee` and `TokenCommittee`


karzak:
The interface approach sounds good. 

Only note is that `VoteThreshold` still needs to be used by both committee types and can be included in the base committee type. This is because a coin voting committee needs both 1) the quorum needed for the tallying to even take place and 2) the `VoteThreshold`, or the percentage of coins that must vote YES for a proposal to pass.

x/committee/types/committee.go
```diff
@@ -13,52 +13,148 @@ import (
...
+		c.ID, c.Description, c.Permissions,
+		c.VoteThreshold.String(), c.ProposalDuration.String(),
+		c.TallyOption,
+	)
+}
+
+//
+type MemberCommittee struct {
+	BaseCommittee `json:"base_committee" yaml:"base_committee"`
+	Members       []sdk.AccAddress `json:"members" yaml:"members"`
```
karzak, line 122: `Members` should be added to the base type, as this defines who is allowed to make proposals. Given that, I think `MemberCommittee` can just be an alias for the base committee type.

karzak (changes requested):

x/committee/types/committee.go
```diff
@@ -73,32 +121,121 @@ func (c Committee) Validate() error {
...
+		Quorum:     quorum,
+		TallyDenom: tallyDenom,
+	}
+}
+
+// GetType returns the type of the committee
+func (c TokenCommittee) GetType() string { return TokenCommitteeType }
+
+// Validate validates the committee's fields
+func (c TokenCommittee) Validate() error {
```
karzak, line 209: This method should return `c.BaseCommittee.Validate()`

x/committee/types/committee.go
```diff
@@ -73,32 +121,121 @@ func (c Committee) Validate() error {
...
+			Permissions:      permissions,
+			VoteThreshold:    threshold,
+			ProposalDuration: duration,
+			TallyOption:      tallyOption,
+		},
+	}
+}
+
+// GetType returns the type of the committee
+func (c MemberCommittee) GetType() string { return MemberCommitteeType }
```
karzak, line 179: Define `Validate` method and have it return `c.BaseCommittee.Validate()`

x/committee/types/committee.go
```diff
@@ -13,32 +13,60 @@ import (
 
 const MaxCommitteeDescriptionLength int = 512
 
-// ------------------------------------------
-//				Committees
-// ------------------------------------------
+type TallyOption int
+
+const (
```
karzak, line 18: Add comment to each of the tally options that describes them.

FPTH - Votes are tallied each block and the proposal passes as soon as the vote threshold is reached
Deadline - Votes are tallied exactly once, when the deadline time is reached

x/committee/types/committee.go
```diff
@@ -13,32 +13,60 @@ import (
...
+	FirstPastThePost TallyOption = iota // 0
+	Deadline         TallyOption = iota // 1
+)
+
+const (
```
karzak, line 23: Add descriptive comments for each committee type

karzak (changes requested):
Left a few nits, should be mergeable after changes

x/committee/types/committee.go
denalimarsh: https://github.com/Kava-Labs/kava/pull/899/commits/5ff4eeb6591dcf2040bb900af67fdfe6b2165c98

x/committee/types/committee.go
denalimarsh: https://github.com/Kava-Labs/kava/pull/899/commits/5ff4eeb6591dcf2040bb900af67fdfe6b2165c98

x/committee/types/committee.go
denalimarsh: https://github.com/Kava-Labs/kava/pull/899/commits/5ff4eeb6591dcf2040bb900af67fdfe6b2165c98

x/committee/types/committee.go
denalimarsh: https://github.com/Kava-Labs/kava/pull/899/commits/5ff4eeb6591dcf2040bb900af67fdfe6b2165c98

karzak (approved):
LGTM

PR #900, opened 2021-04-20
Update MsgVote with vote type

denalimarsh:
Adds `Yes`/`No`/`Abstain` to `MsgVote`

### DIFF
--- a/x/committee/types/msg_test.go
+++ b/x/committee/types/msg_test.go
@@ -56,12 +56,17 @@ func TestMsgVote_ValidateBasic(t *testing.T) {
 	}{
 		{
 			name:       "normal",
-			msg:        MsgVote{5, addr},
+			msg:        MsgVote{5, addr, Yes},
+			expectPass: true,
+		},
+		{
+			name:       "abstain",
+			msg:        MsgVote{5, addr, Abstain},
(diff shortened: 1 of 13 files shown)


x/committee/types/msg_test.go
```diff
@@ -56,12 +56,17 @@ func TestMsgVote_ValidateBasic(t *testing.T) {
 	}{
 		{
 			name:       "normal",
-			msg:        MsgVote{5, addr},
+			msg:        MsgVote{5, addr, Yes},
+			expectPass: true,
+		},
+		{
```
karzak, line 62: Does it make sense to update validate basic to verify that the vote input is valid? If so, could add a test case like

```
{
  name:       "invalid vote",
  msg:        MsgVote{5, addr, 3},
  expectPass: false,
},
```

karzak (changes requested):

x/committee/types/msg_test.go
denalimarsh: https://github.com/Kava-Labs/kava/pull/900/commits/800604b780bb0b4a1bd452d61285ca28505a93d3

karzak (approved):
LGTM

PR #908, opened 2021-05-06
Incentive Test Refactors

rhuairahrighairidh:
Aims
- introduce helper functions to make test setup easier for future refactors
- make existing tests clearer in what they test
- don't change production code

Changes
- Split up `KeeperTestSuite` into several independent `TestSuite`s 
  - https://github.com/Kava-Labs/kava/compare/master..70cbde4
  - I found this made refactoring easier as changes would only break some tests rather than all of them
  - Though it did introduce some duplication in the keeper setup methods

- Use `InitGenesis` to setup state, rather than keeper calls
  - https://github.com/Kava-Labs/kava/compare/70cbde4..11ab19e
  - The tests had a mix of setup via InitGenesis and keeper method callls. This can setup the app in a way that couldn't happen on mainnet, reducing chance of catching bugs.
  - Downside is that you can't always get the exact setup you want, as InitGenesis panics with invalid genesis states. Potential solution is set of fine-grained of tests that do not use testApp. They'd instead just create a Keeper directly and mock other modules.
  - I removed a couple of test cases that didn't work with InitGenesis setup. Left them as todos to add in next PRs

- Add "builder" types to make setting up genesis states less painful.
  - The `AuthGenesisBuilder` seemed helpful in other modules so I moved alongside TestApp.

- Remove `initialTime` from test cases
  - I noticed most test cases had an `initialTime` param that acted as the test app's genesis time. No test really used it so I removed it to simplify them.


denalimarsh (approved):
I agree with the direction here, really like the app-level setup functions. I have some feedback/nits but I'm going to wait until the next iteration of these tests is ready - asking for small revisions on the tests feels premature since the keeper code is actively being refactored and relevant TODOs are already acknowledged.

Since these changes are in development and not quite ready for `master` could we use an `incentive-refactor` development branch as the base branch for this PR?

PR #914, opened 2021-05-22
Committee gov type registration

rhuairahrighairidh:
Gov proposals to create/change/delete committees could not be submitted via the cli, as the new Committee interface type had not been registered on the gov module codec. This PR adds the registration.


denalimarsh (approved):
Good catch +1

PR #915, opened 2021-05-24
[R4R] add community multi-spend proposal

karzak:
Implements a new gov proposal for spending from the community pool and sending funds to multiple addresses at once. 

Did a basic test with kvtool and was able to submit a proposal

Did some gas testing with `kvtool` - this example spends coins to 250 `PeriodicVestingAccounts that have average ~10 vesting periods.

To replicate:

```
# From kvtool directory
git checkout kd-community-spend
make install
kvtool testnet bootstrap --kava.configTemplate master 

# copy proposal to docker container
docker cp big-multi-spend-example.json generated_kavanode_1:/root/kava/multi-spend-example.json

# submit proposal
dkvcli tx gov submit-proposal community-pool-multi-spend multi-spend-example.json --from user1 --gas 10000000

# uses about 40% of block gas
# gasused: 8398012
```


rhuairahrighairidh (commented):
Looks good, but haven't done any testing yet

Is this change to be released as an async upgrade? Currently it will fork if someone submits the new proposal.

One nit is that it'd be nice if the proposal wasn't in `kavadist` as it doesn't share functionality. But I can't think of a better place to put it.

PR #917, opened 2021-05-28
Token holder governance

denalimarsh:
This PR contains token holder governance code changes and audit revisions. Audit revisions are the series of commits from May 26th.


karzak (approved):
Blocked by #921, otherwise LGTM

PR #919, opened 2021-05-28
Emit final proposal tally in close proposal event

denalimarsh:
This PR emits the final proposal tally in the close proposal event.


x/committee/keeper/proposal.go
```diff
@@ -251,13 +251,32 @@ func (k Keeper) enactProposal(ctx sdk.Context, proposal types.Proposal) error {
 // CloseProposal deletes proposals and their votes, emitting an event denoting the final status of the proposal
 func (k Keeper) CloseProposal(ctx sdk.Context, proposal types.Proposal, outcome types.ProposalOutcome) {
 
+	proposalTally := types.ProposalPollingStatus{}
+	committee, found := k.GetCommittee(ctx, proposal.CommitteeID)
+	if found {
+		switch com := committee.(type) {
```
karzak, line 257: Do we want

```
switch committee.GetType()
```

not sure how `committee.(type)` behaves for custom defined types.

karzak (changes requested):
One question, otherwise LGTM

x/committee/keeper/proposal_test.go
```diff
@@ -818,8 +819,28 @@ func (suite *KeeperTestSuite) TestCloseProposal() {
...
+	hasProposalTallyAttr := false
+	for _, attr := range event.Attributes {
+		if string(attr.GetKey()) == "proposal_tally" {
+			hasProposalTallyAttr = true
+			valueStr := string(attr.GetValue())
+			suite.Require().True(strings.Contains(valueStr, "Yes votes:"))
+			suite.Require().True(strings.Contains(valueStr, "Current votes:"))
+			suite.Require().True(strings.Contains(valueStr, "Possible votes:"))
+			suite.Require().True(strings.Contains(valueStr, "Vote threshold:"))
+			suite.Require().True(strings.Contains(valueStr, "Quorum:"))
```
rhuairahrighairidh, line 839: hot tip: there's a `suite.Contains` method that will check if a string contains another string. It'll have nicer error messages and saves some typing.
https://pkg.go.dev/github.com/stretchr/testify@v1.7.0/assert#Contains

x/committee/keeper/proposal.go
rhuairahrighairidh: It should work just the same for custom tyes as for go types

x/committee/keeper/proposal.go
```diff
@@ -251,13 +251,32 @@ func (k Keeper) enactProposal(ctx sdk.Context, proposal types.Proposal) error {
...
+	}
+
 	k.DeleteProposalAndVotes(ctx, proposal.ID)
 
 	ctx.EventManager().EmitEvent(
 		sdk.NewEvent(
 			types.EventTypeProposalClose,
 			sdk.NewAttribute(types.AttributeKeyCommitteeID, fmt.Sprintf("%d", proposal.CommitteeID)),
 			sdk.NewAttribute(types.AttributeKeyProposalID, fmt.Sprintf("%d", proposal.ID)),
+			sdk.NewAttribute(types.AttributeKeyProposalTally, proposalTally.String()),
```
rhuairahrighairidh, line 279: Parsing out the results of a proposal from this attribute will be difficult for clients since we use a custom `String` function.
I think it's reasonably likely that someone will want to parse the results and display them somewhere, so we should probably change this.
However storing the results sounds like a lot of work. So I propose just marshalling `proposalTally` to a json string in the attribute. Then it should be easy enough the fetch the results from the block the proposal closed in. @denalimarsh @karzak thoughts?

I'd use `k.cdc.MarshalJSON`, and if it errors probably just log the error. `CloseProposal` is called in the begin blocker so we don't really want to panic and halt the chain over some string formatting.

denalimarsh: Updated in https://github.com/Kava-Labs/kava/pull/919/commits/afa5818c1a7ef55529d892006f8d2e7bf1e1c648

x/committee/keeper/proposal_test.go
denalimarsh: Updated in https://github.com/Kava-Labs/kava/pull/919/commits/afa5818c1a7ef55529d892006f8d2e7bf1e1c648

karzak (approved):

PR #922, opened 2021-06-02
Swap module scaffolding and params

denalimarsh:
This PR includes the swap module scaffolding and params


x/swap/types/pair.go
```diff
@@ -0,0 +1,94 @@
...
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// Pair defines a tradable token pair
+type Pair struct {
+	TokenA    string  `json:"token_a" yaml:"token_a"`
+	TokenB    string  `json:"token_b" yaml:"token_b"`
+	RewardAPY sdk.Dec `json:"reward_apy" yaml:"reward_apy"`
```
karzak, line 13: Swap module shouldn't have any knowledge of rewards, so this can be removed. This will be defined in `incentive`.

x/swap/types/pair.go
```diff
@@ -0,0 +1,87 @@
...
+			p.TokenA, p.TokenB,
+		)
+	}
+
+	return nil
+}
+
+// Name returns a unique name for a pair in alphabetical order
+func (p Pair) Name() string {
+	if p.TokenA < p.TokenB {
```
nddeluca, line 47: @karzak @denalimarsh @rhuairahrighairidh Should we move this to Validate?  It seems like we could that `TokenA <TokenB` at param creation, and clients can always assume the name will be `:token_a:/:token_b:`.

x/swap/types/pair.go
```diff
@@ -0,0 +1,87 @@
...
+		return fmt.Errorf(
+			"pair cannot have two tokens of the same type, received '%s' and '%s'",
+			p.TokenA, p.TokenB,
+		)
+	}
+
+	return nil
+}
+
+// Name returns a unique name for a pair in alphabetical order
```
karzak, line 45: Not sure if it makes sense to scope it in this PR, but we should define and unit test a `GetName(denomA, denomB string)` method before defining keeper methods. When depositing/swaping, we will get two input denoms and we should have a single method that returns the canonical name given the inputs.

karzak:
> @karzak @denalimarsh @rhuairahrighairidh Should we move this to Validate?  It seems like we could check that `TokenA < TokenB` at pair creation, and clients can always assume the name will be `:token_a:/:token_b:`. 

Agree this should be checked at time of pair creation

karzak (changes requested):

x/swap/keeper/params.go
```diff
@@ -0,0 +1,31 @@
...
+	k.paramSubspace.GetParamSet(ctx, &p)
+	return p
+}
+
+// SetParams sets params on the store
+func (k Keeper) SetParams(ctx sdk.Context, params types.Params) {
+	k.paramSubspace.SetParamSet(ctx, &params)
+}
+
+// GetPairsWithDenom all pairs with the given token denom
```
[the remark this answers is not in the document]
nddeluca: removed, easy to add back in if we need it

x/swap/types/params.go
```diff
@@ -0,0 +1,91 @@
...
+// Parameter keys and default values
+var (
+	KeyPairs       = []byte("Pairs")
+	KeySwapFee     = []byte("SwapFee")
+	DefaultPairs   = Pairs{}
+	DefaultSwapFee = sdk.ZeroDec()
+	MaxSwapFee     = sdk.OneDec()
+)
+
+// Params are governance parameters for the swap module
```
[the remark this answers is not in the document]
nddeluca: Renamed to AllowedPools to make intent clear and not conflict with a Pool data structure in next PR

x/swap/types/pair.go
nddeluca: noted, will plan to introduce in deposit chnages

karzak (approved):
LGTM

rhuairahrighairidh (approved):
looks good

PR #924, opened 2021-06-09
Swap module simulation scaffolding

denalimarsh:
Implements [swap sims ticket](https://app.asana.com/0/1156716699555540/1200386098666785/f)

### DIFF
--- a/x/swap/simulation/genesis.go
+++ b/x/swap/simulation/genesis.go
@@ -1,0 +1,85 @@
+package simulation
+
+import (
+	"fmt"
+	"math/rand"
+	"sort"
+	"strings"
+
+	"github.com/cosmos/cosmos-sdk/codec"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/module"
+	"github.com/cosmos/cosmos-sdk/x/simulation"
+
+	"github.com/kava-labs/kava/x/swap/types"
+)
+
+var (
+	accs []simulation.Account
+)
+
+// GenSwapFee generates a random SwapFee in range [0.01, 1.00]
+func GenSwapFee(r *rand.Rand) sdk.Dec {
+	min := int(1)
+	max := int(100)
+	percentage := int64(r.Intn(int(max)-min) + min)
+	return sdk.NewDec(percentage).Quo(sdk.NewDec(100))
+}
+
+// GenSupportedAssets generates random allowed pools
+func GenAllowedPools(r *rand.Rand) types.AllowedPools {
+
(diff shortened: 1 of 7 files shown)


x/swap/simulation/genesis.go
```diff
@@ -0,0 +1,85 @@
...
+	tokens := []string{genTokenDenom(r), genTokenDenom(r)}
+	sort.Strings(tokens)
+	return tokens[0], tokens[1]
+}
+
+func genTokenDenom(r *rand.Rand) string {
+	denom := strings.ToLower(simulation.RandStringOfLength(r, 3))
+	err := sdk.ValidateDenom(denom)
+	for err != nil {
+		denom = strings.ToLower(simulation.RandStringOfLength(r, 3))
```
karzak, line 55: Can we re-write this to only create valid denoms?

x/swap/simulation/params.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"math/rand"
+
+	"github.com/cosmos/cosmos-sdk/x/simulation"
+)
+
+// ParamChanges defines the parameters that can be modified by param change proposals
+// on the simulation
+func ParamChanges(r *rand.Rand) []simulation.ParamChange {
+	return []simulation.ParamChange{}
```
karzak, line 12: We should define this function to allow pools and swap fee to be changed. https://github.com/cosmos/cosmos-sdk/blob/v0.39.2/x/mint/simulation/params.go#L22 has good example

x/swap/simulation/genesis.go
```diff
@@ -0,0 +1,85 @@
...
+			tokenA, tokenB = genTokenDenoms(r)
+		}
+		newPool := types.NewAllowedPool(tokenA, tokenB)
+		pools[i] = newPool
+	}
+	return pools
+}
+
+func genTokenDenoms(r *rand.Rand) (string, string) {
+	tokens := []string{genTokenDenom(r), genTokenDenom(r)}
```
karzak, line 46: should check denom[0] != denom[1]

karzak (changes requested):

x/swap/simulation/genesis.go
denalimarsh: https://github.com/Kava-Labs/kava/pull/924/commits/8621d97b877d15a420b044b82b183a573c3226ae

x/swap/simulation/params.go
denalimarsh: https://github.com/Kava-Labs/kava/pull/924/commits/71d0eab53138120edc4ebecbd521af0721266184

PR #925, opened 2021-06-09
Swap migrations scaffolding

denalimarsh:
Git history has a lot of commits because I had to merge in master in order to get the latest `migrate/0v_15/migrate.go` migration file in order to update it with the swap module's genesis state. The relevant commits to review are https://github.com/Kava-Labs/kava/commit/3db7f91dafa5b997a33e2d2abd74d6e7f69bda06 and https://github.com/Kava-Labs/kava/commit/9243a0effe024e9a2ba790f9b610824bd488f578.


migrate/v0_15/migrate.go
```diff
@@ -191,3 +192,8 @@ func Committee(genesisState v0_14committee.GenesisState) v0_15committee.GenesisS
 	return v0_15committee.NewGenesisState(
 		genesisState.NextProposalID, committees, proposals, votes)
 }
+
+// Swap migrates new swap module state to a v0.15 swap genesis state
+func Swap(genesisState v0_15swap.GenesisState) v0_15swap.GenesisState {
+	return v0_15swap.NewGenesisState(genesisState.Params)
```
nddeluca, line 198: Should we add the swap module account here as well?

migrate/v0_15/migrate.go
```diff
@@ -191,3 +192,8 @@ func Committee(genesisState v0_14committee.GenesisState) v0_15committee.GenesisS
 	return v0_15committee.NewGenesisState(
 		genesisState.NextProposalID, committees, proposals, votes)
 }
+
+// Swap migrates new swap module state to a v0.15 swap genesis state
+func Swap(genesisState v0_15swap.GenesisState) v0_15swap.GenesisState {
```
nddeluca, line 197: Looks like the Swap state isn't called from MigrateAppState yet

nddeluca (changes requested):
Rebased all those commits out.  Looks great, will note to update the legacy types if we change the params or genesis state.

One small nit -  the Swap function isn't called in MigrateAppState

Also, thoughts on adding the swap module account in the migration?  It looks like we are also missing the validation for the module account in swap.InitGenesis.

migrate/v0_15/migrate.go
karzak: swap module account should be created in/checked to exist in `InitGenesis`. After that, the account state will be preserved in the `Auth` genesis state, so no need to do anything in Swap migration.

nddeluca: Sounds good, I like that approach.  Is the reason we created the module account for hard in migration because we needed `minter` permissions?

migrate/v0_15/migrate.go
denalimarsh: fixed in https://github.com/Kava-Labs/kava/pull/925/commits/d202e32a2eb3f245ad1d3c210f12025c8fc6585a

karzak (approved):

PR #926, opened 2021-06-10
(Swap) MsgDeposit Create Pool

nddeluca:
Handler integration test for acceptance  along with boilerplate for a successful compile. Full implementation will follow to pass the assertions in the handler.


denalimarsh:
Should the base branch for this PR be `swap-acceptance`?

x/swap/keeper/deposit.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/kava-labs/kava/x/swap/types"
+)
+
+func (k Keeper) Deposit(ctx sdk.Context, depositor sdk.AccAddress, amountA sdk.Coin, amountB sdk.Coin) error {
```
denalimarsh, line 11:
```suggestion
func (k Keeper) Deposit(ctx sdk.Context, depositor sdk.AccAddress, coinA sdk.Coin, coinB sdk.Coin) error {
```

Using the naming convention of 'amount' feels weird because each coin has an `amount` and `denom`

x/swap/types/errors.go
```diff
@@ -4,8 +4,7 @@ import (
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 )
 
-// DONTCOVER
-
 var (
-	ErrCustom = sdkerrors.Register(ModuleName, 2, "")
+	ErrNotAllowed     = sdkerrors.Register(ModuleName, 1, "not allowed")
```
denalimarsh, line 8:
```suggestion
	ErrNotAllowed     = sdkerrors.Register(ModuleName, 2, "not allowed")
```
Code 1 is reserved, custom error codes should start from 2

x/swap/types/errors.go
```diff
@@ -4,8 +4,7 @@ import (
...
+	ErrNotImplemented = sdkerrors.Register(ModuleName, 99, "not implemented")
```
denalimarsh, line 9:
```suggestion
	ErrNotImplemented = sdkerrors.Register(ModuleName, 3, "not implemented")
```
Our standard so far has been to order error codes in sequential order

x/swap/keeper/deposit.go
```diff
@@ -0,0 +1,53 @@
...
+	for _, p := range params.AllowedPools {
+		if p.TokenA == amountA.Denom && p.TokenB == amountB.Denom {
+			creationAllowed = true
+		}
+	}
+	if !creationAllowed {
+		return sdkerrors.Wrap(types.ErrNotAllowed, fmt.Sprintf("can not create pool '%s'", poolName))
+	}
+
+	// TODO: extract method, wrap error
```
denalimarsh, line 31: Not convinced that we need to extract this into its own method

denalimarsh: I've left this as a TODO in https://github.com/Kava-Labs/kava/pull/930

denalimarsh (changes requested):
Generally, we've colocated all the low-level store manipulation functions (get, set, delete) in `keeper.go` with the higher-level functions in their respective files. I'm fine breaking with that standard as I think the approach of having the keeper functions for different store keys packaged together is potentially cleaner but just good to be aware that we'll be breaking with existing standards.

I implemented most of the outstanding keeper/type TODOs in https://github.com/Kava-Labs/kava/pull/930.

x/swap/types/errors.go
nddeluca: good catch 👍

x/swap/keeper/deposit.go
nddeluca: 👍 didn't like amount either, coin is much clearer

x/swap/types/errors.go
nddeluca: 👍

x/swap/keeper/deposit.go
nddeluca: 👍 sounds good

app/app.go
```diff
@@ -105,6 +105,7 @@ var (
 		kavadist.ModuleName:         {supply.Minter},
 		issuance.ModuleAccountName:  {supply.Minter, supply.Burner},
 		hard.ModuleAccountName:      {supply.Minter},
+		swap.ModuleAccountName:      nil,
```
karzak, line 108: Just want to note that if we support LP tokens in a future upgrade, we'll have to update this to Minter + Burner here AND write a migration for it (simply updating permissions in app.go doesn't propagate during an upgrade).

x/swap/keeper/deposit.go
```diff
@@ -0,0 +1,62 @@
...
+		return err
+	}
+
+	amount := sdk.NewCoins(coinA, coinB)
+	err = k.supplyKeeper.SendCoinsFromAccountToModule(ctx, depositor, types.ModuleAccountName, amount)
+	if err != nil {
+		return err
+	}
+
+	k.initializePool(ctx, depositor, coinA, coinB)
```
karzak, line 30: error value not checked.  I would also put this before the call the `SendCoinsFromAccountToModule`

x/swap/keeper/deposit.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/kava-labs/kava/x/swap/types"
+)
+
+func (k Keeper) Deposit(ctx sdk.Context, depositor sdk.AccAddress, coinA sdk.Coin, coinB sdk.Coin) error {
```
karzak, line 11: add godoc comments, here and in keeper.go

karzak (changes requested):
Looking pretty solid, just a few minor things.

app/app.go
nddeluca: 👍 good reminder, I forgot that migration is required once update

nddeluca: added code comment for next PR for reminder

x/swap/keeper/deposit.go
nddeluca: resolved in other work

x/swap/keeper/deposit.go
nddeluca: added in new work

PR #932, opened 2021-06-16
Proposal: remove broken link check from CI

rhuairahrighairidh:
CI keeps failing whenever an external website linked to from our docs goes down.
The link check command is still in the makefile for one off checks.


denalimarsh (approved):
I'm strongly in support of this proposal

PR #935, opened 2021-06-17
(Swap) Keeper Deposit to Existing Pool, Complete Pool Liquidity & Share Base Implementation

nddeluca:
Work here is mostly the separation of keeper state and pool functionality, and the expansion of deposit to add liquidity to existing pools. The keeper records are now defined in `types/state` and are basic value objects.

The pool functionality has been moved and split between `DenominatedPool` in `types/denominated_pool.go`, and `BasePool` in `types/base_pool.go`.

The base pool encapsulates the pool functionality, while the denominated pool wraps it and converts to/from the correct coin denoms. Swap functions in both are stubs right now, but adding liquidity, removing liquidity, and share handling is complete and robust.


karzak (changes requested):
Tested deposits with slippage, working as expected. 

Feel free to merge after making these minor edits, only real change I could find was in the client.

PR #938, opened 2021-06-23
Disable incentive sync on export

rhuairahrighairidh:
This PR simplifies the genesis export in v0.14.2 by avoiding syncing on during export. The aim is to make it easier to do the migration and any state bug fixes for kava-8, by releasing this as a v0.14.3. It should be backwards compatible with v0.14.x, so validators can upgrade whenever they like.

Changes
- remove claim sync on export
- add reward indexes to the genesis state (as without syncing they can no longer be assumed to be all zero)
- add more test coverage for keeper methods


karzak (approved):
tACK - tested on current mainnet and genesis export looks good. Note that these changes will need to be backported and rebased on master.

PR #944, opened 2021-07-01
Implement Base Pool Swap Methods

nddeluca:
This finishes the `BasePool` implementation.

Trading coins with a fixed input can be done with `SwapExactAForB` and `SwapExactBForA`, and trading coins with an exact output can be with `SwapAForExactB` and `SwapAForExactB`.  The `DenominatedPool` wraps these methods and provides an interface of `SwapExactInput` and `SwapExactOutput`, which handle the from and to conversions of the coin denominations.

The swap methods panic when provided with non-positive values, and the swap for exact output method panics if the output is greater than or equal to the reserves.

Fee validation is changed to require a fee less than 1.  This prevents division by zero when calculating the input fee for an exact output.  The swap methods will panic if the fee is less than 0 or greater or equal to 1.

The invariant of `(inputReserves + (input - fee)) * (outputReserves - output) >= inputReserves * outputReserves` holds for both exact input and exact output swaps.  The invariant requires us to floor on exact input swaps, and ceil on exact output swaps.  Flooring decreases the output when rounding errors occur, and ceiling increases in the input when rounding errors occur -- both resulting in K being increased and benefiting the pool on rounding errors.  The invariant is checked as a runtime assertion to catch any bugs in the math during simulations, and help ensure any enhancements later do not result in regression.

### DIFF
--- a/x/swap/types/base_pool.go
+++ b/x/swap/types/base_pool.go
@@ -218,7 +218,8 @@ func (p *BasePool) RemoveLiquidity(shares sdk.Int) (sdk.Int, sdk.Int) {
 	return withdrawA, withdrawB
 }
 
-// SwapAForB trades a for b with a percentage fee, returns b and fee amount in a
+// SwapExactAForB trades an exact value of a for b.  Returns the positive amount b
+// that is removed from the pool and the portion of a that is used for paying the fee.
 func (p *BasePool) SwapExactAForB(a sdk.Int, fee sdk.Dec) (sdk.Int, sdk.Int) {
 	b, feeValue := p.swapExactInput(a, p.reservesA, p.reservesB, fee)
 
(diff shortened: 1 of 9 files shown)


x/swap/types/base_pool.go
```diff
@@ -248,6 +336,48 @@ func (p *BasePool) ShareValue(shares sdk.Int) (sdk.Int, sdk.Int) {
 	return sdk.NewIntFromBigInt(&resultA), sdk.NewIntFromBigInt(&resultB)
 }
 
+// assertInvariantAndUpdateRerserves asserts the constant product invariant is not violated, subtracting
+// any fees first, then updates the pool reserves.  Panics if invariant is violated.
+func (p *BasePool) assertInvariantAndUpdateRerserves(newReservesA, feeA, newReservesB, feeB sdk.Int) {
```
karzak, line 341: Rerserves -> Reserves

karzak (approved):
Logic looks good

PR #949, opened 2021-07-07
Swap queries: deposit, pool, pools

denalimarsh:
Implement Swap module queriers for
- deposits
- pool
- pools


karzak (approved):
tACK - functioning well with kvtool - didn't check every edge case, so will be worth it to give these queries a workout on internal/external testnets.

PR #950, opened 2021-07-07
Swap users accumulate rewards

rhuairahrighairidh:
This PR adds rewards incentivising swap deposits.
It is missing the ability to claim swap rewards. This will be added in a future PR.

Main Changes
- Add a new SwapClaim type (and associated keeper Get/Set methods and genesis udpates)
- Add Sync and Init methods for SwapClaims mirroring other reward types
- Adds hooks from the swap module to incentive to call Sync and Init

I revised the tests for Sync and Init to ensure all code paths were covered. If I get time I'll tidy up the tests for the other rewards to match (in future PR).

The hooks in swap module are not connected yet as this is based off master. I ran a trial integration with `swap-acceptance` and it seemed to be working. I'll add some app/integration tests once the branches are merged.

PR #951, opened 2021-07-08
Update querier and client for swap claims

rhuairahrighairidh:
This PR is a follow on to #950 adding in the query methods. Merge this one first.

Main changes
- Refactor existing querier and rest/cli code to dedupe it. The rest/cli endpoints are the same, but the rpc endpoints were simplified.
- Restructure `reward-factors` query response so it matches the data more closely. Simplify by removing ability to filter by rewarded denom/collateral type. Don't think anyone uses this endpoint.
- Add swap claims querying to querier
- Refactor the main swap syncing method so you can sync a claim without setting it in the store. Add a `GetSyncedSwapClaim` method used by the querier (and soon the payout methods).

I haven't added tests around the querier. Not great practise but I'm punting on it to get the remaining state machine fixes in before code freeze. Bugs will probably be caught in bugbash.


x/incentive/keeper/rewards_swap.go
```diff
@@ -92,5 +98,24 @@ func (k Keeper) SynchronizeSwapReward(ctx sdk.Context, poolID string, owner sdk.
 	claim.Reward = claim.Reward.Add(newRewards...)
 	claim.RewardIndexes = claim.RewardIndexes.With(poolID, globalRewardIndexes)
 
-	k.SetSwapClaim(ctx, claim)
+	return claim
+}
+
+// GetSynchronizedSwapClaim fetches a swap claim from the store and syncs rewards for all pools.
+func (k Keeper) GetSynchronizedSwapClaim(ctx sdk.Context, owner sdk.AccAddress) (types.SwapClaim, bool) {
```
karzak, line 105: +1 this is way cleaner than the previous sync methods, should refactor the others as well when we have time.

karzak (approved):
Changes look good. As you mention, worth exercising these queries in bugbash/testnet.

x/incentive/client/cli/query.go
```diff
@@ -77,89 +77,51 @@ func queryRewardsCmd(queryRoute string, cdc *codec.Codec) *cobra.Command {
...
-					if err != nil {
-						return err
-					}
+				params := types.NewQueryRewardsParams(page, limit, owner, boolUnsynced)
+				claims, err := executeDelegatorRewardsQuery(queryRoute, cdc, cliCtx, params)
+				if err != nil {
+					return err
+				}
+				return cliCtx.PrintOutput(claims)
+			case "swap":
```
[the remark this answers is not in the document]
rhuairahrighairidh: added in https://github.com/Kava-Labs/kava/pull/951/commits/8c957890243e146c6c32f4b272584abe32f57f06

x/incentive/keeper/rewards_swap.go
rhuairahrighairidh: Hoping to get to it before code freeze

PR #952, opened 2021-07-08
Support Multiple Deposits

nddeluca:
Update a depositors existing shares when adding liquidity instead of overwriting them.


karzak (approved):
tACK - tested with kvtool (along with queries to expose pool stats/deposits). Multiple deposits working as expected.

PR #954, opened 2021-07-12
Incentive genesis refactor: remove sync on export

rhuairahrighairidh:
This PR ports the changes to v0.14 genesis in #938 to the new version.

Main changes
- remove claim sync during export
- add reward indexes to the genesis state (as without syncing they can no longer be assumed to be all zero)
  - unlike v0.14 I grouped accumulation times and reward indexes into a struct to keep the genesis state a bit tidier
- update the migration
- update the test genesis builder type so that global indexes are initialized like the old init genesis used to do. This makes all the old tests work again, but doesn't mirror mainnet that well. On mainnet global indexes are not set if they are zero. If I get time I'll do some refactoring of the integration tests to fix this.

The TODOs in migrations will be picked up when we finalise mainnet params.


karzak (approved):
Added a copy of kava-7 incentive mainnet state exported using the updated export genesis from #938

PR #955, opened 2021-07-12
Users can claim swap rewards

rhuairahrighairidh:
This PR adds msgs for claiming swap rewards, and integrates incentive fully with the swap module.

The hooks in the swap module aren't super elegant. Open to ideas on how to improve them.

There's one msg for each claim type (hard, usdx, swap, delegator) and a duplicated set for claiming from a validator vesting accounts. Each msg (except usdx) has a new field `DenomsToClaim []string` that will claim a subset of the denoms from the reward. An empty value will claim all rewards.
I played around with other ways of structuring the messages and this felt like the least bad way of doing it. In a future refactor it would be great to combine them all into just one msg, and generally reduce duplication between reward types.

Main Changes:
- [x] Add the ability to claim a subset of rewards to delegator rewards (eg claim only SWAP tokens, leaving HARD rewards in claim)
- [x] Add subset claiming to other msgs
- [x] Implements hooks in the swap module
- [x] Add `MsgClaimSwapReward` and `MsgClaimSwapRewardVVesting` and associated handlers, client, tests, etc
- [x] Add more integration tests


x/incentive/types/msg_test.go
```diff
@@ -1,72 +1,523 @@
...
+	validAddress := sdk.AccAddress(crypto.AddressHash([]byte("KavaTest1")))
+
+	type expectedErr struct {
+		wraps error
+		pass  bool
+	}
+	type msgArgs struct {
+		sender         sdk.AccAddress
+		multiplierName string
+		denomsToClaim  []string
```
karzak, line 434: denomsToClaim isn't used in this test

karzak (approved):
tACK - tested pretty thoroughly with kvtool, ran some basic code checks - looking pretty clean.

x/incentive/types/msg_test.go
rhuairahrighairidh: oops, fixed in https://github.com/Kava-Labs/kava/pull/955/commits/1c244265bdbd8db4499d223ccf0183162a1603e9

PR #959, opened 2021-07-13
Add E2E Swap Support

nddeluca:
Adds support for swapping with exact input and exact outputs with slippage and deadline.

- [x] Swap Messages
- [x] Swap Events
- [x] Handler, Keeper w/ test coverage of swap functionality
- [x] Add keeper tests to ensure only spendable coins are tradable
- [x] Add test coverage for pool & denominated pool errors/panics
- [x] Validate exact swap output is less than pool reserves in keeper
- [x] Test panic on module account not enough funds
- [x] Add rest tx endpoints for swap messages
- [x] Validate that swap output is greater than zero on ExactForTokens


karzak (approved):
tACK - just a few nits

PR #966, opened 2021-07-15
update swap spec: add msgs, state, events

karzak:
Documents swap messages, state transitions, genesis state, and events.


nddeluca (approved):
looks good

PR #967, opened 2021-07-15
Hook Regression Test + Deposit Refactor

nddeluca:
- [x] Add regression test for hooks (added before deposit refactoring)
- [x] Refactor Deposit pool and share commit
- [x] Call keeper public hooks to prevent panic when hooks are not set
- [x] Add panic on SetPool and SetDepositorShares if the records are invalid
- [x] Dangling Start on Invariant Test Suite (will complete in next PR)


x/swap/keeper/hooks_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+	// third deposit into pool calls BeforePoolDepositModified with shares from last deposit
+	swapHooks.On("BeforePoolDepositModified", suite.Ctx, types.PoolIDFromCoins(deposit), depositor_2.GetAddress(), shareRecord.SharesOwned).Once()
+	err = suite.Keeper.Deposit(suite.Ctx, depositor_2.GetAddress(), sdk.NewCoin("ukava", sdk.NewInt(3e6)), sdk.NewCoin("usdx", sdk.NewInt(15e6)), sdk.MustNewDecFromStr("0.0015"))
+	suite.Require().NoError(err)
+
+	// test hooks with a full withdraw of all shares
+	shareRecord, found = suite.Keeper.GetDepositorShares(suite.Ctx, depositor_1.GetAddress(), types.PoolIDFromCoins(deposit))
+	suite.Require().True(found)
+	// all shares given to before deposito modified
```
rhuairahrighairidh, line 84:
```suggestion
	// all shares given to before deposit modified
```

x/swap/keeper/deposit.go
```diff
@@ -75,6 +76,15 @@ func (k Keeper) Deposit(ctx sdk.Context, depositor sdk.AccAddress, coinA sdk.Coi
 		return sdkerrors.Wrapf(types.ErrSlippageExceeded, "slippage %s > limit %s", slippage, slippageLimit)
 	}
 
+	k.updatePool(ctx, poolID, pool)
+	if shareRecord, hasExistingShares := k.GetDepositorShares(ctx, depositor, poolID); hasExistingShares {
+		k.BeforePoolDepositModified(ctx, poolID, depositor, shareRecord.SharesOwned)
+		k.updateShares(ctx, depositor, poolID, shareRecord.SharesOwned.Add(shares))
+	} else {
+		k.AfterPoolDepositCreated(ctx, poolID, depositor, shares)
+		k.updateShares(ctx, depositor, poolID, shares)
```
rhuairahrighairidh, line 85: nit: I'd maybe switch the order of these just to keep the `After...` name accurate. Then if any hook listener calls `GetDepositorShares`, the new shares will be in the store.

rhuairahrighairidh (approved):
Looks good. Thanks for cleaning up the hook calls.

karzak (approved):
tACK

x/swap/keeper/deposit.go
nddeluca: Great catch! Wasn't thinking about ordering.  I added a regression test to cover all hook ordering, and fixed

PR #968, opened 2021-07-15
Update Pool ID Separator

nddeluca:
Updates the pool id separator to be `:` - `ukava:usdx`, etc.

In addition, updates the store key separator to be `|` in order to not conflict with the pool key separator.


karzak (approved):
Looks good.

PR #971, opened 2021-07-21
Separate multipliers per claimed denom

rhuairahrighairidh:
This PR changes claiming so that the user must specify a separate multiplier for each denom they're claiming.
In the params, there is a separate set of available multipliers for each reward denom.

External Changes:
- msgs and params have changed format
- It is no longer possible to claim all denoms automatically. Users must fetch their claimable rewards from their claim object, then constuct a claim msg using those denoms, specifying a multiplier for each one.
- If we add a new reward denom to params, we must also add a new set of multipliers. Otherwise users will not be able to claim the new denom.

The current implementation has high gas costs. If I get time it's probably worth optimising it given how frequent claims are. Each claimed denom is basically treated as a separate claim msg, which already were pretty gas expensive.


x/incentive/keeper/claim.go
```diff
@@ -140,8 +152,10 @@ func (k Keeper) ClaimDelegatorReward(ctx sdk.Context, owner, receiver sdk.AccAdd
 		return sdkerrors.Wrapf(types.ErrClaimNotFound, "address: %s", owner)
 	}
 
-	claimingCoins := types.FilterCoins(syncedClaim.Reward, denomsToClaim)
-	rewardCoins := types.MultiplyCoins(claimingCoins, multiplier.Factor)
```
karzak, line 144: Pretty sure these methods aren't used anymore and can be deleted from `types`

karzak (approved):
tACK - changes look good, just some dead code to remove. Can merge after making those changes.

PR #973, opened 2021-07-22
Fix bug in syncing rewards

rhuairahrighairidh:
The `GetSynchronizedSwapClaim` method had a bug where it would not sync rewards for recently rewarded pools.

For example. Given a user with a hard:usdx swap deposit, and that pool received no rewards. If rewards were added (via gov) then the user claimed their swap reward, they would miss out on rewards for their hard:usdx deposit.
Though if they ever modified their deposit it would sync correctly.

I also review hard and delegator for the same behaviour.

Note: I'll rebase this onto master once #971 is merged.


rhuairahrighairidh:
Closing in favour of #977 which bases the changes on master

PR #974, opened 2021-07-25
Fix incentive usdx/borrow/supply reward calculation bug

rhuairahrighairidh:
This PR fixes calculation of borrow, supply, and cdp rewards in claim syncing. Calculations were using raw borrow/supply amounts rather than "normalized" amounts.

Main Changes
- Use normalized borrow/supply/usdx amounts in reward syncing
- Add integration tests to demonstrated the bug
- Extract sync methods that don't write to state. This should make further refactors easier (such as removing simulate methods)

Background:
Rewards are paid out in porportion to a user's share of a rewarded activity, eg their percentage of the total cdp borrows.
The values used to calculate this percentage are referred to as "source shares" and "total source shares". In swap this is just the pool shares. One divided by the other gives the percentage of the pool a  user owns.

However in hard and cdp, users' borrows/deposits technically change every block as interest is accumulated. The syncing optimisation in incentive requires a sync before any change, so we'd need to sync every block.
We can get around this by calculating the percentage using the "normalized" borrow/deposits. Normalized amounts are a sort of "pre-interest" value of the borrow/deposit.

Prior to the PR we were using the normalized total borrow/deposits in accumulation, but not for syncing indivudual claims.


x/incentive/keeper/rewards_borrow.go
```diff
@@ -38,8 +38,14 @@ func (k Keeper) AccumulateHardBorrowRewards(ctx sdk.Context, rewardPeriod types.
 }
 
 // getHardBorrowTotalSourceShares fetches the sum of all source shares for a borrow reward.
-// In the case of hard borrow, this is the total borrowed divided by the borrow interest factor.
-// This give the "pre interest" value of the total borrowed.
+//
+// In the case of hard borrow, this is the total borrowed divided by the borrow interest factor (for a particular denom).
+// This give the "pre interest" or "normalized" value of the total borrowed. This is an amount, that if it was borrowed when
```
karzak, line 43: This give -> This gives

karzak (approved):
tACK

PR #977, opened 2021-07-26
Fix bug in syncing swap rewards

rhuairahrighairidh:
This is a resubmission of #973 having rebased the changes

The GetSynchronizedSwapClaim method had a bug where it would not sync rewards for recently rewarded pools.

For example. Given a user with a hard:usdx swap deposit, and that pool receiving no rewards. If rewards were added (via gov) and the user claimed their swap reward, they would miss out on rewards for their hard:usdx deposit.
Though if they ever modified their deposit it would sync correctly.

I also reviewed hard and delegator for the same behaviour.


karzak (approved):
tACK.

As a follow up, probably worth documenting the `Initialize` methods for all the rewards a little more fully in the spec. Just to make it clear what reward factors should/should not be present.

PR #987, opened 2021-08-10
Reset Periodic Vesting Accounts for Kava-8

nddeluca:
This PR removes all **vested** periods from periodic vesting accounts at the time genesis of kava-8.

This is done to optimize gas, transaction, and storage costs due to a high number of already vested periods for some users.

If a periodic vesting account is fully vested (has zero vested periods), the account will be converted back into a base account with no vesting information.

If the original vesting amount left is less than the delegated vesting amount, the delegated vesting amount is set to the original vesting value, and the change in delegated vesting is added to delegated free.  The total delegated funds does not change.

The unit tests cover all the edge cases I could think of in the vesting period and delegation handling, and the migrate tests assert the following:
- The number of accounts after the migration is the same
- There are no vested periods left for periodic vesting accounts.
- There are no periodic vesting accounts with zero vesting periods
- Delegated vesting is not greater than original vesting for all periodic vesting accounts
- The total owned coins (base coins) for all accounts is not modified
- The spendable coins at genesis time, 30 days after, 90 days after, 180 days after and 365 days after are equal for the unmigrated and migrated accounts.

### DIFF
--- a/migrate/v0_15/account_test.go
+++ b/migrate/v0_15/account_test.go
@@ -59,7 +59,7 @@ func TestMigrateAccount_PeriodicVestingAccount_Vesting(t *testing.T) {
 		},
 		vesting.Period{
 			Length: 45 * 24 * 60 * 60, // +15 days - vesting
-			Amount: sdk.NewCoins(sdk.NewCoin("ukava", sdk.NewInt(1e6))),
+			Amount: sdk.NewCoins(sdk.NewCoin("ukava", sdk.NewInt(2e6))),
 		},
 	}
 
@@ -68,8 +68,15 @@ func TestMigrateAccount_PeriodicVestingAccount_Vesting(t *testing.T) {
 
 	MigrateAccount(vacc, genesisTime)
 
+	expectedVestingPeriods := vesting.Periods{
+		vesting.Period{
+			Length: 15 * 24 * 60 * 60,
+			Amount: sdk.NewCoins(sdk.NewCoin("ukava", sdk.NewInt(2e6))),
+		},
+	}
+
 	assert.Equal(t, genesisTime.Unix(), vacc.StartTime, "expected vesting start time to equal genesis time")
-	assert.Equal(t, 1, len(vacc.VestingPeriods), "expected only one vesting period left")
+	assert.Equal(t, expectedVestingPeriods, vacc.VestingPeriods, "expected only one vesting period left")
 }
 
 func TestResetPeriodVestingAccount_NoVestingPeriods(t *testing.T) {
@@ -218,7 +225,7 @@ func TestResetPeriodVestingAccount_MultiplePeriods(t *testing.T) {
 		},
 	}
 
-	assert.Equal(t, sdk.NewCoins(sdk.NewCoin("ukava", sdk.NewInt(2e6))), vacc.OriginalVesting, "expected original vesting to be unchanged")
+	assert.Equal(t, sdk.NewCoins(sdk.NewCoin("ukava", sdk.NewInt(2e6))), vacc.OriginalVesting, "expected original vesting to be updated")
 	assert.Equal(t, newVestingStartTime.Unix(), vacc.StartTime, "expected vesting start time to be updated")
 	assert.Equal(t, expectedEndtime, vacc.EndTime, "expected vesting end time end at last period")
 	assert.Equal(t, expectedPeriods, vacc.VestingPeriods, "expected vesting periods to be updated")
--- a/migrate/v0_15/account_test.go
+++ b/migrate/v0_15/account_test.go
@@ -28,9 +28,10 @@ func TestMigrateAccount_BaseAccount(t *testing.T) {
 	addr := sdk.AccAddress(pub.Address())
 	acc := auth.NewBaseAccount(addr, balance, pub, 1, 1)
 
-	migratedAcc := MigrateAccount(acc, time.Now())
+	accCopy := *acc
+	migratedAcc := MigrateAccount(&accCopy, time.Now())
 
-	assert.Equal(t, acc, migratedAcc)
+	assert.Equal(t, acc, migratedAcc, "expected account to be unmodified")
 }
 
 func TestMigrateAccount_PeriodicVestingAccount_NoPeriods(t *testing.T) {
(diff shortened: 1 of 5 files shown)


migrate/v0_15/account_test.go
```diff
@@ -0,0 +1,284 @@
...
+			Length: 0, // 0 days
+			Amount: sdk.NewCoins(sdk.NewCoin("ukava", sdk.NewInt(1e6))),
+		},
+		vesting.Period{
+			Length: 15 * 24 * 60 * 60, // 15 days
+			Amount: sdk.NewCoins(sdk.NewCoin("ukava", sdk.NewInt(1e6))),
+		},
+	}
+
+	assert.Equal(t, sdk.NewCoins(sdk.NewCoin("ukava", sdk.NewInt(2e6))), vacc.OriginalVesting, "expected original vesting to be unchanged")
```
karzak, line 221: "expected original vesting to be changed". correct? Looks like it goes from 3e6 to 2e6

migrate/v0_15/account_test.go
```diff
@@ -0,0 +1,284 @@
...
+	spendableBefore := vacc.SpendableCoins(newVestingStartTime)
+
+	ResetPeriodicVestingAccount(vacc, newVestingStartTime)
+
+	// new period length 15 days
+	expectedEndtime := newVestingStartTime.Add(15 * 24 * time.Hour).Unix()
+	// new period length changed, amount unchanged
+	expectedPeriods := vesting.Periods{
+		vesting.Period{
+			Length: 0, // 0 days
```
karzak, line 212: Bit of a nit/extreme edge case, but why keep a vesting period with length 0?

migrate/v0_15/account_test.go
```diff
@@ -0,0 +1,284 @@
...
+		},
+	}
+
+	vacc := createVestingAccount(balance, vestingStartTime, periods)
+	genesisTime := vestingStartTime.Add(30 * 24 * time.Hour)
+
+	MigrateAccount(vacc, genesisTime)
+
+	assert.Equal(t, genesisTime.Unix(), vacc.StartTime, "expected vesting start time to equal genesis time")
+	assert.Equal(t, 1, len(vacc.VestingPeriods), "expected only one vesting period left")
```
karzak, line 72: probably worth specifying the exact expected periods

migrate/v0_15/migrate_test.go
```diff
@@ -99,3 +103,89 @@ func TestIncentive(t *testing.T) {
...
+func TestAuth_AccountConversion(t *testing.T) {
+	bz, err := ioutil.ReadFile(filepath.Join("testdata", "kava-7-test-auth-state.json"))
+	require.NoError(t, err)
+
+	var genesisState auth.GenesisState
+	cdc := app.MakeCodec()
+	cdc.MustUnmarshalJSON(bz, &genesisState)
+
+	migratedGenesisState := Auth(genesisState, GenesisTime)
+	require.Equal(t, len(genesisState.Accounts), len(migratedGenesisState.Accounts), "expected the number of accounts after migration to be equal")
```
karzak, line 148: I would check that the resulting genesis is valid:

```
err = auth.ValidateGenesis(migratedGenesisState)
require.NoError(t, err)
```

migrate/v0_15/migrate_test.go
```diff
@@ -99,3 +103,89 @@ func TestIncentive(t *testing.T) {
...
+			require.Equal(t, oldTotalDelegated, newTotalDelegated, "expected total amount of tracked delegations to not change")
+
+			// delegated vesting must be less or equal to original vesting
+			require.True(t, vacc.DelegatedVesting.IsAllLTE(vacc.OriginalVesting), "expected delegated vesting to be less or equal to original vesting")
+
+			// vested coins must be nil for the new account
+			require.Equal(t, sdk.Coins(nil), vacc.GetVestedCoins(GenesisTime), "expected no vested coins at genesis time")
+
+			// vesting coins must not be nil
+			require.NotEqual(t, sdk.Coins(nil), vacc.GetVestingCoins(GenesisTime), "expected vesting coins to be greater than 0")
```
karzak, line 188: could maybe add an assertion here than the number of vesting periods on the new vesting account is less than or equal to the number of vesting periods on the old vesting account.

migrate/v0_15/migrate_test.go
```diff
@@ -99,3 +103,89 @@ func TestIncentive(t *testing.T) {
...
+		futureDate = GenesisTime.Add(90 * 24 * time.Hour)
+		require.Equal(t, oldAcc.SpendableCoins(futureDate), acc.SpendableCoins(futureDate), "expected spendable coins to not chain")
+		// check 180 days
+		futureDate = GenesisTime.Add(180 * 24 * time.Hour)
+		require.Equal(t, oldAcc.SpendableCoins(futureDate), acc.SpendableCoins(futureDate), "expected spendable coins to not chain")
+		// check 365 days
+		futureDate = GenesisTime.Add(365 * 24 * time.Hour)
+		require.Equal(t, oldAcc.SpendableCoins(futureDate), acc.SpendableCoins(futureDate), "expected spendable coins to not chain")
+
+		if vacc, ok := acc.(vesting.PeriodicVestingAccount); ok {
```
karzak, line 171:
```
if vacc, ok := acc.(*vesting.PeriodicVestingAccount)
```

migrate/v0_15/migrate_test.go
```diff
@@ -99,3 +103,89 @@ func TestIncentive(t *testing.T) {
...
+			// old account must be a periodic vesting account
+			oldVacc, ok := oldAcc.(vesting.PeriodicVestingAccount)
```
karzak, line 173: oldVacc, ok := oldAcc.(*vesting.PeriodicVestingAccount)

migrate/v0_15/migrate_test.go
```diff
@@ -99,3 +103,89 @@ func TestIncentive(t *testing.T) {
...
+	cdc := app.MakeCodec()
+	cdc.MustUnmarshalJSON(bz, &genesisState)
+
+	migratedGenesisState := Auth(genesisState, GenesisTime)
+
+	assert.Equal(t, genesisState.Params, migratedGenesisState.Params, "expected auth parameters to not change")
+}
+
+func TestAuth_AccountConversion(t *testing.T) {
+	bz, err := ioutil.ReadFile(filepath.Join("testdata", "kava-7-test-auth-state.json"))
```
karzak, line 140: Mostly curiosity, but do you know at what block this was exported?

karzak (changes requested):
Core logic looks good, mostly minor suggestions

migrate/v0_15/account_test.go
nddeluca: It's still vesting at the newStartTime.  The comparison `currentTime < startTime + periodLength`.  If we do not keep the period with zero length, then the spendable balance will increase and not be equal to the previous spendable balance at newStartTime.

migrate/v0_15/migrate_test.go
nddeluca: No, looks like it was a current height export on 2021-08-05.  I can update to a specific height

karzak (approved):
LGTM

PR #991, opened 2021-08-12
translation

Jewcub:


Jewcub:
and a few tiny errors/improvements in the English one
- UTG -> UTC
- add 'sh' to code examples

PR #993, opened 2021-08-12
Fix Account Migration for Additive Vesting Periods

nddeluca:
Fixes the periodic vesting account migration to properly compare old accounts with new accounts in the acceptance test, and updates the reset logic to treat period lengths as additive to correct failing acceptance.

Also, adds support for running v0_15 migration tests in CI.  Requires #994 for tests to pass.


karzak (approved):
tACK - manually tested a few vesting accounts of interest and can confirm that the resulting account state is as expected.

rhuairahrighairidh (approved):
looks good, walked through all the tests and everything makes sense to me

PR #996, opened 2021-08-13
SWP airdrop for USDX suppliers to hard

karzak:
Uses a snapshot a hard deposits from block 1543671 (Aug 4th, 14:00 UTC). Allocates 1,000,000 SWP to USDX depositors by weighing each USDX deposit by the deposit amount and splitting it into buckets (same method as we use in `Auction` module). Tests verify that total allocation is exactly 1M SWP, the total supply of SWP, and that all USDX depositors are at least included in the map (note that some deposits of USDX to Hard were small enough that the rounded airdrop amount is 0).


rhuairahrighairidh (changes requested):
Code looks good. Though I think relying on the json file in data/ will cause problems for some validators with fancy setups.
I think embedding the json as a string in a go file is the easiest solution, but we should sidebar about it. I have the same problem in the incentive usdx migrations.

rhuairahrighairidh (approved):
looks good.

Why did testdata/kava-7-auth-state.json change?

karzak:
> looks good.
> 
> Why did testdata/kava-7-auth-state.json change?

The previous `kava-7-auth-state.json` was taken before block 1543671, which didn't have all the accounts that get the SWP airdrop - which caused tests to fail.

PR #997, opened 2021-08-16
add state to swap genesis

karzak:
Adds genesis state to swap module. Note that incentives and token distirubition will be addressed in a follow on PR.


denalimarsh (approved):
LGTM

PR #998, opened 2021-08-16
Incentive migration state fixes: USDX Minting

rhuairahrighairidh:
This PR migrates usdx minting claims and ensures the output state is valid.

Main Changes
- All usdx minting claim indexes are overwritten with the current global indexes. This is needed as when params were removed usdx claims stopped syncing (so claim indexes stopped updating). This enables rewards to be switched back on safely.
- The above step would drop any rewards left unsynced at time of param removal. So these are added back in. The package `usdx_rewards` calculates what these rewards should be and writes the result to `rewards.go` which is included in the migration.
- Matching the hard PR, I changed the big Incentive test to use a full genesis file but did not commit the file. To run tests add a genesis file and comment out`t.Skip()`. I added several tests to ensure the usdx claims are valid and match all the cdps.

TODO
- [x] Export old state from archive node and regenerate missed rewards
- [x] Add more tests and validation to missing reward calculation.


migrate/v0_15/incentive.go
```diff
@@ -71,19 +75,21 @@ func Incentive(incentiveGS v0_14incentive.GenesisState) v0_15incentive.GenesisSt
 	)
```
karzak, line 76: What is the idea for this TODO? Is the thought to set previous accumulation time to genesis time, in case the export is missing one of the genesis accumulation times?

karzak (approved):
Functionality seems good, tested a full migration with state from kava-7, manually inspected a few claims. None of queestions are blocking, do think a test verifying the amount added to claims would be a good sanity check.

migrate/v0_15/incentive.go
rhuairahrighairidh: It was for just usdx rewards. When the params were removed on kava-7 the previous accumulation time stopped updating. So if rewards are ever enabled again, the begin blocker could try and accumulate rewards since params were removed.
It's not really necessary as the accumulators take into account the start time. It would only be a problem if rewards were added with a really old start time.

PR #1000, opened 2021-08-17
Swap audit error handling tasks

denalimarsh:
This PR isolates the recommended revisions for error handling.


x/swap/keeper/swap.go
```diff
@@ -103,7 +103,7 @@ func (k Keeper) commitSwap(
 	}
 
 	if err := k.supplyKeeper.SendCoinsFromModuleToAccount(ctx, types.ModuleAccountName, requester, sdk.NewCoins(swapOutput)); err != nil {
-		panic(err)
+		return err
```
nddeluca, line 106: This error is a panic since the module account will always have funds equal or greater than the swap output amount.  If this error occurs, it's a bug and the module invariant will be broken, so a panic is preferred.  Normal circumstances will never cause this error, and there isn't a way to sign a message that will produce this error.

x/swap/keeper/withdraw.go
```diff
@@ -35,12 +33,12 @@ func (k Keeper) Withdraw(ctx sdk.Context, owner sdk.AccAddress, shares sdk.Int,
...
 	poolRecord, found := k.GetPool(ctx, poolID)
 	if !found {
-		panic(fmt.Sprintf("pool %s not found", poolID))
+		return sdkerrors.Wrap(types.ErrPoolNotFound, poolID)
 	}
 
 	pool, err := types.NewDenominatedPoolWithExistingShares(poolRecord.Reserves(), poolRecord.TotalShares)
 	if err != nil {
-		panic(fmt.Sprintf("invalid pool %s: %s", poolID, err))
+		return sdkerrors.Wrapf(types.ErrInvalidPool, "%s: %s", poolID, err)
```
nddeluca, line 41: Ditto here as well -- a stored pool must always be valid.  If it's not, it's a bug.

x/swap/keeper/withdraw.go
```diff
@@ -57,7 +55,7 @@ func (k Keeper) Withdraw(ctx sdk.Context, owner sdk.AccAddress, shares sdk.Int,
 
 	err = k.supplyKeeper.SendCoinsFromModuleToAccount(ctx, types.ModuleAccountName, owner, withdrawnAmount)
 	if err != nil {
-		panic(err)
+		return err
```
nddeluca, line 58: Same as other ModuleToAccount

PR #1001, opened 2021-08-18
Fix hard incentive state in migrations

rhuairahrighairidh:
Migrate hard claims and ensure the migrated state is valid.

Changes
- refactor basic hard claim migration and add unit test
- add new claims for borrows/deposits without them. Drop any missed rewards
- fix indexes in all claims to ensure they match the denoms in borrows/deposits. I think these errors were a hold over from the original migration on kava-7.
- change acceptance test to use a full genesis file. I didn't commit a genesis file to avoid bloating the repo history. So to run these tests add a genesis file to `testdata` and comment out `t.Skip()`. I'm sure there's a better solution (git LFS?).
- add checks in the acceptance test to ensure migrated claims are valid. Ideally these checks would be invariants in incentive, but that would take longer to verify.


migrate/v0_15/incentive.go
```diff
@@ -86,30 +90,59 @@ func Incentive(incentiveGS v0_14incentive.GenesisState) v0_15incentive.GenesisSt
 	}
 
 	// Migrate Hard protocol claims (includes creating new Delegator claims)
+	hardClaims, delegatorClaims, err := migrateHardLiquidityProviderClaims(incentiveGS.HardLiquidityProviderClaims)
```
karzak, line 93: For claims that only have delegator reward indexes, the migration ends up creating an empty `HardLiquidityProviderClaim`, that looks like:

```
{
        "base_claim": {
            "owner": "kava1qqw4ym0h3x6cp0lgscz2jp8kujnny98j0pzjst",
            "reward": []
        },
        "supply_reward_indexes": [],
        "borrow_reward_indexes": []
},
```

I assume there's no risk for future missing/incorrect rewards with having these in the store?

karzak (changes requested):
Question about creating empty claims, otherwise looks good.

migrate/v0_15/incentive.go
rhuairahrighairidh: good catch. If they don't have any borrow or deposits it shouldn't cause problems, but probably best to fix. It avoids unecessary state.

rhuairahrighairidh: Actually, not inlcluding these claims a bit complex. It's hard to drop them without first cleaning up the old state. Some claims with deposits/borrows are missing indexes.

Extra claims don't cause problems, when a new deposit/borrow/delegation is created any indexes in the claim are overwritten.

It would still be nice to clear out unecessary state. But I'll try do another PR, and prioritise merging this and the delegator PR.

karzak (approved):

PR #1004, opened 2021-08-18
Incentive migration state fixes: delegator

rhuairahrighairidh:
Migrate delegator claims and ensure the migrated state is valid.

Main Changes
- Check for delegations without claims and add in new claims. Back calculated rewards are not added.
- Add in the same genesis test as in #1001 and #998. Check migrated delegator claims to ensure they will accrue rewards correctly in the future.


karzak (approved):
Migration looks good. Only note is similar to #1001  - this creates a bunch of empty `DelegatorClaim` objects, including for users that don't currently have a delegation.  Don't think that's a problem, but want to make sure.

PR #1006, opened 2021-08-18
feat: add swap incentives

karzak:
Defines SWP incentives for pools and KAVA delegators. 

Tests verify that yearly SWP rewards match what was specified in prop 59 and that all incentivized pools are indeed found in the swap genesis state.


rhuairahrighairidh (approved):
looks good, I cross checked numbers with the proposal

PR #1008, opened 2021-08-24
Add supply migration for kava-8

karzak:
Adds SWP token to `supply.GenesisState` so that `supply` module invariants pass

### DIFF
--- a/migrate/v0_15/migrate.go
+++ b/migrate/v0_15/migrate.go
@@ -31,6 +31,7 @@ var (
 	GenesisTime                  = time.Date(2021, 8, 30, 15, 0, 0, 0, time.UTC)
 	ChainID                      = "kava-8"
 	SwpDelegatorRewardsPerSecond = sdk.NewCoin("swp", sdk.NewInt(198186))
+	SwpTotalSupply               = sdk.NewCoin("swp", sdk.NewInt(250000000e6))
 )
 
 // Migrate translates a genesis file from kava v0.14 format to kava v0.15 format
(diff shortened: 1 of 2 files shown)


migrate/v0_15/migrate.go
```diff
@@ -258,6 +266,14 @@ func DistributeSwpTokens(genesisState auth.GenesisState) auth.GenesisState {
 	return auth.NewGenesisState(genesisState.Params, accounts)
 }
 
+// MigrateSupply reconciles supply from kava-3 to kava-4
```
nddeluca, line 269: Comment needs to be updated?

migrate/v0_15/migrate_test.go
```diff
@@ -36,6 +36,29 @@ func TestMain(m *testing.M) {
 	os.Exit(m.Run())
 }
 
+func TestMigrateFull(t *testing.T) {
```
nddeluca, line 39: Nice like this test 👍

migrate/v0_15/migrate.go
```diff
@@ -74,6 +74,14 @@ func MigrateAppState(v0_14AppState genutil.AppMap) {
 		v0_14AppState[auth.ModuleName] = v0_15Codec.MustMarshalJSON(Auth(v0_15Codec, authGenState, GenesisTime))
 	}
 
+	// Migrate supply app state
+	if v0_14AppState[supply.ModuleName] != nil {
+		var supplyGenState supply.GenesisState
+		v0_14Codec.MustUnmarshalJSON(v0_14AppState[supply.ModuleName], &supplyGenState)
+		delete(v0_14AppState, supply.ModuleName)
+		v0_14AppState[supply.ModuleName] = v0_15Codec.MustMarshalJSON(Supply(supplyGenState, sdk.NewCoin("swp", sdk.NewInt(250000000e6))))
```
nddeluca, line 82: 250 million looks correct. @rhuairahrighairidh agree?

nddeluca: Might be nice to put in a named constant

nddeluca (commented):
Looks good, should wait for second look from @rhuairahrighairidh before merge

migrate/v0_15/migrate_test.go
rhuairahrighairidh: seconded

rhuairahrighairidh (approved):
looks good, tested it locally with a genesis export from a week or so ago

PR #1030, opened 2021-10-22
kavadist module upgrade for v0.44

DracoLi:
Updates the kavadist module to 0.44.

A few things are missing ATM and will be added in:
- ~Tests that uses test_common (ignored for now since test_common is not migrated)~
- migrations (will look into this on a separate pr)
- ~mint.go uses the cdp module so a lot of the logic there need to be left out for now~
- ~Protobuf needs to be regenerated once we have that setup in the branch.~

Note: New tests (ie genesis & grpc queries) will be in a separate follow up PR.


x/kavadist/module.go
```diff
@@ -28,115 +31,135 @@ var (
...
-	return ModuleCdc.MustMarshalJSON(gs)
+// ExportGenesis returns kavadist module's exported genesis state as raw JSON bytes.
+func (am AppModule) ExportGenesis(ctx sdk.Context, cdc codec.JSONCodec) json.RawMessage {
+	genState := ExportGenesis(ctx, am.keeper)
+	return cdc.MustMarshalJSON(genState)
 }
 
-// BeginBlock module begin-block
+// ConsensusVersion implements ConsensusVersion.
+func (AppModule) ConsensusVersion() uint64 { return 2 }
```
drklee3, line 154: Just a minor thing, should return 1 instead of 2:
https://docs.cosmos.network/master/architecture/adr-041-in-place-store-migrations.html#module-consensusversion

```suggestion
func (AppModule) ConsensusVersion() uint64 { return 1 }
```

drklee3: Oops too late now but we can update it later

DracoLi: opps, yeah i'll need to update this. will create a PR

DracoLi: https://github.com/Kava-Labs/kava/pull/1039

PR #1031, opened 2021-10-22
Upgrade pricefeed to v44.3

drklee3:
Notable change in `Keeper.SetPrice()`, raw price store key is changed to `append(append(RawPriceFeedPrefix, []byte(marketId)...), []byte(oracleAddr)...)`.  Raw price is marshalled and set per oracle instead of all at once as a slice.


x/pricefeed/keeper/keeper.go
```diff
@@ -227,16 +207,27 @@ func (k Keeper) GetCurrentPrices(ctx sdk.Context) types.CurrentPrices {
...
-		return types.PostedPrices{}, nil
-	}
-	var prices types.PostedPrices
-	err := k.cdc.UnmarshalBinaryBare(bz, &prices)
-	if err != nil {
-		return types.PostedPrices{}, err
+func (k Keeper) GetRawPrices(ctx sdk.Context, marketId string) []types.PostedPrice {
+	var pps []types.PostedPrice
+	k.IterateRawPrices(ctx, func(pp types.PostedPrice) (stop bool) {
+		if pp.MarketId == marketId {
```
rhuairahrighairidh, line 213: Store iterators iterate through all keys in the store with a certain prefix. Since you have the raw price key as `RawPriceFeedPrefix<market id><oracle>` you could create an iterator for `RawPriceFeedPrefix<market id>` which would return the prices just for one market. It avoids unmarshalling and checking market ID on all prices.

Swap does this in `IterateDepositorSharesByOwner`.

Be careful with variable length keys though. Iterating over "btcb:usd" would also iterate over the prices under "btcb:usd:30" which is a different market. This can be fixed by converting market ids to a fixed length of bytes in the key.

I'll leave as a nit (your discretion on implementing or not) as I don't think the change gives us anything. Any performance increase is likely tiny given how few markets we have.

rhuairahrighairidh (changes requested):
Looks good, left some minor change requests.

rhuairahrighairidh (approved):
changes look good

proto/kava/pricefeed/v1beta1/pricefeed.proto
```diff
@@ -0,0 +1,37 @@
...
+  string          quote_asset = 3;
+  repeated string oracles     = 4;
+  bool            active      = 5;
+}
+
+// PostedPrice defines a price for market posted by a specific oracle.
+message PostedPrice {
+  option (gogoproto.goproto_stringer) = false;
+
+  string market_id      = 1 [(gogoproto.customname) = "MarketID"];
```
DracoLi, line 24: should we consider updating our code to use `Id` instead? We have a number of these vars and it'll be good to just generate them without overriding with `customname`
cc @rhuairahrighairidh @denalimarsh

drklee3: @nddeluca mentioned using it since it follows I believe Go initialism styling (?) and reduces the diff.  Although it sometimes leads to inconsistency since it can't be used for fields that are referenced in grpc-gateway query paths. So still not 100% sure on keeping it

PR #1035, opened 2021-10-27
Spike: simpler local chain setup

rhuairahrighairidh:
An experiment in simplying local chain setup

Running a local testing chain takes a lot of setup: https://docs.cosmos.network/master/run-node/run-node.html

In the past this overhead has lead to less testing, and made it difficult for non-backend developers to try out new features.

`kvtool` attempted to solve this. It generates a docker-compose file to run a kava chain and optionally other services.

However docker-compose adds complexity, docker builds are slow, and kvtool is not always up to date with kava.

This PR is a spike to simply setup:

An example home folder is added in contrib/devnet/home. Then adding `--home <path to kava>/contrib/devnet/home` to any `kava` command will use that configuration. It contains an example genesis file with funded accounts, and priv keys for sending txs.

- `kava --home <path to kava>/contrib/devnet/home start` will launch a chain (including rest and rpc servers)
- `kava --home <path to kava>/contrib/devnet/home tx  ...` will send a tx
- `kava --home <path to kava>/contrib/devnet/home query ...` will read data

Note, `kava start` won't work immediately. The data directory needs to be initialised first (if I commit initialised data files, I can't git ignore them). So I added a line to `make start`. There's probably better work arounds.


rhuairahrighairidh:
kvtool seems to be working for now

PR #1041, opened 2021-11-01
Upgrade bep3 module to v44

drklee3:


app/test_common.go
```diff
@@ -148,6 +151,22 @@ func (tApp TestApp) CheckBalance(t *testing.T, ctx sdk.Context, owner sdk.AccAdd
 	require.Equal(t, expectedCoins, coins)
 }
 
+// Create a new auth genesis state from some addresses and coins. The state is returned marshalled into a map.
+func (tApp TestApp) NewAuthGenState(ctx sdk.Context, cdc codec.JSONCodec, addresses []sdk.AccAddress, coins []sdk.Coins) GenesisState {
```
DracoLi, line 155: are we able to use the helper `NewFundedGenStateWithSameCoins` instead? it seems all usage cases of this utils here have the same coins too.

app/test_common.go
```diff
@@ -148,6 +151,22 @@ func (tApp TestApp) CheckBalance(t *testing.T, ctx sdk.Context, owner sdk.AccAdd
...
+	// Create GenAccounts
+	accounts := authtypes.GenesisAccounts{}
+	for i := range addresses {
+		accounts = append(accounts, authtypes.NewBaseAccount(addresses[i], nil, 0, 0))
+		// FundAccount moved to cosmos-sdk/x/bank/testutil in future release
+		if err := simapp.FundAccount(tApp.bankKeeper, ctx, addresses[i], coins[i]); err != nil {
```
DracoLi, line 161: we have a `FundAccount` util in test_commons too that can be used instead i believe

proto/kava/bep3/v1beta1/bep3.proto
```diff
@@ -0,0 +1,113 @@
...
+  bytes                    time_based_limit = 4
+      [(gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int", (gogoproto.nullable) = false];
+}
+
+// SwapStatus is the status of an AtomicSwap
+enum SwapStatus {
+  option (gogoproto.goproto_enum_prefix) = false;
+
+  // AtomicSwap is unspecified
+  SWAP_STATUS_UNSPECIFIED = 0x00;
```
DracoLi, line 52: is this the same as `SWAP_STATUS_UNSPECIFIED = 0`?

app/test_common.go
drklee3: Yep seems like it! nice catch

app/test_common.go
drklee3: Removed the function that uses it now so no longer needed, but thanks for pointing that out -- I'll use the one we have in test_common if I need it later

proto/kava/bep3/v1beta1/bep3.proto
drklee3: Yep I think I copied it from the old source -- will remove the hex `0x` prefix, doesn't seem necessary and makes it less clear.

x/bep3/handler.go
```diff
@@ -3,84 +3,33 @@ package bep3
 import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/kava-labs/kava/x/bep3/keeper"
+	"github.com/kava-labs/kava/x/bep3/types"
 )
 
 // NewHandler creates an sdk.Handler for all the bep3 type messages
-func NewHandler(k Keeper) sdk.Handler {
+func NewHandler(k keeper.Keeper) sdk.Handler {
```
DracoLi, line 11: i believe we can remove this in favour of the msg server?

x/bep3/types/codec.go
```diff
@@ -2,17 +2,43 @@ package types
...
 
-// RegisterCodec registers concrete types on amino
-func RegisterCodec(cdc *codec.Codec) {
-	cdc.RegisterConcrete(MsgCreateAtomicSwap{}, "bep3/MsgCreateAtomicSwap", nil)
-	cdc.RegisterConcrete(MsgRefundAtomicSwap{}, "bep3/MsgRefundAtomicSwap", nil)
-	cdc.RegisterConcrete(MsgClaimAtomicSwap{}, "bep3/MsgClaimAtomicSwap", nil)
+var (
+	amino = codec.NewLegacyAmino()
+
+	// ModuleCdc references the global x/gov module codec. Note, the codec should
```
DracoLi, line 32: nit: 
```suggestion
	// ModuleCdc references the global x/bep3 module codec. Note, the codec should
```

x/bep3/types/genesis.go
```diff
@@ -29,16 +21,16 @@ func NewGenesisState(params Params, swaps AtomicSwaps, supplies AssetSupplies, p
...
-		AtomicSwaps{},
-		AssetSupplies{},
+		[]AtomicSwap{},
+		[]AssetSupply{},
 		DefaultPreviousBlockTime,
 	)
 }
 
 // Equal checks whether two GenesisState structs are equivalent.
 func (gs GenesisState) Equal(gs2 GenesisState) bool {
```
DracoLi, line 31: i believe we can just remove the `Equal` func since they are not used anywhere. also allows us to not used amino here

x/bep3/keeper/grpc_query.go
```diff
@@ -0,0 +1,153 @@
+package keeper
```
rhuairahrighairidh, line 1: I'd add tests covering the query methods. Particularly around the filtered paginate logic.

x/bep3/client/cli/query.go
```diff
@@ -231,107 +234,108 @@ $ kvcli q bep3 swaps --page=2 --limit=100
 `,
```
rhuairahrighairidh, line 234: super nit: I'd move the Example text to the Example field on the command as the other commmands do. It's fine to span several lines.

It messes up autogenerated cli docs otherwise.

rhuairahrighairidh: Do the `status` and `direction` flags need to use the new string names like "SWAP_STATUS_OPEN"? If so we should update the docs here.

rhuairahrighairidh (changes requested):
Gave it a quick review, looks good, mainly just needs updating to the more recent updrade decisions.

Also the antehandler in app can be updated to include the authorized bep3 addresses again.

x/bep3/client/cli/query.go
drklee3: They use the original values when converting from strings, so it shouldn't require changes to clients

https://github.com/Kava-Labs/kava/blob/91cd58c6ee78214a4eca1866987c071f7b9955f2/x/bep3/types/swap.go#L126-L138

x/bep3/abci.go
```diff
@@ -4,12 +4,14 @@ import (
...
 
 // BeginBlocker on every block expires outdated atomic swaps and removes closed
 // swap from long term storage (default storage time of 1 week)
-func BeginBlocker(ctx sdk.Context, k Keeper) {
-	if ctx.BlockTime().After(ModulePermissionsUpgradeTime) {
+func BeginBlocker(ctx sdk.Context, k keeper.Keeper) {
+	if ctx.BlockTime().After(types.ModulePermissionsUpgradeTime) {
 		err := k.EnsureModuleAccountPermissions(ctx)
 		if err != nil {
 			k.Logger(ctx).Error(fmt.Sprintf("couldn't update module account permissions: %v", err))
```
rhuairahrighairidh, line 17: This was added as a softfork to fix an issue after a chain upgrade last year; the bep3 mod account didn't have the correct permissions.
It could be removed, though it would be nice to validate account permissions somewhere.

PR #1042, opened 2021-11-02
add kavadist to begin blocker

karzak:


DracoLi (approved):
oops, forgot to this in, good catch 👍

PR #1043, opened 2021-11-02
Committee module v44 upgrade - Types

DracoLi:
Committee module v44 upgrade. 

This PR only upgrades the types used in the committee module, with the exclusion of the `SubParamChangePermission` and `SimpleParamChangePermission` permissions as I'm planning to rewrite that for a more reflexible implementation in a separate PR.


x/committee/types/codec.go
```diff
@@ -42,26 +25,64 @@ func RegisterCodec(cdc *codec.Codec) {
...
+	registry.RegisterInterface(
+		"kava.committee.v1beta1.Committee",
+		(*Committee)(nil),
+		// TODO: Question: Might not need the base one since we just have token and member?
+		&BaseCommittee{},
+		&TokenCommittee{},
+		&MemberCommittee{},
+	)
+
+	// TODO: QUESTION: Is this needed?
```
DracoLi, line 49: thoughts?

x/committee/types/committee.go
```diff
@@ -1,185 +1,92 @@
...
 	BondDenom           = "ukava"
 )
 
-func init() {
-	// CommitteeChange/Delete proposals are registered on gov's ModuleCdc (see proposal.go).
-	// But since these proposals contain Committees, these types also need registering:
-	govtypes.ModuleCdc.RegisterInterface((*Committee)(nil), nil)
-	govtypes.RegisterProposalTypeCodec(BaseCommittee{}, "kava/BaseCommittee")
-	govtypes.RegisterProposalTypeCodec(MemberCommittee{}, "kava/MemberCommittee")
-	govtypes.RegisterProposalTypeCodec(TokenCommittee{}, "kava/TokenCommittee")
```
DracoLi, line 39: i believe these shouldn't be needed anymore since its for amino

x/committee/types/committee.go
```diff
@@ -211,23 +128,40 @@ func (c BaseCommittee) HasPermissionsFor(ctx sdk.Context, appCdc *codec.Codec, p
...
 }
 
 // GetProposalDuration is a getter for committee ProposalDuration
 func (c BaseCommittee) GetProposalDuration() time.Duration { return c.ProposalDuration }
 
 // SetProposalDuration is a setter for committee ProposalDuration
-func (c BaseCommittee) SetProposalDuration(proposalDuration time.Duration) BaseCommittee {
+func (c *BaseCommittee) SetProposalDuration(proposalDuration time.Duration) {
 	c.ProposalDuration = proposalDuration
-	return c
```
DracoLi, line 225: @denalimarsh do you remember why we don't set this directly on the committee? i noticed we are using these in simulations but the return values were never used.

x/committee/types/codec.go
```diff
@@ -42,26 +25,64 @@ func RegisterCodec(cdc *codec.Codec) {
...
+	registry.RegisterImplementations(
```
DracoLi, line 50: any ideas?

rhuairahrighairidh: Looks like it's not needed.
The `registry` is created in https://github.com/Kava-Labs/kava/blob/ccd7807b7ce759d69b8005d325fb25e00c0a315b/app/params/encoding.go#L22 and passed down to the modules.
Looking at the cosmos code for the registry `RegisterInterface` calls `RegisterImplementations` underneath. https://github.com/cosmos/cosmos-sdk/blob/b75c29fc15d3320ec0c7596dbd7c787c48dccad8/codec/types/interface_registry.go#L98-L117 so I think we don't need it.

x/committee/types/msg.go
```diff
@@ -100,58 +90,16 @@ func (vt *VoteType) Unmarshal(data []byte) error {
...
-		return err
-	}
-
-	*vt = bz2
-	return nil
-}
-
 // String implements the Stringer interface.
 func (vt VoteType) String() string {
+	// TODO: QUESTION: Can we just use the protobuf generated string implementation?
```
DracoLi, line 91: @denalimarsh thoughts?

proto/kava/committee/v1beta1/committee.proto
```diff
@@ -0,0 +1,64 @@
...
+  uint64         id          = 1 [(gogoproto.customname) = "ID"];
+  string         description = 2;
+  repeated bytes members     = 3 [
+    (cosmos_proto.scalar) = "cosmos.AddressBytes",
+    (gogoproto.casttype)  = "github.com/cosmos/cosmos-sdk/types.AccAddress"
+  ];
+  repeated google.protobuf.Any permissions = 4 [(cosmos_proto.accepts_interface) = "Permission"];
+
+  // Smallest percentage that must vote for a proposal to pass
+  bytes vote_threshold = 5
```
rhuairahrighairidh, line 27: nit: Going off what Nick said, I think it's slightly easier for clients if fields tagged with customtype for sdk Int and Dec are strings.
Looking at [what gogo does with the tag](https://github.com/gogo/protobuf/blob/master/custom_types.md), it calls the `(Un)Marshal` methods on the Int and Dec types and drops those bytes directly into the encoded proto message. The Int and Dec methods use a text marshalling func, so the bytes will always correspond to a string.
It makes it slightly clearer to clients that the `vote_threshold` field is a decimal marshalled to string, rather than some custom binary marshalling.

x/committee/types/proposal.go
```diff
@@ -110,30 +69,29 @@ func (ccp CommitteeChangeProposal) ProposalRoute() string { return RouterKey }
 // ProposalType returns the type of the proposal.
 func (ccp CommitteeChangeProposal) ProposalType() string { return ProposalTypeCommitteeChange }
 
+func (ccp CommitteeChangeProposal) GetNewCommittee() Committee {
+	committee, err := UnpackCommittee(ccp.NewCommittee)
+	if err != nil {
+		panic(err)
+	}
+	return committee
+}
```
rhuairahrighairidh, line 78: If I'm understanding correctly, the `CommitteeChangeProposal` will need a `UnpackInterfaces` method as well. Otherwise the cached value won't be populated when unmarshalled.

rhuairahrighairidh (approved):
Looks good. I'd like to understand some of the interface marshalling edge cases a bit more (particularly how proto and amino interact), but I can't think of any problems so far. Interested to see if we run into any problems in the keeper.

x/committee/types/proposal.go
rhuairahrighairidh: maybe add a `var _ UnpackInterfacesMessage = CommitteeChangeProposal{}` type check?

proto/kava/committee/v1beta1/committee.proto
DracoLi: make sense, I can update this

x/committee/types/proposal.go
DracoLi: yep, that is correct. i'll add this in

PR #1044, opened 2021-11-02
Add additional pricefeed keeper tests

drklee3:
Covers the following cases and more

* When an oracle posts a price, it should be accepted
* When a non-oracle posts a price, it should error and not be accepted
* Given some stored valid and expired raw prices, when the current price is set (k.SetCurrentPrices), it should set the price ignoring expired raw prices.
* grpc querier tests


x/pricefeed/keeper/grpc_query_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+	_, err = suite.keeper.Price(sdk.WrapSDKContext(suite.ctx), &types.QueryPriceRequest{MarketId: "invalid"})
+	suite.Equal("rpc error: code = InvalidArgument desc = invalid market ID", err.Error())
+
+	prices, err := suite.keeper.Prices(sdk.WrapSDKContext(suite.ctx), &types.QueryPricesRequest{})
+	suite.NoError(err)
+
+	// Includes previous same price
+	suite.Contains(prices.Prices, expectedPrice)
+}
```
rhuairahrighairidh, line 92: nit: Looks like this test tests 4 different scenarios. In general I'd recommend splitting up tests like these to make it clear.
It makes it easier to for future devs to see what behvaiour is tested. And large tests tend to get larger and harder to read over time.

Could do separate tests like `TestGrpcPrice_NoPriceSet()`, or use separate `suite.Run(...`s within a single test method, or do a table driven test like in `kavadist/grpc_query_test.go`

x/pricefeed/keeper/grpc_query_test.go
```diff
@@ -0,0 +1,201 @@
...
+		{MarketID: "tstusd", BaseAsset: "tst", QuoteAsset: "usd", Oracles: oracles, Active: true},
+	})
+	suite.keeper.SetParams(suite.ctx, params)
+
+	res, err = suite.keeper.Oracles(sdk.WrapSDKContext(suite.ctx), &types.QueryOraclesRequest{MarketId: "tstusd"})
+	suite.NoError(err)
+	suite.ElementsMatch(res.Oracles, oracles)
+
+	_, err = suite.keeper.Oracles(sdk.WrapSDKContext(suite.ctx), &types.QueryOraclesRequest{MarketId: "invalid"})
+	suite.Equal("rpc error: code = InvalidArgument desc = invalid market ID", err.Error())
```
rhuairahrighairidh, line 160: nit: Not finding a market feels more like a `NotFound` error code than an `InvalidArgument`

rhuairahrighairidh (changes requested):
I added a commit to switch the store key encoding to use length prefixing. A bit marginal, but saves having to pick a suffix that will never collide with market ids. I added a confluence doc with an overview of diferent encoding schemes.

A few requests:

Do you want to follow the msgServer pattern for the grpc querier? ie attach the grpc methods to a "queryServer" struct rather than the keeper. It helps avoid name collisions and separates concerns.

There's an old `doc.go` that can be removed.

There's a few places where the 'marketID' function argument names have been capitalised to 'MarketID'. Could you revert them to reduce the changes from master. It reduces chances of small bugs getting in.

Could you add a genesis test to check genesis state is added to the store correctly, and then exported correctly? Noticed this wasn't covered by the exisitng test.

rhuairahrighairidh (approved):
nice, looks good

Issue #1046, opened 2021-11-04
Fail local testnet docker-compose up


soaryong-c:
in kava directory
```
> build-docker-local-kava
> localnet-start
```

failed by 
- The binary kvd cannot be found. Please add the binary to the shared folder. Please use the BINARY environment variable if the name of the binary is not 'kvd' E.g.: -e BINARY=kvd_my_test_version

So, i change volumes in docker-compose file to  **./build:/kvd:Z**

PR #1047, opened 2021-11-04
Fix fail local testnet docker-compose up

soaryong-c:
Fail local testnet docker-compose up #1046


rhuairahrighairidh (approved):
Looks good, thanks for the fix!

PR #1048, opened 2021-11-04
Add Swagger For GRPC Queries

nddeluca:
- Adds make proto-swagger-gen command
- Adds a clients/docs with a vendored swagger-ui 4.0.0
- Reorders make proto-all to generate go code first (helpful linting is failing).

To view:
```
cd client/docs/swagger-ui && python3 -m http.server
```

The generated swagger still needs some improvements and should be combined with the old swagger.yml with depreciation notices.


rhuairahrighairidh (approved):
lgtm
I'd add a note to confluence about adding module's docs. Looks like we need to add a few lines to `config.json` for each module.

PR #1049, opened 2021-11-04
Add Custom Address Type For Swap State Records

nddeluca:


nddeluca:
Closing in favor of casttype

PR #1051, opened 2021-11-05
Upgrade CDP types to v44

drklee3:


proto/kava/cdp/v1beta1/genesis.proto
```diff
@@ -0,0 +1,146 @@
...
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+    (gogoproto.nullable)   = false
+  ];
+  string liquidation_penalty = 7 [
+    (cosmos_proto.scalar)  = "cosmos.Dec",
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
+    (gogoproto.nullable)   = false
+  ];
+  // No single byte type, use a uint32
+  uint32 prefix                   = 8;
```
drklee3, line 107: There isn't a single `byte` type for protobuf so not sure which one would be preferred:

`uint32` and casting between byte/uint32 in places like keeper vs `bytes` and just using the first item.  Leaning towards just `uint32` as it represents a single item vs possible multiple ones with `bytes` type which seems less clear and could maybe be prone to indexing errs.

x/cdp/types/cdp.go
```diff
@@ -204,29 +185,41 @@ func (augcdps AugmentedCDPs) String() string {
...
 }
 
-// TotalCDPPrincipal is a total principal of a given collateral type
-type TotalCDPPrincipal struct {
-	CollateralType string   `json:"collateral_type" yaml:"collateral_type"` // string representing the unique collateral type of the CDP
-	Amount         sdk.Coin `json:"amount" yaml:"amount"`                   // Amount of principal stored in this CDP
+// TotalPrincipals a collection of TotalPrincipal objects
+type TotalPrincipals []TotalPrincipal
+
+func (t *TotalPrincipal) String() string {
```
drklee3, line 191: Do we want to keep implement the stringer methods ourselves? Or use the proto generated ones

DracoLi: i vote for the generated stringer methods to reduce code & its usually pretty good i think. Not sure if any downsides.

rhuairahrighairidh (changes requested):

proto/kava/cdp/v1beta1/query.proto
```diff
@@ -0,0 +1,174 @@
...
+// QueryCdpsRequest is the params for a filtered CDP query, the request type for the Query/Cdps RPC method.
+message QueryCdpsRequest {
+  string collateral_type = 1;
+  string owner           = 2 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+  uint64 id              = 3 [(gogoproto.customname) = "ID"];
+  string ratio           = 4 [
+    (cosmos_proto.scalar)  = "cosmos.Dec",
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
+    (gogoproto.nullable)   = false
+  ];
```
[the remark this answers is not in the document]
DracoLi: what do you mean by
> The sdk.Dec type marshals to a string, but without the decimal point
Do you mean it doesn't parse the decimal part when included in a grpc request param?

proto/kava/cdp/v1beta1/genesis.proto
drklee3: 👍  I'll use the collateral type, not too clear why there's the separate prefix originally -- is there a need to keep store key/values short?

x/cdp/client/rest/rest.go
```diff
@@ -1,74 +1,74 @@
 package rest
 
-import (
-	"github.com/gorilla/mux"
+// import (
```
DracoLi, line 3: to avoid ignoring all files, I tried just creating an empty go.mod file under these folders that not migrated. I think that worked out well and also preserves the original diff once these are migrated. no need to change here i think but something to consider for the future.

x/cdp/types/msg.go
```diff
@@ -45,9 +37,15 @@ func (msg MsgCreateCDP) Type() string { return "create_cdp" }
 
 // ValidateBasic does a simple validation check that doesn't require access to any other information.
 func (msg MsgCreateCDP) ValidateBasic() error {
-	if msg.Sender.Empty() {
+	if msg.Sender == "" {
```
DracoLi, line 40: fyi, i found that `AccAddressFromBech32` does the empty check so we should be okay to no include this check

DracoLi (approved):
LGTM, just some minor informational comments

proto/kava/cdp/v1beta1/query.proto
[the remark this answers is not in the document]
rhuairahrighairidh: However, since changing direction to not include grpc queriers in the upgrade, it might be easier to just remove them here and save this problem for the future.

proto/kava/cdp/v1beta1/genesis.proto
rhuairahrighairidh: No strong need to keep them short. But they do need to allow iteration. There's a few keeper methods that iterator over cdps with the same collateral-type.

I wrote up some notes on different encoding options here: https://kava-labs.atlassian.net/wiki/spaces/ENG/pages/1030488065/Store+Key+Encoding.

rhuairahrighairidh (approved):
looks good

PR #1053, opened 2021-11-08
Update v44 pricefeed addresses to use bytes

drklee3:


proto/kava/pricefeed/v1beta1/pricefeed.proto
```diff
@@ -12,19 +13,25 @@ option (gogoproto.verbose_equal_all) = true;
...
-  string          market_id   = 1 [(gogoproto.customname) = "MarketID"];
-  string          base_asset  = 2;
-  string          quote_asset = 3;
-  repeated string oracles     = 4;
-  bool            active      = 5;
+  string         market_id   = 1 [(gogoproto.customname) = "MarketID"];
+  string         base_asset  = 2;
+  string         quote_asset = 3;
+  repeated bytes oracles     = 4 [
+    (cosmos_proto.scalar) = "cosmos.AddressBytes",
```
rhuairahrighairidh, line 20: do you know of any docs on what the `(cosmos_proto.scalar)` does?

drklee3: Doesn't seem to be used yet, maybe later in cosmos v0.45.  Just added to be consistent with swap module

https://github.com/cosmos/cosmos-sdk/issues/9694

rhuairahrighairidh (commented):
Changes look good. Couple of additional requests:
- Do you want to separate the proto query types from the store types. See #1050 and the swap query types.
- There's a TODO in `module.go` for registering invariants. The pricefeed module doesn't have any invariants so we can remove the TODO.
- Minor code quality improvements: `CalculateMedianPrice` and `calculateMeanPrice` are just math functions, so they don't need access to the ctx, so that argument could be removed. Also `calculateMeanPrice` will panic if anyone ever passes in a slice of more than 2 prices. I'd probably swap the slice arg for two individual prices args.

rhuairahrighairidh (approved):
LGTM

PR #1054, opened 2021-11-08
[R4R] feat: add upgrade module

karzak:
Adds the cosmos-sdk `upgrade` module to app. 

Note - doesn't implement a custom handler like gaia is doing here: https://github.com/cosmos/gaia/blob/277a25b8940f0cfb1dc2d05fe11dc6e646840988/app/app.go#L567 - this seems to be built for automating the migration process and flagging which modules are new (ie - don't need a migration). 

Tested with kvtool and was able to submit a software upgrade proposal and have it halt the chain using regular `gov` module. When `committee` is ready, we should add a test there that confirms the safety committee can also halt the chain.


nddeluca:
@karzak updated to reuse the existing skip upgarde height and @rhuairahrighairidh I moved defaultHomeDir references back to the var set in app.go.

nddeluca (approved):

PR #1055, opened 2021-11-09
Use stdout for cmd outputs; Output snake_case json for status command

nddeluca:
Set the command output for Print/Println, so stdout is used.  In addition, vendor the status command and add json tags for snake_case output.


rhuairahrighairidh (approved):
looks good

PR #1056, opened 2021-11-10
Upgrade v44: Committee keeper

DracoLi:
Areas covered in this PR:
- Upgrade committee keepers to v44
- New grpc queries for the committee module
- Added the committee module to the app
- A large amount of tests for the keeper and the module are upgraded

Few things are not upgrade:
- Not all tests are upgraded. All tests that dependent on either another kava module, the upgrade module, or because it uses the sub params permissions are not upgraded.
- committee client is not migrated yet
- no tests for gprc queries, I plan to cover these in a separate PR since we are aiming to get the module in first.


x/committee/testutil/types.go
```diff
@@ -1,11 +1,24 @@
...
+	"github.com/stretchr/testify/assert"
 )
 
 // Avoid cluttering test cases with long function names
 func I(in int64) sdk.Int                    { return sdk.NewInt(in) }
 func D(str string) sdk.Dec                  { return sdk.MustNewDecFromStr(str) }
 func C(denom string, amount int64) sdk.Coin { return sdk.NewInt64Coin(denom, amount) }
 func Cs(coins ...sdk.Coin) sdk.Coins        { return sdk.NewCoins(coins...) }
+
+func AssertProtoMessageJSON(t *testing.T, cdc codec.Codec, expected proto.Message, actual proto.Message) {
```
DracoLi, line 18: is there a better place to put assertion helpers like this that might be used by other modules as well?

x/committee/proposal_handler_test.go
```diff
@@ -144,32 +144,32 @@ func (suite *ProposalHandlerTestSuite) TestProposalHandler_ChangeCommittee() {
...
 
 			// Check
 			if tc.expectPass {
 				suite.NoError(err)
 				// check committee is accurate
-				actualCom, found := suite.keeper.GetCommittee(suite.ctx, tc.proposal.NewCommittee.GetID())
+				actualCom, found := suite.keeper.GetCommittee(suite.ctx, tc.proposal.GetNewCommittee().GetID())
 				suite.True(found)
-				suite.Equal(tc.proposal.NewCommittee, actualCom)
+				testutil.AssertProtoMessageJSON(suite.T(), suite.app.AppCodec(), tc.proposal.GetNewCommittee(), actualCom)
```
DracoLi, line 163: note: had to do this because it seems after proto.message marshal and unmarshals, the empty slices will get converted to nil slices, which will fail the equal assertion. So here I updated this to compare their json data instead.

x/committee/keeper/querier.go
```diff
@@ -166,39 +165,18 @@ func queryVote(ctx sdk.Context, path []string, req abci.RequestQuery, keeper Kee
...
 	if !found {
-		return nil, sdkerrors.Wrapf(types.ErrUnknownProposal, "%d", params.ProposalID)
-	}
-
-	committee, found := keeper.GetCommittee(ctx, proposal.CommitteeID)
-	if !found {
-		return nil, sdkerrors.Wrapf(types.ErrUnknownCommittee, "%d", proposal.CommitteeID)
-	}
-
-	var pollingStatus types.ProposalPollingStatus
```
DracoLi, line 186: `ProposalPollingStatus` has the exact same fields as `QueryTallyResponse` so I believe this legacy rest endpoint should return the same data

x/committee/testutil/types.go
rhuairahrighairidh: Not currently but it would be nice to start one. I had a quick look at other big go repos and couldn't see any standard people use. Maybe `/testutil` or `/test`?

x/committee/keeper/grpc_query.go
```diff
@@ -0,0 +1,183 @@
...
+func (k QueryHandler) Votes(c context.Context, req *types.QueryVotesRequest) (*types.QueryVotesResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "empty request")
+	}
+
+	ctx := sdk.UnwrapSDKContext(c)
+	votes := k.GetVotesByProposal(ctx, req.ProposalID)
+	votesResp := types.QueryVotesResponse{
+		Votes: make([]types.QueryVoteResponse, len(votes)),
+	}
```
rhuairahrighairidh, line 112: For the token committee the number of votes could get pretty large. So it's probably worth implementing pagination on this new votes endpoint.

I can't imagine a scenario where we'd have enough committees to be worth paginating. There's quite a few proposals, but this endpoint only returns the active ones, which there probably won't be many of. So only paginating votes seems ok.

Most of the cosmos modules use a Paginate function in their grpc servers.

x/committee/keeper/grpc_query.go
```diff
@@ -0,0 +1,183 @@
...
+	"google.golang.org/grpc/status"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/kava-labs/kava/x/committee/types"
+)
+
+var _ types.QueryServer = QueryHandler{}
+
+type QueryHandler struct {
```
DracoLi, line 16: testing with creating a new handler for grpc queries instead of using the keeper here to avoid potential func conflicts. thoughts? @nddeluca @rhuairahrighairidh

rhuairahrighairidh: +1 to avoiding name conflicts. Could also not embed the keeper to avoid exporting Keeper methods to users of the `QueryHandler`

rhuairahrighairidh: +1 to avoiding name conflicts. Could also not embed the keeper to avoid exporting Keeper methods to uses of the `QueryHandler`

rhuairahrighairidh (commented):
looks good, left a few thoughts

x/committee/keeper/grpc_query.go
DracoLi: agreed, I can add this in

DracoLi:
@rhuairahrighairidh thx for the review. feedback should all be addressed here https://github.com/Kava-Labs/kava/pull/1056/commits/9ad70deec2836c73fcd7ab7ac5cfd49bcc501df7. A few are for grpc queries which I know we can ignore for now but I included them anyways since i was pretty close to finishing them up.

x/committee/types/msg.go
```diff
@@ -72,7 +75,15 @@ func (msg MsgSubmitProposal) GetSignBytes() []byte {
 
 // GetSigners returns the addresses of signers that must sign.
 func (msg MsgSubmitProposal) GetSigners() []sdk.AccAddress {
-	return []sdk.AccAddress{msg.Proposer}
+	return []sdk.AccAddress{msg.GetProsper()}
+}
+
+func (msg MsgSubmitProposal) GetProsper() sdk.AccAddress {
```
rhuairahrighairidh, line 81: nit: `GetProposer()`?

x/committee/proposal_handler_test.go
rhuairahrighairidh: good to know, looks like go's nil and empty slices get encoded to the same thing in the protobuf wire format

rhuairahrighairidh (approved):
looks good

PR #1045, opened 2021-11-02
Spike: generated cli docs

rhuairahrighairidh:
Demo of generated cli reference docs.

View by running `make docs-develop` then going to http://localhost:8080

Generation code is super messy, don't merge it.

PR #1057, opened 2021-11-10
added CLI docs

samcode206:
adding documentation for kvcli


docs/.vuepress/config.js
```diff
@@ -1,141 +1,251 @@
 module.exports = {
-  theme: "cosmos",
-  title: "Kava Documentation",
-  themeConfig: {
-    logo: {
-      src: "/logo.svg",
+  "theme": "cosmos",
```
rhuairahrighairidh, line 2: Any reason for adding the quotes?
In general it's easier to see code changes when not combined with formatting changes.

rhuairahrighairidh:
I'll leave to others to review the docs content and structure. But going for manual docs vs autogenerated looks good. I'll close spike #1045

docs/.vuepress/config.js
samcode206: @rhuairahrighairidh I was using yaml to write the changes and taking the json output which is why you see the double quotes, I can revert it back to no quotes!

PR #1059, opened 2021-11-11
feat: add support for ibc (ics-20)

karzak:
Adds ibc ([ics-20](https://github.com/cosmos/ibc/tree/master/spec/app/ics-020-fungible-token-transfer)) and dependencies.


rhuairahrighairidh (approved):
I gave this a look over and compared against gaia v6. Looks ok, but can't really tell without digging in more. Approving for now to unblock others.

PR #1060, opened 2021-11-11
Upgrade v44: Committee client + ParamsChangePermission 

DracoLi:
This PR upgrades committee client to v44 and also adds `ParamsChangePermission`. 

`ParamsChangePermission` aims to replace the original `SimpleParamChangePermission` and `SubParamChangePermission` in terms of functionality. 
Currently, `ParamsChangePermission` only supports the functionalities of `SimpleParamChangePermission` but I'll be adding support for sub param change and required sub param attrs as a followup PR as that will take some time.


rhuairahrighairidh (approved):
Haven't manually tested, but looks ok. It's probably worth revisiting once cdp is merged so we can get the proposal tests running.

PR #1061, opened 2021-11-16
fix: update circulating supply

karzak:
Updates circulating supply (`kvcli q vesting circulating-supply` `/vesting/circulatingsupply`)  to accurately reflect inflationary KAVA that has been minted for staking rewards/SAFU fund. The current endpoint reflects a max circulating supply of ~110M KAVA, when in fact the max circulating supply of KAVA is ~150M. 

The circulating supply calculation for KAVA is updated to:
1. Start with the total supply of KAVA
2. Subtract any remaining vesting tokens for team/investors (~10M tokens as of Nov 5th, 2021). 

After this update is applied, the circulating supply will go from `100625000 KAVA` to `~140968892 KAVA`.


denalimarsh (approved):
Changes LGTM

PR #1062, opened 2021-11-16
Update CDP type keys to use collateralType strings

drklee3:
Missed in #1051

- Updates the expected keepers to match the pricefeed changes for addresses.
- Uses string `collateralType` for keys over `denomByte` as the collateral byte prefix is removed in favor of using `collateralType` directly.


DracoLi (approved):
LGTM

nddeluca (approved):

PR #1067, opened 2021-11-17
Upgrade auth genesis builder

rhuairahrighairidh:
Small PR to resinstate the auth genesis test helper.
It's used heavily in incentive tests.

I ran into import cycle by moving it to it's own package, so just left it where it was. It could be maybe moved along with the TestApp in the future.

Originally the builder was meant to be immutable and easily copyable. Turns out slices can subtly break this so I switched the builder to use pointer receivers to try and make it clear it shouldn't be copied. Shouldn't require any changes to incentive tests.


nddeluca (approved):

nddeluca:
Integrated with master -- there was a new test helper that I updated to also use the builder.

PR #1070, opened 2021-11-19
Add Legacy POST /txs Endpoint

nddeluca:
Allows clients to broadcast amino json encoded transactions.

Since sigverify in v44 requires a sequence in the signatures, and legacy transactions do not include this information, we fetch the current on-chain sequence for each signer when converting.


rhuairahrighairidh (approved):
LGTM

PR #1071, opened 2021-11-19
Use multi stage Docker file

drklee3:
Reduces Docker image size from `2.5GB` to `73.5MB`.

`/go/pkg/mod` alone was taking up 1.2GB which is only needed for the build step.


nddeluca:
Nice, lgtm.  I tested build locally, will merge and push a new image for upgrade-v44

PR #1073, opened 2021-11-19
Upgrade hard keeper to v44

drklee3:
Does not contain grpc_query, will be added in a separate PR

Will retarget to `upgrade-v44` when hard types merged.


x/hard/keeper/keeper.go
```diff
@@ -1,32 +1,33 @@
...
 type Keeper struct {
 	key             sdk.StoreKey
-	cdc             *codec.Codec
-	paramSubspace   subspace.Subspace
+	cdc             codec.Codec
+	paramSubspace   paramtypes.Subspace
 	accountKeeper   types.AccountKeeper
-	supplyKeeper    types.SupplyKeeper
+	bankKeeper      types.BankKeeper
 	stakingKeeper   types.StakingKeeper
```
rhuairahrighairidh, line 22: nit: looks like staking keeper isn't used anymore and could be removed. The method `BondDenom` isn't used.

x/hard/keeper/keeper.go
```diff
@@ -153,64 +154,76 @@ func (k Keeper) IterateBorrows(ctx sdk.Context, cb func(borrow types.Borrow) (st
...
-		store.Set([]byte{}, []byte{})
+		store.Set(types.BorrowedCoinsPrefix, []byte{})
 	} else {
-		bz := k.cdc.MustMarshalBinaryBare(borrowedCoins)
-		store.Set([]byte{}, bz)
+		bz, err := borrowedCoins.MarshalJSON()
+		if err != nil {
+			panic(err)
+		}
+		store.Set(types.BorrowedCoinsPrefix, bz)
```
rhuairahrighairidh, line 163: I think it'd be slightly better to use protobuf here for coins marshalling.

The standard library json (un)marshalling is usually slow compared to gogo protobuf. And keeping store marshalling all proto makes it easier to reason about as a whole.

You could create a wrapper type around `sdk.Coins` a bit like `sdk.DecProto` or `cdptypes.OwnerCDPIndex` and use it only for (un)marshalling to the store.

x/hard/keeper/repay.go
```diff
@@ -112,7 +112,7 @@ func (k Keeper) ValidateRepay(ctx sdk.Context, sender, owner sdk.AccAddress, coi
 	}
 
 	senderAcc := k.accountKeeper.GetAccount(ctx, sender)
-	senderCoins := senderAcc.SpendableCoins(ctx.BlockTime())
+	senderCoins := k.bankKeeper.SpendableCoins(ctx, senderAcc.GetAddress())
```
rhuairahrighairidh, line 115: nit: looks like `senderAcc` is never used so the `GetAccount` line could be removed.

x/hard/keeper/timelock.go
```diff
@@ -1,23 +1,23 @@
 package keeper
```
rhuairahrighairidh, line 1: This file is leftover from when hard used to contain the functionality now in incentive. Looks like the methods aren't used anywhere, so it could be deleted.

x/hard/keeper/liquidation.go
```diff
@@ -91,9 +91,10 @@ func (k Keeper) SeizeDeposits(ctx sdk.Context, keeper sdk.AccAddress, deposit ty
 		}
 	}
 	if !keeperRewardCoins.Empty() {
-		k.DecrementSuppliedCoins(ctx, keeperRewardCoins)
-		err := k.supplyKeeper.SendCoinsFromModuleToAccount(ctx, types.ModuleAccountName, keeper, keeperRewardCoins)
-		if err != nil {
+		if err := k.DecrementSuppliedCoins(ctx, keeperRewardCoins); err != nil {
```
rhuairahrighairidh, line 94: nice catch

rhuairahrighairidh (changes requested):
Left a couple of requests, but overall looks good.
I think we made the right call by keeping the number of keeper changes low. It would have taken me significantly longer to review if there was a lot of business logic changes.

rhuairahrighairidh (approved):
Changes look good

nddeluca (approved):
Made some minor updates and merged in the latest upgrade-v44 changes.  Looks good

PR #1077, opened 2021-11-23
Upgrade incentive types to v44

denalimarsh:
Updates incentive types folder to v44


x/incentive/types/msg.go
```diff
@@ -39,35 +33,33 @@ func (msg MsgClaimUSDXMintingReward) Type() string { return "claim_usdx_minting_
...
+	sender, err := sdk.AccAddressFromBech32(msg.Sender)
+	if err != nil {
+		panic(err)
+	}
+	return []sdk.AccAddress{sender}
 }
 
 // NewMsgClaimUSDXMintingRewardVVesting returns a new MsgClaimUSDXMintingReward.
-func NewMsgClaimUSDXMintingRewardVVesting(sender, receiver sdk.AccAddress, multiplierName string) MsgClaimUSDXMintingRewardVVesting {
+func NewMsgClaimUSDXMintingRewardVVesting(sender, receiver string, multiplierName string) MsgClaimUSDXMintingRewardVVesting {
```
drklee3, line 62: For the `NewMsgxxx()` functions I've been keeping the address args as `sdk.AccAddress` then converting it to string in the returned struct, eg

```go
return MsgClaimUSDXMintingRewardVVesting{
		Sender:         sender.String(),
                Receiver:       receiver.String(),
		MultiplierName: multiplierName,
	}
}
```

Depends on where the functions are used, but seems like they're mostly when we already have an address so it might be easier for use that way and prevents possible need to revalidate here

rhuairahrighairidh (approved):
Looks good. One slight problem is that MultiplierName is now an int32, so amino encodes it as a number rather than string. So incentive/params will return numbers for multiplier names, breaking backwards compatiability. I think the easiest fix might be to simplify and remove the enum altogether, but it will be easier to address in future PRs.

PR #1081, opened 2021-11-24
Fix committee proposal query

DracoLi:
This fixes the issue where proposal content was not properly unmarshalled in the legacy proposal query.
Also registers the `CommunityPoolMultiSpendProposal` from kavadist as we were missing it.

### DIFF
--- a/x/committee/client/common/query.go
+++ b/x/committee/client/common/query.go
@@ -73,10 +73,10 @@ func QueryProposalByID(cliCtx client.Context, cdc *codec.LegacyAmino, queryRoute
 	res, height, err := cliCtx.QueryWithData(fmt.Sprintf("custom/%s/%s", queryRoute, types.QueryProposal), bz)
 
 	if err == nil {
-		var proposal *types.Proposal
+		var proposal types.Proposal
 		cdc.MustUnmarshalJSON(res, &proposal)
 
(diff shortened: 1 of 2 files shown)


nddeluca (approved):

x/committee/client/common/query.go
```diff
@@ -73,10 +73,10 @@ func QueryProposalByID(cliCtx client.Context, cdc *codec.LegacyAmino, queryRoute
 	res, height, err := cliCtx.QueryWithData(fmt.Sprintf("custom/%s/%s", queryRoute, types.QueryProposal), bz)
 
 	if err == nil {
-		var proposal *types.Proposal
+		var proposal types.Proposal
 		cdc.MustUnmarshalJSON(res, &proposal)
 
-		return proposal, height, nil
+		return &proposal, height, nil
```
thomasmost, line 79: @DracoLi So in this case the value of `proposal` is still undefined(?) but the **pointer** has been updated to by `MustUnmarshalJSON` to have data?

DracoLi: I believe we were passing the address of a pointer to unmarshall before. That didn't work well with proto's unpack interface, which resulted in the content (which requires unpacking) to not be processed, thus resulting in a null value.

PR #1082, opened 2021-11-25
Fix legacy amino msg encoding for committee

nddeluca:
The LegacyAmino codec now requires messages to be registered as references.


DracoLi (approved):
👍

rhuairahrighairidh:
Not sure if this change needs to be applied to all amino registered types. We currently do a mix of pointers and non-pointers. But I'll merge this to unblock front end work.

PR #1083, opened 2021-11-29
Update cli to use the correct client ctx fetcher

DracoLi:
I noticed a number of client cli queries are using `GetClientTxContext` as opposed to `GetClientQueryContext`. This PR should fix all existing occurrences of this issue.


drklee3 (approved):
LGTM, seems like they're all right now.

Another minor thing if we want to update it here too:

PR #1084, opened 2021-11-29
fix: register legacy amino cdp messages with pointer

karzak:


rhuairahrighairidh (approved):
LGTM, though I haven't tested it

PR #1086, opened 2021-12-02
Upgrade v44: Auction v016 migration

DracoLi:
- Add v16 auction migrations
- Add kava x/committee and x/auction migrations to the v16 migration process. 
- Added x/committee and x/auction sample v15 genesis data to the v15-v16 migration test.


rhuairahrighairidh (approved):
looks good. I double checked the legacy aution types against master, and walked through the migration and tests.

PR #1090, opened 2021-12-03
Upgrade v44: bep3 v016 migration

DracoLi:
- Added v16 x/bep3 migrations
- Added kava x/bep3 to the v16 migration process.
- Added x/bep3 sample v15 genesis data to the v15-v16 migration test.


rhuairahrighairidh (approved):
looks good, I double checked the legacy types against v0.15.2, and walked through the migration and test.
In past migration's we've had to reset the block height in the swaps, but this time we could avoid it if we don't reset the block height for the new chain. I'll add a card to track.

PR #1091, opened 2021-12-03
Upgrade v44: v016 migrations for swap, kavadist, and validator-vesting

DracoLi:
- Add v015 migration for swap, kavadist and validator-vesting.
- x/swap does not seem to change the json from my tests.
- x/kavadist seems not to change expect I noticed a tiny precision difference in the marshalled `time.Time` data (see comment).
- x/validator-vesting genesis data is deleted.


migrate/v0_16/testdata/genesis-v16.json
```diff
@@ -793,6 +793,117 @@
...
+            "start": "2022-06-01T14:00:00Z"
+          },
+          {
+            "end": "2024-06-01T14:00:00Z",
+            "inflation": "1.000000000782997609",
+            "start": "2023-06-01T14:00:00Z"
+          }
+        ]
+      },
+      "previous_block_time": "2021-11-05T21:13:12.856088470Z"
```
DracoLi, line 822: v15 is `2021-11-05T21:13:12.85608847Z`
There's an extra decimal point here.

rhuairahrighairidh: I think I found the reason. In the proto json spec: https://developers.google.com/protocol-buffers/docs/proto3#json it says the number of digits after the decimal point should always be 0, 3, 6, or 9. So I imagine gogoproto is adding an extra 0 to comply.

DracoLi: oh nice, great find!

rhuairahrighairidh (approved):
looks good,
only nit would be to add a validator vesting state to the `migrate/v0_16/testdata/genesis-v15.json` test

DracoLi:
> only nit would be to add a validator vesting state to the migrate/v0_16/testdata/genesis-v15.json test
@rhuairahrighairidh yep! I noticed that too and included this in the next migration PR https://github.com/Kava-Labs/kava/pull/1093/files#diff-d5ea94e46366f57f0f013be36892231d83807e11dbc0e215f33b382002518379R4

rhuairahrighairidh:
ah nice

PR #1093, opened 2021-12-07
Upgrade v44: Migration for issuance, pricefeed, & cdp

DracoLi:
- issuance
- pricefeed
- cdp


rhuairahrighairidh (approved):
looks good, I double checked the types and walked through the migrations and tests

PR #1095, opened 2021-12-08
Upgrade incentive client

DracoLi:
Upgrades incentive client to v44.

Note:
- grpc queries are not created for the incentive module since it adds a lot more extra complexity right now for little benefit. Therefore, the cli is upgraded to use the legacy querier.

### DIFF
--- a/x/incentive/client/cli/query.go
+++ b/x/incentive/client/cli/query.go
@@ -7,9 +7,8 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 
-	"github.com/cosmos/cosmos-sdk/client/context"
+	"github.com/cosmos/cosmos-sdk/client"
 	"github.com/cosmos/cosmos-sdk/client/flags"
-	"github.com/cosmos/cosmos-sdk/codec"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/version"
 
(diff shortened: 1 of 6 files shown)


x/incentive/client/cli/query.go
```diff
@@ -63,13 +68,16 @@ func queryRewardsCmd(queryRoute string, cdc *codec.Codec) *cobra.Command {
...
 		Args: cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			cliCtx := context.NewCLIContext().WithCodec(cdc)
+			cliCtx, err := client.GetClientQueryContext(cmd)
+			if err != nil {
+				return err
+			}
 
 			page := viper.GetInt(flags.FlagPage)
 			limit := viper.GetInt(flags.FlagLimit)
```
drklee3, line 83: I think `viper.GetInt(...)` can be replaced with `cmd.Flags().GetInt(...)` to be consistent with the other module clients

x/incentive/client/cli/query.go
```diff
@@ -86,62 +94,62 @@ func queryRewardsCmd(queryRoute string, cdc *codec.Codec) *cobra.Command {
...
 					return err
 				}
-				swapClaims, err := executeSwapRewardsQuery(queryRoute, cdc, cliCtx, params)
+				swapClaims, err := executeSwapRewardsQuery(cliCtx, params)
 				if err != nil {
 					return err
 				}
 				if len(hardClaims) > 0 {
-					cliCtx.PrintOutput(hardClaims)
+					cliCtx.PrintObjectLegacy(hardClaims)
```
drklee3, line 143: Would return for these since `PrintObjectLegacy` also return errors

```suggestion
					return cliCtx.PrintObjectLegacy(hardClaims)
```

x/incentive/client/cli/tx.go
```diff
@@ -22,292 +18,151 @@ const multiplierFlag = "multiplier"
...
+			msg := types.NewMsgClaimHardReward(sender.String(), selections)
 			if err := msg.ValidateBasic(); err != nil {
 				return err
 			}
-			return utils.GenerateOrBroadcastMsgs(cliCtx, txBldr, []sdk.Msg{msg})
+			return tx.GenerateOrBroadcastTxCLI(cliCtx, cmd.Flags(), &msg)
 		},
 	}
 	cmd.Flags().StringToStringVarP(&denomsToClaim, multiplierFlag, multiplierFlagShort, nil, "specify the denoms to claim, each with a multiplier lockup")
 	cmd.MarkFlagRequired(multiplierFlag)
```
drklee3, line 100: `cmd.MarkFlagRequired()` returns an error, seems like cosmos just discards the error without handling it, but maybe we can just panic for it

drklee3 (commented):
Some minor things and unhandled errors found with `golangci-lint`

x/incentive/client/cli/query.go
drklee3: Misunderstood this, we can just handle handle error but not return since we're printing multiple things here.

x/incentive/client/cli/tx.go
DracoLi: good idea, i'll update to panic here

x/incentive/client/cli/query.go
DracoLi: right, i'll update to return only if we get an error

drklee3 (approved):

nddeluca (approved):

PR #1096, opened 2021-12-08
feat: add export to app

karzak:


app/export.go
```diff
@@ -60,25 +67,25 @@ func (app *App) prepForZeroHeightGenesis(ctx sdk.Context, jailWhiteList []string
...
-	app.stakingKeeper.IterateValidators(ctx, func(_ int64, val staking.ValidatorI) (stop bool) {
-		accumCommission := app.distrKeeper.GetValidatorAccumulatedCommission(ctx, val.GetOperator())
-		if accumCommission.IsZero() {
-			return false
-		}
-		_, err := app.distrKeeper.WithdrawValidatorCommission(ctx, val.GetOperator())
-		if err != nil {
-			log.Fatal(err)
-		}
+	// withdraw all validator commission
```
nddeluca, line 70: Duplicate comment here

app/export.go
```diff
@@ -135,8 +149,8 @@ func (app *App) prepForZeroHeightGenesis(ctx sdk.Context, jailWhiteList []string
 
 	// Iterate through validators by power descending, reset bond heights, and
 	// update bond intra-tx counters.
-	store := ctx.KVStore(app.keys[staking.StoreKey])
-	iter := sdk.KVStoreReversePrefixIterator(store, staking.ValidatorsKey)
+	store := ctx.KVStore(app.keys[stakingtypes.StoreKey])
+	iter := sdk.KVStoreReversePrefixIterator(store, stakingtypes.ValidatorsKey)
 	counter := int16(0)
 
 	for ; iter.Valid(); iter.Next() {
```
nddeluca, line 156: The next line `addr := sdk.ValAddress(iter.Key()[1:])` should be changed to `		addr := sdk.ValAddress(stakingtypes.AddressFromValidatorsKey(iter.Key()))`.

The keys are now `prefix + addrLength + addressbytes`, and the helper `AddressFromValidatorsKey` accounts for this by using `key[2:]` instead of `key[1:]`

cmd/kava/cmd/app.go
```diff
@@ -110,9 +111,22 @@ func (ac appCreator) appExport(
...
 
-	panic("TODO") // TODO
+	homePath, ok := appOpts.Get(flags.FlagHome).(string)
+	if !ok || homePath == "" {
+		return servertypes.ExportedApp{}, errors.New("application home not set")
+	}
+
+	var tempApp *app.App
+	if height != -1 {
+		tempApp = app.NewApp(logger, db, homePath, traceStore, ac.encodingConfig, app.Options{SkipLoadLatest: true, InvariantCheckPeriod: uint(1)})
```
nddeluca, line 121: Do we intentionally hardcode `InvariantCheckPeriod` here?  We could use the `server.FlagInvCheckPeriod` flag -- unsure it it matters for `LoadHeight` & `ExportAppStateAndValidators`

nddeluca (changes requested):
We need to update the iterator key -> ValAddress logic, otherwise lgtm

nddeluca (approved):

rhuairahrighairidh (commented):
It would be good to have a test for `prepForZeroHeightGenesis`. I'm not entirley sure what it does, and if it does it correctly. It feels a bit forgotten about in gaia, but still makes complex state changes.
Alternatively, now that chains can start from any height, it could maybe be removed.

PR #1097, opened 2021-12-09
Remove incentive multiplier name enum

rhuairahrighairidh:
Switching the MultiplierName to an enum caused a change in the legacy querier - multiplier names changed from strings to numbers. This caused some problems for the front end.

This PR addresses this by removing the enum altogether.
Available multipliers are specified in the params. Any claim must reference one to claim successfully. So all they need is some identifier.
Limiting this identifier to one of small,medium,large doesn't add much given the available multipliers are bounded by those listed in params.


x/hard/types/tx.pb.go
```diff
@@ -562,7 +562,7 @@ type MsgClient interface {
 	Borrow(ctx context.Context, in *MsgBorrow, opts ...grpc.CallOption) (*MsgBorrowResponse, error)
 	// Repay defines a method for repaying funds borrowed from hard liquidity pool.
 	Repay(ctx context.Context, in *MsgRepay, opts ...grpc.CallOption) (*MsgRepayResponse, error)
-	// Liquidate defines a method for attempgint to liquidate a borrower that is over their loan-to-value.
+	// Liquidate defines a method for attempting to liquidate a borrower that is over their loan-to-value.
```
rhuairahrighairidh, line 565: I think proto gen hadn't been run after this change.

DracoLi (approved):
LGTM

denalimarsh (approved):
Changes LGTM

PR #1101, opened 2021-12-09
Remove validator vesting keeper

DracoLi:
Removed the validator-vesting keeper and updated the legacy querier to use the bank keeper directly.

Also moved the querier out since it no longer needs to be in the keeper folder.


rhuairahrighairidh (approved):
I didn't do any manual testing, but looks good (modulo merge conflict)

PR #1103, opened 2021-12-14
Fix bep3 msg validation

rhuairahrighairidh:
Bep3 msg validation did not check the byte length of the random number hash. Instead it was inspecting the encoded hex string.

This was missed eariler as the msg tests were not running.


denalimarsh (approved):
tACK

PR #1105, opened 2021-12-16
Resolve some v44 migration todos

rhuairahrighairidh:
Small PR to tidy up some of the TODOs hanging around app since the upgrade to v0.44.

I added the line to set upgrade version map. Not strictly needed, we only need the upgrade module to halt the chain when a software upgrade proposal passes, which it does. This version map just populates an upgrade grpc endpoint.


nddeluca (approved):
looks good 👍

PR #1110, opened 2021-12-20
Fix auctions query pagination

drklee3:
Includes pagination request in the query.  Also updates command flags to use cobra instead of viper and simplifies parameter variables.  The few uses of `strings.TrimSpace` were removed since the only time where there would be leading/trailing spaces is when flags include quotes ie `--denom "    ukava    "`


DracoLi (approved):
👍

PR #1111, opened 2021-12-20
Filter auctions query with request parameters

drklee3:
Updates auction grpc query to use `FilteredPaginate` and filter based on passed request parameters.

### DIFF
--- a/x/auction/keeper/grpc_query_test.go
+++ b/x/auction/keeper/grpc_query_test.go
@@ -1,0 +1,137 @@
+package keeper_test
+
+import (
+	"testing"
+	"time"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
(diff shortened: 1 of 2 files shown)


x/auction/keeper/grpc_query_test.go
```diff
@@ -0,0 +1,137 @@
...
+		{
+			"owner and denom",
+			types.QueryAuctionsRequest{
+				Owner: addrs[0].String(),
+				Denom: "hard",
+			},
+			auctions[3:4],
+		},
+		{
+			"owner, denom, type, phaseå",
```
DracoLi, line 107:
```suggestion
			"owner, denom, type, phase",
```

DracoLi (approved):
LGTM

PR #1112, opened 2021-12-21
Add support for subparam rules for ParamsChangePermission

DracoLi:
This PR updates `ParamsChangePermission` to match the functionality of `SubParamChangePermission` that allows changes to specific sub-key within module parameter keys, 

In order to create subparam rules in a flexible manner, the idea here is to store rules for sub params in a single object that can work across modules. In this case, the `SubparamRequirement` contains the requirements for subparam key and value so that it can target specific records in subparams and the `allowed_subparam_attr_changes` slice in it contains the rules for which properties can be updated.

Note, since we are no longer unmarsharling the params into its target param struct, we do lose some type validation there. But the proposal handler from cosmos should do some type validation right after our logic so it should be okay.

--

Remaining TODOs:
- [x] A couple of test cases are not written out

### DIFF
--- a/x/committee/types/permissions.go
+++ b/x/committee/types/permissions.go
@@ -1,9 +1,11 @@
 package types
 
 import (
+	"encoding/json"
 	fmt "fmt"
+	"reflect"
+	"strings"
 
-	"github.com/cosmos/cosmos-sdk/codec"
 	"github.com/cosmos/cosmos-sdk/codec/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
@@ -24,7 +26,7 @@ func init() {
 
 // Permission is anything with a method that validates whether a proposal is allowed by it or not.
 type Permission interface {
-	Allows(sdk.Context, codec.Codec, ParamKeeper, PubProposal) bool
+	Allows(sdk.Context, ParamKeeper, PubProposal) bool
 }
 
 func PackPermissions(permissions []Permission) ([]*types.Any, error) {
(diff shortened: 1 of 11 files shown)


x/committee/types/permissions.go
```diff
@@ -122,14 +123,149 @@ func (changes AllowedParamsChanges) FilterByParamChange(paramChange paramspropos
...
+		}
+
+		var currentValue MultiSubparamChanges
+		subspace, found := pk.GetSubspace(paramsChange.Subspace)
+		if !found {
+			return false
+		}
+		raw := subspace.GetRaw(ctx, []byte(paramsChange.Key))
+		if err := json.Unmarshal(raw, &currentValue); err != nil {
+			panic(err)
```
DracoLi, line 248: panic here to match `subspace.Get` behaviour. We should never have bad data stored in subspace so I think panic here make sense.

x/committee/types/permissions.go
```diff
@@ -122,14 +123,154 @@ func (changes AllowedParamsChanges) FilterByParamChange(paramChange paramspropos
...
+				data2, err := json.Marshal(incoming[k])
+				if err != nil {
+					return false
+				}
+				if string(data) != string(data2) {
+					return false
+				}
+			} else if v != incoming[k] {
+				return false
+			}
```
rhuairahrighairidh, line 212: Couple of thoughts:
- if json marshalling the map isn't deterministic, the string comparison could fail incorrectly. Looks like the current implementation is deterministic, but doesn't guarantee it will always be. https://stackoverflow.com/questions/44755089/does-serialized-content-strictly-follow-the-order-in-definition-use-encoding-json-package
- pricefeed Markets contains a field `Oracles` which is a slice of addresses. This might cause the `else if v != incoming[k]` comparison to fail as slices can't be compared.

DracoLi: :+1: updated to use `reflect.DeepEqual` after our discussion. https://github.com/Kava-Labs/kava/pull/1112/commits/327afa4c87e90b0cf03dc3a804e2ca590b5b2b26

rhuairahrighairidh (approved):
Looks good. I added a warning about determinism in `validateParamChangesAreAllowed` to future devs.
All our state machine code needs to be deterministic, otherwise different validators can get different results for a block, which causes chain halts.

PR #1113, opened 2021-12-21
Remove expired vesting periods from accounts in migrations

drklee3:
`ResetPeriodicVestingAccount` function copied from v15 migration.

#### Testing

Updates the tests to use a fixed test genesis time to allow for the actual migration genesis time to be changed without breaking tests. 

Test data json file changes:

* The existing periodic vesting account with expired periods is migrated to a base account in v16.
* Adds another periodic vesting account with 2 vesting periods, one which is expired and one pending.  The resulting migrated account only has 1 period and the start time updated.


migrate/v0_16/migrate.go
```diff
@@ -13,8 +13,8 @@ import (
 )
 
 var (
-	GenesisTime = time.Date(2021, 11, 30, 15, 0, 0, 0, time.UTC)
-	ChainID     = "kava-8"
+	GenesisTime = time.Date(2022, 1, 19, 16, 0, 0, 0, time.UTC)
```
nddeluca, line 16: Genesis time is updated to launch date

migrate/v0_16/migrate.go
```diff
@@ -13,8 +13,8 @@ import (
...
+	ChainID     = "kava-9"
```
nddeluca, line 17: Chain id is updated to `kava-9`

migrate/v0_16/migrate_test.go
```diff
@@ -43,3 +56,115 @@ func TestMigrateFull(t *testing.T) {
 		tApp.InitializeFromGenesisStatesWithTimeAndChainID(newGenDoc.GenesisTime, newGenDoc.ChainID, app.GenesisState(appMap))
 	})
 }
+
+func TestAccountBalances(t *testing.T) {
+	t.Skip() // avoid committing test data
```
nddeluca, line 61: Skipped in integration, passing locally against exported state of block 1627000

migrate/v0_16/legacyaccounts/accounts.go
```diff
@@ -0,0 +1,485 @@
+package legacyaccounts
```
nddeluca, line 1: This is used in the migrate test instead of the legacy types from the cosmos-sdk.  The cosmos-sdk types do not implement Spendable, etc methods.

nddeluca:
Added an e2e test and account migration looks good

nddeluca (approved):

PR #1115, opened 2021-12-28
Migrate v15 SubParamPermission to v16 ParamsChangePermission

DracoLi:
Migrate v15 `SubParamPermission` with requires for different module params to the v16 format. All functionalities and intentions of the original restrictions should be preserved.

### DIFF
--- a/x/committee/legacy/v0_16/migrate.go
+++ b/x/committee/legacy/v0_16/migrate.go
@@ -2,6 +2,8 @@ package v0_16
 
 import (
 	"fmt"
+	"reflect"
+	"sort"
 
 	proto "github.com/gogo/protobuf/proto"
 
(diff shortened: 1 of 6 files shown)


x/committee/legacy/v0_16/migrate.go
```diff
@@ -15,12 +17,176 @@ import (
...
+	key      []byte
+	subspace string
+}
+
+// migrateSubParamPermissions converts v15 SubParamChangePermissions to v16 ParamsChangePermission
+func migrateSubParamPermissions(permission v015committee.SubParamChangePermission) *v016committee.ParamsChangePermission {
+	changes := v016committee.AllowedParamsChanges{}
+
+	// migrate allowed params
+	pairsToAvoid := []subspaceKeyPair{
```
DracoLi, line 68: avoiding a couple params here since we are handling them later and the new subparam permission does not need them to be recorded twice.

rhuairahrighairidh (approved):
looks good

PR #1118, opened 2021-12-30
Make QueryAuctionsResponse `auctions` field plural

drklee3:
This makes it consistent with other requests that respond with multiple values.

This is a breaking change, but as far as I know only auction bot currently requires and uses this query response.


nddeluca (approved):
👍

go.mod
```diff
@@ -13,7 +13,6 @@ require (
 	github.com/grpc-ecosystem/grpc-gateway v1.16.0
 	github.com/spf13/cast v1.3.1
 	github.com/spf13/cobra v1.2.1
-	github.com/spf13/viper v1.8.1
```
nddeluca, line 16: This is from me removing the `kvcli`, `kvd` directories and not cleaning of up mods

PR #1119, opened 2022-01-03
fix: add genesis migrations for ibc and related modules

karzak:
Defines default genesis state for `capability`, `ibc`, and `transfer` modules - migrations we're failing because these were not defined.

Commented out the `TestMigrateGenesisDoc` as it seems brittle and doesn't  explicitly test what we care about - that the resulting genesis state is valid and can be used to start a new chain. Happy to update/fix it if we think it's worth keeping around.

### DIFF
--- a/migrate/v0_16/testdata/genesis-v16.json
+++ b/migrate/v0_16/testdata/genesis-v16.json
@@ -1479,6 +1479,51 @@
         }
       ]
     },
+    "capability": {
+      "index": "1",
+      "owners": []
+    },
+    "transfer": {
+      "denom_traces": [],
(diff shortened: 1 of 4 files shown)


migrate/v0_16/testdata/genesis-v16.json
```diff
@@ -1479,6 +1479,51 @@
         }
       ]
     },
+    "capability": {
```
nddeluca, line 1482: @karzak fixed the e2e genesis test by adding the default gen state for the ibc modules

nddeluca (approved):

PR #1122, opened 2022-01-06
use string over bytes and keep hex encoding in messages

nddeluca:


rhuairahrighairidh (approved):
LGTM

denalimarsh (approved):
Confirming that this fixes the type issues.

PR #1123, opened 2022-01-06
Use Coin type uniformly in tx.proto files

nddeluca:
Non-breaking change -- updates the proto definition for issuance messages.


drklee3 (approved):
There also seems like some extra `(gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Coin"` options set in bep3 Coin fields that can be removed to make them consistent too

nddeluca:
@drklee3 updated those other places

rhuairahrighairidh (approved):
LGTM

PR #1131, opened 2022-01-11
Add IBC markets to pricefeed migration

drklee3:


rhuairahrighairidh (changes requested):
Revisting this, we'll need the new price denoms added to the stability committee permissions. Otherwise the difference in pricefeed markets and committee permissions will dissallow any changes to the markets by the stability committee.

#1132 has an example of adding permissions for the hard params

x/pricefeed/legacy/v0_16/migrate.go
```diff
@@ -16,9 +17,99 @@ func migrateParams(params v015pricefeed.Params) v016pricefeed.Params {
...
+	newMarkets := []v016pricefeed.Market{
+		{
+			MarketID:   "atom:usd",
+			BaseAsset:  "atom",
+			QuoteAsset: "usd",
+			Oracles:    oracles,
+			Active:     true,
+		},
+		{
+			MarketID:   "atom:usd:30",
```
DracoLi, line 42: been wanting to ask this, what are the `:30` markets?

DracoLi (commented):
looks good, I think just need to update the committee permissions like @rhuairahrighairidh mentioned.

x/pricefeed/legacy/v0_16/migrate.go
drklee3: TWAP, time weighted average price over 30 minutes

The ones without the `:30` suffix are spot prices

DracoLi: awesome, thx!

migrate/v0_16/kava.go
```diff
@@ -59,8 +59,11 @@ func migrateKavaAppState(appState genutiltypes.AppMap, clientCtx client.Context)
 		var genState v015committee.GenesisState
 		v15Codec.MustUnmarshalJSON(appState[v015committee.ModuleName], &genState)
 
+		var pricefeedGenState v015pricefeed.GenesisState
+		v15Codec.MustUnmarshalJSON(appState[v015pricefeed.ModuleName], &pricefeedGenState)
+
 		// replace migrated genstate with previous genstate
-		appState[v015committee.ModuleName] = v16Codec.MustMarshalJSON(v016committee.Migrate(genState))
+		appState[v015committee.ModuleName] = v16Codec.MustMarshalJSON(v016committee.Migrate(genState, pricefeedGenState))
```
drklee3, line 66: Any recommendations for handling missing states? Seems like missing module states just results in silently skipping over those modules, not sure if we would want to do the same here -- would a panic or warning be ideal here if pricefeed genesis state is missing? 

Also doesn't seem too ideal to have to pass the pricefeed genesis state from here through all of the following functions, but doesn't seem like there's another option to get pricefeed markets other than hard coding them in.

DracoLi: since committee migration depends on the price feed one, i agree that it might be good to panic here if it is missing.
I think passing genState to the migrate function should be okay? Cosmos migration also do this for x/bank and I don't see any other good options.

DracoLi (approved):

rhuairahrighairidh (approved):
looks good

PR #1132, opened 2022-01-12
Update migrations to include uatom IBC denom money market

denalimarsh:
Update hard and committee module migrations to include uatom's IBC denom. I set the initial limit is set to ~$1m.

### DIFF
--- a/x/committee/legacy/v0_16/migrate.go
+++ b/x/committee/legacy/v0_16/migrate.go
@@ -184,7 +184,7 @@ func migrateSubParamPermissions(
 			requirements = append(requirements, requirement)
 		}
 
-		if isStabilityCommittee {
+		if committeeID == KavaStabilityCommitteeID {
 			// Add permissions for existing pricefeed markets that are missing in allowed_markets
 		outer:
 			for _, oldPricefeedMarket := range oldPricefeedState.Params.Markets {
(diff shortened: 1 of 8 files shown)


denalimarsh:
I also made the assumption that the pricefeed would be `atom:usd:30`.

rhuairahrighairidh (changes requested):
Hard params look good. I think the hard governance committee will also need the new permission. Otherwise the committee won't be able to vote through changes to hard perissions. @karzak to confirm.

karzak:
@rhuairahrighairidh is correct- the hard committee needs Atom money market permission added. It also needs swp money market permission (forgot to do this in an early pr).

denalimarsh:
Added `swp` and the uatom ibc denom to the Hard Governance committee. My solution targets committee by ID, so I updated the ID of `Kava God Committee (testing only)` so the Hard Governance committee ID matches up with its mainnet ID.

x/committee/legacy/v0_16/migrate.go
```diff
@@ -107,7 +108,7 @@ func migrateSubParamPermissions(permission v015committee.SubParamChangePermissio
 		}
 
 		// add new requirement for stability committee
-		if isStabilityCommittee {
+		if committeeID == 1 {
```
DracoLi, line 111: thoughts on creating some vars in this file for these magic numbers so it's clear the targeting committees?

denalimarsh: Added in https://github.com/Kava-Labs/kava/pull/1132/commits/6f1b340758b6c52d4d5de874abe6d6812c141522

DracoLi (approved):
LGTM

karzak (approved):
LGTM - manually ran mainnet migrations and confirmed that Hard + Committee state looks good.

rhuairahrighairidh (approved):
change look good, I'll fix the merge conflicts

PR #1133, opened 2022-01-13
Create rollback.md

JusTTeam:
I tried to reproduce the previous instruction. Please check if everything is correct.


karzak (approved):
Thanks for catching this!

PR #1135, opened 2022-01-14
Reset bep3 swaps for zero height in migration

rhuairahrighairidh:
This PR copies over behaviour we've had in previous migrations.
Bep3 swaps use block height to track timeouts so they need updating as the block height is reset to zero in kava-9.

- Completed Swaps
  - closed height is set to 1 so they are only stored in long term storage for 1 + default storage time (in block times)
- Expired Swaps
  - expired height is reset to 1. Not strictly needed as keeper code operates off status
- Open swaps
  - Incoming swaps are expired, causing bnb side to expire
  - Outoing swaps have their expiry extended by 48hrs after chain launch to leave time for claiming in case bnb side was claimed during launch downtime.

I also reinstated the app export test. I'd originally planned to reset swaps in the `prepForZeroHeight` function but stopped as it would've taken longer. Swaps shouldn't need reseting after kava-9 as future launches don't need the block height reset.


karzak (approved):
LGTM

PR #1136, opened 2022-01-14
Validate module genesis states after migration

drklee3:
This catches some possible invalid module genesis states, but does not catch everything -- invalid migration state could still be emitted as there's still some validation in module `InitGenesis` functions.


DracoLi (commented):
nice, great additions

drklee3:
closing since chain is already migrated so this shouldn't be necessary anymore

PR #1137, opened 2022-01-14
Update comos-sdk to forked version with upstream patches

nddeluca:
- Note: This patch is state breaking due to the block gas fix.


karzak (approved):
LGTM

PR #1138, opened 2022-01-16
IBC enhancements

faddat:
These changes:

* use the latest ibc-go v1.x
* update to go 1.17 
* update to tendermint 0.34.15
* update to tm-db 0.6.6
* use a patched rocksdb library from the cosmos org because the earlier one fails

All together these changes will improve the performance of Kava when IBC relayers run kava nodes.


faddat:
Also, if there's time you may want to consider using ibc v2

nddeluca (approved):

PR #1147, opened 2022-01-19
IBC v3 upgrade

DracoLi:
Upgrades IBC to v3.

This is merging into the `upgrade-evm` branch because it is required by ethermint v0.10.0

https://github.com/cosmos/ibc-go/blob/release/v3.0.x/docs/migrations/v1-to-v2.md
https://github.com/cosmos/ibc-go/blob/release/v3.0.x/docs/migrations/v2-to-v3.md


app/app.go
```diff
@@ -444,16 +444,18 @@ func NewApp(
 		keys[ibctransfertypes.StoreKey],
 		ibctransferSubspace,
 		app.ibcKeeper.ChannelKeeper,
```
karzak, line 446: Do you understand why this is passed twice (albeit once as an "Ics4Wrapper")- seems weird?

karzak (approved):
One clarifying question - not blocking. Looked over the app changes pretty closely and they seemed correct.

Makefile
```diff
@@ -209,7 +209,6 @@ proto-update-deps:
 	rsync -r --chmod 644 --include "*.proto" --include='*/' --exclude='*' $(TENDERMINT_PATH)/proto third_party
 	rsync -r --chmod 644 --include "*.proto" --include='*/' --exclude='*' $(COSMOS_SDK_PATH)/proto third_party
 	rsync -r --chmod 644 --include "*.proto" --include='*/' --exclude='*' $(IBC_GO_PATH)/proto third_party
-	rsync -r --chmod 644 --include "*.proto" --include='*/' --exclude='*' $(IBC_GO_PATH)/third_party/proto/confio third_party/proto
```
nddeluca, line 212: Actually, think the proofs proto here just moved -- may just want to add a copy for https://github.com/cosmos/ibc-go/blob/main/third_party/proto/proofs.proto?  Believe IBC protos still depend on it

third_party/proto/ibc/lightclients/tendermint/v1/tendermint.proto
```diff
@@ -2,11 +2,11 @@ syntax = "proto3";
 
 package ibc.lightclients.tendermint.v1;
 
-option go_package = "github.com/cosmos/ibc-go/modules/light-clients/07-tendermint/types";
+option go_package = "github.com/cosmos/ibc-go/v3/modules/light-clients/07-tendermint/types";
 
 import "tendermint/types/validator.proto";
 import "tendermint/types/types.proto";
-import "confio/proofs.proto";
+import "proofs.proto";
```
nddeluca, line 9: re othter comment; import path changed, so still need to incldue proofs in third_party

app/app.go
DracoLi: yeah great point, this is from the migration docs for v2->v3.
> The transferkeeper.NewKeeper(...) now takes in an ICS4Wrapper. The ICS4Wrapper should be the IBC Channel Keeper unless ICS 20 is being connected to a middleware application.
So unless we are using a middleware here, we should just pass the channel keeper here.

Makefile
DracoLi: Oh yeah nice find, i totally missed this. Yeah I'll make sure to include this file.

DracoLi: https://github.com/Kava-Labs/kava/pull/1147/commits/a32e7af74efa5827b527a21b4bb9ef73d53a1025

PR #1148, opened 2022-01-20
WIP Add ethermint

DracoLi:
- [ ] Use ethermint's `InitCmd` to allow JSON-RPC to be started
- [ ] Add evm module too app.go
- [ ] Use ethermint's custom app template & app config since new properties are added
- [ ] Update init-new-chain to be able to start the chain with evm support
- [ ] Chain id updated to match evm specifications
- [ ] Cosmos SDK updated to 0.45.0 because ethermint requires it. This requires adding all modules to the begin and end blockers.
- [ ] Update `proto-update-deps` to include ethermint's proto files and add the new proto files
- [ ] Updated ante handler to be able to handle both cosmos and eth transactions


app/app.go
```diff
@@ -350,7 +370,7 @@ func NewApp(
 		appCodec,
 		keys[authtypes.StoreKey],
 		authSubspace,
-		authtypes.ProtoBaseAccount,
+		ethermint.ProtoAccount,
```
DracoLi, line 373: this changes all BaseAccounts to an EthAccount, which is unlikely what we will do. However, keeping it this for now to simplify dev before we figure out how to properly handle this.

DracoLi:
just going to merge this since new work is being developed on this PR

PR #1151, opened 2022-01-25
Fix breaking evm tests

DracoLi:
Fixes existing tests with ethermint integration

### DIFF
--- a/x/swap/testutil/suite.go
+++ b/x/swap/testutil/suite.go
@@ -20,6 +20,7 @@ import (
 	abci "github.com/tendermint/tendermint/abci/types"
 	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
 	tmtime "github.com/tendermint/tendermint/types/time"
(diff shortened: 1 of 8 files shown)


x/swap/testutil/suite.go
```diff
@@ -103,15 +104,15 @@ func (suite *Suite) NewAccountFromAddr(addr sdk.AccAddress, balance sdk.Coins) a
 // CreateVestingAccount creates a new vesting account from the provided balance and vesting balance
 func (suite *Suite) CreateVestingAccount(initialBalance sdk.Coins, vestingBalance sdk.Coins) authtypes.AccountI {
 	acc := suite.CreateAccount(initialBalance)
-	bacc := acc.(*authtypes.BaseAccount)
+	bacc := acc.(*ethermint.EthAccount)
```
DracoLi, line 107: unlikely we will change the base account to EthAccount but it's easier to keep it for now during development

PR #1153, opened 2022-01-25
Fix app module ordering for multi-spend proposal handler

nddeluca:
Instantiates the kavadistkeeper before it's reference in the gov handler.

This fixes an issue where the struct has default values, which leaves the distrkeeper nil.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -421,6 +421,16 @@ func NewApp(
 		scopedIBCKeeper,
 	)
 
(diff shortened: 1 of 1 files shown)


DracoLi (approved):

rhuairahrighairidh (approved):
looks good.
Haven't tested, but it looks like adding the commitee route on gov (ln441) has the same problem. I added a card to track it

PR #1154, opened 2022-01-28
Wrap bank keeper for EVM to convert decimals

drklee3:
Sending transactions seem to work properly with `ukava` as the evm_denom now.

### DIFF
--- a/app/evm_keeper.go
+++ b/app/evm_keeper.go
@@ -1,8 +1,6 @@
 package app
 
 import (
-	"fmt"
-
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	bankkeeper "github.com/cosmos/cosmos-sdk/x/bank/keeper"
 	evmtypes "github.com/tharsis/ethermint/x/evm/types"
@@ -40,7 +38,6 @@ func (bk EVMBankKeeper) SendCoinsFromModuleToAccount(ctx sdk.Context, senderModu
 
 func (bk EVMBankKeeper) SendCoinsFromAccountToModule(ctx sdk.Context, senderAddr sdk.AccAddress, recipientModule string, amt sdk.Coins) error {
 	actualAmt := convertCoinsFromEvm(amt)
-	fmt.Printf("SendCoinsFromAccountToModule: original %v, actualAmt: %v", amt, actualAmt)
 
 	return bk.bankKeeper.SendCoinsFromAccountToModule(ctx, senderAddr, recipientModule, actualAmt)
 }
--- a/app/evm_keeper.go
+++ b/app/evm_keeper.go
@@ -7,6 +7,13 @@ import (
 	evmtypes "github.com/tharsis/ethermint/x/evm/types"
 )
 
+// ExpectedEVMBankKeeper the expected interface for the EVM bank keeper wrapper.
+type ExpectedEVMBankKeeper interface {
+	evmtypes.BankKeeper
+
+	SpendableCoins(ctx sdk.Context, addr sdk.AccAddress) sdk.Coins
+}
+
 // Convertion between native gas KAVA (6) to EVM (18).
 // 12 decimal difference, so 1_000_000_000_000.
 var conversionMultiplier = sdk.NewInt(1_000_000_000_000)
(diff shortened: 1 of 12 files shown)


app/evm_keeper.go
```diff
@@ -0,0 +1,85 @@
...
+
+// EVMBankKeeper is a wrapper for bank keeper that converts between EVM and native decimals
+type EVMBankKeeper struct {
+	bankKeeper bankkeeper.Keeper
+}
+
+var _ evmtypes.BankKeeper = (*EVMBankKeeper)(nil)
+
+// NewEVMBankKeeper returns a wrapped bank keeper that converts between EVM and native decimals
+func NewEVMBankKeeper(bk bankkeeper.Keeper) EVMBankKeeper {
```
nddeluca, line 21: It'd be good to unit test all of the functions here -- table tests with specific examples would be great for a wrapper like this.

app/evm_keeper.go
```diff
@@ -0,0 +1,85 @@
...
+	coin.Amount = coin.Amount.Mul(conversionMultiplier)
+
+	return coin
+}
+
+// convertCoinFromEvm converts an EVM sdk.Coin with 18 decimals back to native 6 decimal sdk.Coin
+func convertCoinFromEvm(coin sdk.Coin) sdk.Coin {
+	coin.Amount = coin.Amount.Quo(conversionMultiplier)
+
+	if coin.IsZero() {
```
nddeluca, line 68: I *think* we actually want to truncate to zero here and return an error from any method converting from evm 18 decimals to 6 decimals.

Even though it's small, rounding up ends up operating on more coins than specified -- which should be avoided.  An explicit error is probably best?

nddeluca: The edge cases also make great test examples -- most conversion code is much easier to understand from the tests in addition to having the implementation.

app/evm_keeper.go
```diff
@@ -0,0 +1,85 @@
...
+		coin.Amount = sdk.OneInt()
+	}
+
+	return coin
+}
+
+// convertCoinFromEvm converts EVM sdk.Coins with 18 decimals back to native 6 decimal sdk.Coins
+func convertCoinsFromEvm(coins sdk.Coins) sdk.Coins {
+	// Use a new sdk.Coins, do NOT modify in place as it is used afterwards
```
nddeluca, line 77: This type of behavior tends to get lost over time, if it's possible lets add a test case that ensures we do not modify in place.

app/evm_keeper.go
```diff
@@ -0,0 +1,85 @@
...
+	evmtypes "github.com/tharsis/ethermint/x/evm/types"
+)
+
+// Convertion between native gas KAVA (6) to EVM (18).
+// 12 decimal difference, so 1_000_000_000_000.
+var conversionMultiplier = sdk.NewInt(1_000_000_000_000)
+
+// EVMBankKeeper is a wrapper for bank keeper that converts between EVM and native decimals
+type EVMBankKeeper struct {
+	bankKeeper bankkeeper.Keeper
```
nddeluca, line 15: Since we intend to only use with the EVM -- we could also depend the abstract https://github.com/tharsis/ethermint/blob/main/x/evm/types/interfaces.go#L29.  Also, it's usually preferable to depend on interfaces where we can (and if it doesn't introduce unnecessary complexity).

app/evm_keeper.go
```diff
@@ -0,0 +1,85 @@
...
+
+// convertCoinToEvm converts a sdk.Coin with native decimals to an sdk.Coin with evm decimals
+func convertCoinToEvm(coin sdk.Coin) sdk.Coin {
+	coin.Amount = coin.Amount.Mul(conversionMultiplier)
+
+	return coin
+}
+
+// convertCoinFromEvm converts an EVM sdk.Coin with 18 decimals back to native 6 decimal sdk.Coin
+func convertCoinFromEvm(coin sdk.Coin) sdk.Coin {
```
nddeluca, line 65: I like the unexported helpers here 👍

app/evm_keeper.go
```diff
@@ -0,0 +1,85 @@
...
+var _ evmtypes.BankKeeper = (*EVMBankKeeper)(nil)
+
+// NewEVMBankKeeper returns a wrapped bank keeper that converts between EVM and native decimals
+func NewEVMBankKeeper(bk bankkeeper.Keeper) EVMBankKeeper {
+	return EVMBankKeeper{
+		bankKeeper: bk,
+	}
+}
+
+func (bk EVMBankKeeper) GetBalance(ctx sdk.Context, addr sdk.AccAddress, denom string) sdk.Coin {
```
nddeluca, line 27: Could add doc comments for the exported functions

app/evm_keeper.go
```diff
@@ -0,0 +1,85 @@
...
+// Convertion between native gas KAVA (6) to EVM (18).
+// 12 decimal difference, so 1_000_000_000_000.
+var conversionMultiplier = sdk.NewInt(1_000_000_000_000)
+
+// EVMBankKeeper is a wrapper for bank keeper that converts between EVM and native decimals
+type EVMBankKeeper struct {
+	bankKeeper bankkeeper.Keeper
+}
+
+var _ evmtypes.BankKeeper = (*EVMBankKeeper)(nil)
```
nddeluca, line 18: Nice, like these types of assertions

nddeluca (changes requested):
Looks great.

Just two things:
  - Let's add full test coverage -- helps clarity and maintainability.
  - I think we should avoid rounding small amounts up to 1

app/evm_keeper.go
drklee3: Yeah, having an error here instead does makes sense. It should only be zero if it's dealing with values less than `1ukava` which shouldn't be valid anyways I suppose. Related thing I need to also check is if typical gas fees via wallets like MetaMask are less than `1ukava` which doesn't seem like great UX if users need to manually adjust it each time to not have it error.

Will add more tests for this after the weekend

drklee3: I think this might need specific logic to handle gas fees or some modification to fee calculations possibly in Ethermint, as some test transactions via MetaMask had default gas fees < 1ukava which causes transactions to fail with a panic.

DracoLi (commented):
do we want to return only the spendable balance for the EVM to account for vesting accounts?

drklee3:
> do we want to return only the spendable balance for the EVM to account for vesting accounts?

Right, added now. I left the test case to check vesting account balances as a TODO since EthAccount and vesting accounts aren't compatible right now

DracoLi (approved):

PR #1155, opened 2022-01-31
Remove panic if Swagger disabled in config

drklee3:
Swagger config option is now ignored

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -784,11 +784,7 @@ func (app *App) RegisterAPIRoutes(apiSvr *api.Server, apiConfig config.APIConfig
 	authtx.RegisterGRPCGatewayRoutes(clientCtx, apiSvr.GRPCGatewayRouter)
 	ModuleBasics.RegisterGRPCGatewayRoutes(clientCtx, apiSvr.GRPCGatewayRouter)
 
-	// register swagger API from root so that other applications can override easily
-	if apiConfig.Swagger {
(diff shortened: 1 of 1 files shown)


app/app.go
```diff
@@ -784,11 +784,7 @@ func (app *App) RegisterAPIRoutes(apiSvr *api.Server, apiConfig config.APIConfig
 	authtx.RegisterGRPCGatewayRoutes(clientCtx, apiSvr.GRPCGatewayRouter)
 	ModuleBasics.RegisterGRPCGatewayRoutes(clientCtx, apiSvr.GRPCGatewayRouter)
 
-	// register swagger API from root so that other applications can override easily
-	if apiConfig.Swagger {
-		// TODO where should old and new swagger docs be served? Should files be embedded in the binary?
-		panic("TODO: register swagger in app")
-	}
+	// Swagger API configuration is ignored
```
karzak, line 787: I assume there's no upstream functionality that happens - ie we don't need to do an explicit override?

```
  if apiConfig.Swagger {
    apiConfig.Swagger = false
  }
```

drklee3: Did not seem to be needed, the swagger API route needs to be manually added in app.go for it to be actually enabled so cosmos-sdk doesn't actually do anything with the config value

https://github.com/cosmos/gaia/blob/555fdd5872dc7ae3ec9a917047d689666d483f67/app/app.go

karzak (approved):

rhuairahrighairidh (approved):

PR #1157, opened 2022-02-02
Include all Kava grpc-gateway and legacy endpoints to Swagger 

drklee3:
- Adds all the remaining generated grpc-gateway Kava module swagger files (via `make proto-swagger-gen`)
- Adds IBC-go swagger data, fetched via `make proto-update-deps`
- Adds and cleans up legacy swagger file with only Kava module legacy endpoints. These endpoints have been marked as deprecated.
- Removes old `swagger-ui` directory

[sc-553]


karzak (approved):
Looks good - tested it out locally and ran a few queries.

Minor dev UX Q (cc @rhuairahrighairidh  @nddeluca ):  I assume there's no objection with having the legacy endpoints grayed out as deprecated?

PR #1158, opened 2022-02-03
Split existing auction `bid_duration` parameter into `forward_bid_duration` and `reverse_bid_duration`

drklee3:
Separates bid duration parameters depending on forward/reverse bids.  New end time behavior is as follows:

PlaceForwardBidCollateral:
* If the current forward bid converts the auction to a reverse bid (reached maxBid), `auction.EndTime = now + reverse_bid_duration`
* Otherwise `auction.EndTime = now + forward_bid_duration`

PlaceReverseBidCollateral:
* Always updates `auction.EndTime = now + reverse_bid_duration`

This is a breaking change for state.


proto/kava/auction/v1beta1/genesis.proto
```diff
@@ -25,14 +25,15 @@ message Params {
...
-  bytes increment_surplus = 3
+  bytes increment_surplus = 4
       [(gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec", (gogoproto.nullable) = false];
 
-  bytes increment_debt = 4
+  bytes increment_debt = 5
       [(gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec", (gogoproto.nullable) = false];
 
-  bytes increment_collateral = 5
+  bytes increment_collateral = 6
```
rhuairahrighairidh, line 37: It's best to not change field numbers in proto files. The don't need to be sequential, they're just unique ids.

Don't think it actually matters in this case of json genesis files. But generally if you keep field numbers the same it allows data encoded with an old proto file to be read by code with an updated proto file. Or the otherway round; some old code can still read messages sent to it encoded with a new proto file.

In this case you could do something like leave `bid_duration` as it and use as the forward bid duration, and add a new field `reverse_bid_duration` with id 6.

x/auction/keeper/auctions.go
```diff
@@ -443,7 +449,7 @@ func (k Keeper) PlaceBidDebt(ctx sdk.Context, auction *types.DebtAuction, bidder
 		auction.MaxEndTime = ctx.BlockTime().Add(k.GetParams(ctx).MaxAuctionDuration) // set maximum ending time on receipt of first bid
 		auction.HasReceivedBids = true
 	}
-	auction.EndTime = earliestTime(ctx.BlockTime().Add(k.GetParams(ctx).BidDuration), auction.MaxEndTime) // increment timeout, up to MaxEndTime
+	auction.EndTime = earliestTime(ctx.BlockTime().Add(k.GetParams(ctx).ReverseBidDuration), auction.MaxEndTime) // increment timeout, up to MaxEndTime
```
rhuairahrighairidh, line 452: I think it would be safer to use the longer forward bid duration here. These auctions trigger when collateral auctions have failed to raise enough usdx to cover debt. So the system mints kava and sells it to raise more usdx. If these auctions clear too fast, a huge amount of kava could be minted.

proto/kava/auction/v1beta1/genesis.proto
drklee3: Right, thought breaking changes would be fine but makes more sense if the proto files are backwards compatible until next versions

x/auction/types/params.go
```diff
@@ -70,16 +81,24 @@ func (p *Params) ParamSetPairs() paramtypes.ParamSetPairs {
...
 		return err
 	}
 
-	if p.BidDuration > p.MaxAuctionDuration {
-		return errors.New("bid duration param cannot be larger than max auction duration")
+	if p.ForwardBidDuration > p.MaxAuctionDuration {
+		return errors.New("forward bid duration param cannot be larger than max auction duration")
+	}
+
+	if p.ReverseBidDuration > p.MaxAuctionDuration {
```
DracoLi, line 100: do we need to validate forward duration is more than reverse? I don't remember if that was a requirement or not.

rhuairahrighairidh (approved):
Looks good

x/auction/types/params.go
karzak: No need - it's a valid config to have forward > reverse

PR #1171, opened 2022-02-17
EVM akava coins handling

DracoLi:
Creating a BankKeeper wrapper for x/evm that only allows requests for the "akava" denom.

This bank keeper will convert `akava` request from x/evm to both the `ukava` and remaining `akava` needed. Then, depending on the request, both `akava` and `ukava` denoms are used from the bank keeper to enable the use of both denoms as a single one by the evm.

Additional logic is also implemented to properly handle the minting of `akava` from `ukava` when an account does not have enough `akava` for transfers and also converting extra `akava` to `ukava`.


x/evmutils/keeper/bank_keeper.go
```diff
@@ -0,0 +1,209 @@
...
+	// convert 1 ukava to akava if sender account does not have enough akava to cover
+	akavaNeeded := coins.AmountOf(EvmDenom)
+	akavaBal := k.bk.GetBalance(ctx, senderAddr, EvmDenom).Amount
+	if akavaBal.LT(akavaNeeded) {
+		if err := k.convertOneUkavaToAkava(ctx, senderAddr); err != nil {
+			return err
+		}
+	}
+
+	return k.bk.SendCoinsFromAccountToModule(ctx, senderAddr, recipientModule, coins)
```
DracoLi, line 84: note: this module should never have enough akava for a ukava so we shouldn't need to do `convertAkavaToUkava` here

x/evmutils/keeper/bank_keeper.go
```diff
@@ -0,0 +1,207 @@
...
+
+	if len(coins) == 0 {
+		return coins, nil
+	}
+
+	if err := validateEvmCoins(coins); err != nil {
+		return splitCoins, err
+	}
+
+	// note: we should always have 1 coin here since coins cannot have dup denoms after we validate.
```
karzak, line 193:
```suggestion
	// note: we should always have len(coins) == 1 here since coins cannot have dup denoms after we validate.
```

karzak (changes requested):
Gave it a first pass - will give it another look after I use it with kvtool a bit more. 

Didn't find any issues with the implementation - just a few additional tests I'd like to see

DracoLi:
working on a big update that tracks akava outside of bank keeper. closing this PR

PR #1175, opened 2022-02-22
Further speed up docker builds

rhuairahrighairidh:
This PR increases docker build speeds any time dependencies are changed.

It uses the new cache feature to cache the download directory for go modules. So any time go.mod is changed, all the dependencies don't need to be redownloaded, only the new ones.


drklee3 (approved):
LGTM, does seem like we can remove the dependency cache layer with the extra buildkit mount

PR #1177, opened 2022-02-24
Create eth accounts from within evm

rhuairahrighairidh:
Wrap the account keeper passed to the evm to ensure that all accounts created by the evm keeper (eg during contract account creation or via transfering funds to a new address) are `EthAccount`s not `BaseAccount`s

There is still an edge case where a base account can be created on the kava side, but with the predicted address of a future contract deployment. Then when the contract is deployed, the code hash cannot be set. The failure appears when the contract is later called.


app/app.go
```diff
@@ -449,7 +448,10 @@ func NewApp(
 	)
 	app.evmKeeper = evmkeeper.NewKeeper(
 		appCodec, keys[evmtypes.StoreKey], tkeys[evmtypes.TransientKey], evmSubspace,
-		app.accountKeeper, NewEVMBankKeeper(app.bankKeeper), app.stakingKeeper, app.feeMarketKeeper,
+		NewEVMAuthKeeper(app.accountKeeper, authtypes.ProtoBaseAccount),
```
DracoLi, line 451: should `ethermint.ProtoAccount` be passed here?

rhuairahrighairidh: yes, updated in https://github.com/Kava-Labs/kava/pull/1177/commits/82e67bb64b2975af089c42aa55622c6e4f46fbe0

DracoLi:
once [this is](https://github.com/Kava-Labs/kava/pull/1179) is merged, i think we should consider moving this keeper into that module

rhuairahrighairidh:
Closing in favour of #1181. It converts accounts to Eth accounts when needed.

PR #1179, opened 2022-03-01
EvmUtil bank keeper

DracoLi:
`evmutil.Keeper` stores additional evm account data for addresses. The `EvmBankKeeper` uses the `evmutil.Keeper` to keep track of akava balance. 

## Testing Notes
- Metamask displays the correct balance, including the extra akava 
Screen Shot 2022-03-02 at 12 22 05 PM

- Gas seems to be correctly handled and after exporting genesis I was able to confirm the correctness in ukava, akava, and the akava collected by the fee collector. (all of akava on accounts add up to the total ukava held by the module)


app/app.go
```diff
@@ -192,6 +196,7 @@ var (
 		govtypes.ModuleName:             {authtypes.Burner},
 		ibctransfertypes.ModuleName:     {authtypes.Minter, authtypes.Burner},
 		evmtypes.ModuleName:             {authtypes.Minter, authtypes.Burner}, // used for secure addition and subtraction of balance using module account
+		evmutiltypes.ModuleName:         {authtypes.Minter, authtypes.Burner}, // used for the conversion of ukava <-> akava
```
DracoLi, line 199: might not be needed

x/evmutil/genesis.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/kava-labs/kava/x/evmutil/types"
+)
+
+// InitGenesis initializes the store state from a genesis state.
+func InitGenesis(ctx sdk.Context, keeper keeper.Keeper, gs *types.GenesisState) {
+	if err := gs.Validate(); err != nil {
+		panic(fmt.Sprintf("failed to validate %s genesis state: %s", types.ModuleName, err))
+	}
+
+	accounts := keeper.GetAllAccounts(ctx)
```
DracoLi, line 18: get it from gs

x/evmutil/keeper/bank_keeper.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	if ukava.IsPositive() {
+		if err := k.bk.SendCoinsFromAccountToModule(ctx, senderAddr, recipientModule, sdk.NewCoins(ukava)); err != nil {
+			return err
+		}
+	}
+
+	return k.akavaKeeper.SendBalance(ctx, senderAddr, k.GetModuleAddress(recipientModule), akavaNeeded)
+}
```
rhuairahrighairidh, line 104: Do these methods both need the `ConvertAkavaToUkava` and `ConvertOneUkavaToAkava` sections?
Otherwise the module account could end up with more than 1ukava worth of akava?

DracoLi: the module should always be a passthrough account so no need to have `ConvertAkavaToUkava` here when sending funds to it. ethermint will burn the sent funds right after sending it to the module.

DracoLi: Same with send from module to account, there should always be enough because ethermint mints the coins onto the module first.

x/evmutil/keeper/bank_keeper.go
```diff
@@ -0,0 +1,238 @@
...
+func NewEvmBankKeeper(akavaKeeper Keeper, bk types.BankKeeper, ak types.AccountKeeper) EvmBankKeeper {
+	return EvmBankKeeper{
+		akavaKeeper: akavaKeeper,
+		bk:          bk,
+		ak:          ak,
+	}
+}
+
+// GetBalance returns the total **spendable** balance of akava for a given account by address.
+func (k EvmBankKeeper) GetBalance(ctx sdk.Context, addr sdk.AccAddress, denom string) sdk.Coin {
```
karzak, line 47: Overall implementation looks good- tested in kvtool and balances looked correct - also fixes the gas issue 🎉. 

A couple non-blocking notes - 

* I think `ConvertAkavaToUkava` and `ConvertOneUkavaToAkava` would benefit from table driven tests
* Would like to see unit tests for `SendBalance` and `SetBalance`
* We should add a spec - would especially like a description or diagram that summarizes transaction flow through the module account

karzak (approved):

PR #1184, opened 2022-03-09
Add evm bank invariants

rhuairahrighairidh:
Ths PR adds a couple of invariants to help lock down bank state.
One detects when akava balances aren't exactly backed by ukava (ie when the evmutil module account's ukava balance (converted to akava) doesn't equal the sum of all akava balances).
Another detects any akava balances greater than one ukava (ie akava balance > `keeper.ConversionMultipler`).

There's a couple of edge cases where these aren't met:
The first can break when coins are burned in the evm (currently only possible via selfdestruct: https://ethereum.stackexchange.com/questions/16188/best-way-to-burn-ethers-and-other-ethereum-tokens)
The second invariant can break when gas fees are taken in the evm antehandler.


DracoLi (approved):
👍

PR #1187, opened 2022-03-16
Add spec for evmutil

DracoLi:


karzak (approved):
Few nits, otherwise LGTM

PR #1190, opened 2022-03-17
Savings module params

denalimarsh:
And parameters to the savings module, which is just a list of supported denoms.

Should be merged after the scaffolding PR.


rhuairahrighairidh (changes requested):
looks good, I'd add a test to check the `Get/SetParams` methods work as expected.

karzak (approved):
+1 to adding Get/Set params test, otherwise LGTM

denalimarsh:
> looks good, I'd add a test to check the `Get/SetParams` methods work as expected.

Added in latest.

PR #1169, opened 2022-02-16
Removal of `docs` directory

stephensprinkle:
This directory has now been promoted to it’s own repo: https://github.com/Kava-Labs/docs

Deployed site has been updated to point to this directory: https://docs.kava.io

PR #1196, opened 2022-03-24
Fix proto-gen after docs removal

rhuairahrighairidh:
After docs were removed (https://github.com/Kava-Labs/kava/pull/1169) `make proto-gen` failed as it tried to read a template for generating markdown docs. This PR adds the template back in. Not sure if anyone uses the proto markdown docs, but easy enough to keep them.

Also fixed an issue with linting and ran proto-all.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -174,7 +174,7 @@ proto-lint:
 	@$(DOCKER_BUF) lint --error-format=json
 
 proto-check-breaking:
-	@$(DOCKER_BUF) breaking --against $(HTTPS_GIT)#branch=upgrade-v44
+	@$(DOCKER_BUF) breaking --against $(HTTPS_GIT)#branch=master
 
 GOOGLE_PROTO_URL = https://raw.githubusercontent.com/googleapis/googleapis/master/google/api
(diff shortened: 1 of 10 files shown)


karzak (commented):
When I run this locally, I get changes to `go.mod` that require running `go mod tidy` afterwards. 

Tried just adding and additional `go mod tidy` to the end of protocgen.sh, but that didn't fix it.

rhuairahrighairidh:
> When I run this locally, I get changes to `go.mod` that require running `go mod tidy` afterwards.
> 
> Tried just adding and additional `go mod tidy` to the end of protocgen.sh, but that didn't fix it.

It's because the protogen script is run inside a docker container, which uses go 1.16, which formats the go.mod differently to go 1.17.
I can't update tendermint's docker images, but instead I removed `go mod tidy` from the script. In the future we might need to run it ourselves after proto generation. But forcing updating dependencies to be a bit more manual doesn't seem too bad.

nddeluca:
Updated to use the v0.3 tendermint image -- uses go 1.17 to fix the go mod issue

karzak (approved):

PR #1201, opened 2022-03-30
Add savings params to incentive module

denalimarsh:
Adds savings params to incentive module


proto/kava/incentive/v1beta1/params.proto
```diff
@@ -72,8 +72,11 @@ message Params {
   repeated MultiRewardPeriod swap_reward_periods = 5
       [(gogoproto.castrepeated) = "MultiRewardPeriods", (gogoproto.nullable) = false];
 
-  repeated MultipliersPerDenom claim_multipliers = 6
+  repeated MultiRewardPeriod savings_reward_periods = 6
+      [(gogoproto.castrepeated) = "MultiRewardPeriods", (gogoproto.nullable) = false];
+
+  repeated MultipliersPerDenom claim_multipliers = 7
```
karzak, line 78: @rhuairahrighairidh not familiar enough with proto types, is it OK to renumber the types like this or should we append types to the end?

karzak (approved):
Question is non-blocking, can always renumber after if needed.

proto/kava/incentive/v1beta1/params.proto
denalimarsh: Going to merge for now, if we need to reorder can do it in a follow on PR

rhuairahrighairidh: In general field numbers shouldn't be changed as it breaks compatibility. Something proto encoded with an old version of the software won't be able to be decoded with the new version. eg the old version encodes claim multipliers into field number 6 in the binary output, then the new version tries to decode field 6 as savings reward periods but the types won't match up.

I think it doesn't strictly matter for params as they're never binary encoded, but probably best to follow standard practises: https://developers.google.com/protocol-buffers/docs/proto3#updating eg give any new fields a new unused number.

You can use `make proto-check-breaking` to check for changes from master.

PR #1202, opened 2022-03-30
Implement savings module hooks interface/types

denalimarsh:
Implement savings hooks types + interface. They clashed with Hard hook names in the incentive module so had to prefix each hook with `Savings`


karzak (approved):
Looks good

PR #1204, opened 2022-03-31
Add savings claim types to incentive module

denalimarsh:
- New SavingsClaim type + tests.
- New MsgClaimSavingsClaim msg type + tests.
- Basic keeper logic to get/set Savings claims
- Basic keep logic for Savings reward accumulators including reward indexes and previous accrual times.


karzak (approved):
LGTM

PR #1205, opened 2022-03-31
Add savings to incentive genesis

denalimarsh:
Adds savings claims and savings reward state to incentive module genesis


karzak (changes requested):
LGTM after proto types are re-numbered

karzak (approved):

PR #1206, opened 2022-04-01
evm migrations for tesnet alpha 2

rhuairahrighairidh:
Add migrations for migrating evm-alpha testnet to a new version.

These migrations will need to updated for the mainet migration from v0.16.

I tested this out in kvtool and it's working as expected. The only migration run is the feemarket one. There's new fields in the evm params, but I think the zero values are ok.

Upgrade process:
- submit a `SoftwareUpgradeProposal`
  - proposal.title, propsal.description: anything
  - proposal.plan.name: use the const `upgradeName` set in `app.go`
  - proposal.plan.height: the height to upgrade at, looks like setting a time is deprecated
  - other fields aren't necessary
- vote it through
- at the plan height the chain will halt
- stop the node, start up the new binary
- it should run the migration and continue with new blocks


karzak (approved):
LGTM

PR #1209, opened 2022-04-15
Incentive savings hooks + init/sync of savings claims

denalimarsh:
- Keeper logic for init/sync of savings reward claims.
- Introduce savings hooks to incentive module and call them from the savings module

Currently marked as draft as I'm still working on the sync savings claim tests.

### DIFF
--- a/x/incentive/keeper/rewards_savings.go
+++ b/x/incentive/keeper/rewards_savings.go
@@ -38,20 +38,50 @@ func (k Keeper) AccumulateSavingsRewards(ctx sdk.Context, rewardPeriod types.Mul
 	}
 }
 
+// InitializeSavingsReward initializes a savings claim by creating the claim and
+// setting the reward factor indexes
+func (k Keeper) InitializeSavingsReward(ctx sdk.Context, deposit savingstypes.Deposit) {
+	claim, found := k.GetSavingsClaim(ctx, deposit.Depositor)
+	if !found {
+		claim = types.NewSavingsClaim(deposit.Depositor, sdk.Coins{}, nil)
+	}
+
+	var rewardIndexes types.MultiRewardIndexes
+	for _, coin := range deposit.Amount {
+		globalRewardIndexes, found := k.GetSavingsRewardIndexes(ctx, coin.Denom)
+		if !found {
+			globalRewardIndexes = types.RewardIndexes{}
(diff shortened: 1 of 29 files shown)


x/savings/keeper/deposit.go
```diff
@@ -21,13 +21,20 @@ func (k Keeper) Deposit(ctx sdk.Context, depositor sdk.AccAddress, coins sdk.Coi
 	}
 
 	currDeposit, foundDeposit := k.GetDeposit(ctx, depositor)
-	amount := coins
+
+	deposit := types.NewDeposit(depositor, coins)
 	if foundDeposit {
-		amount = amount.Add(currDeposit.Amount...)
+		deposit.Amount = deposit.Amount.Add(currDeposit.Amount...)
+		k.hooks.BeforeSavingsDepositModified(ctx, deposit, getDenoms(coins))
```
karzak: pretty sure `incomingDenoms` should be defined as only the new denoms. As currently written, if I have an existing `ukava` deposit to the savings module and add another `ukava` deposit, getDenoms will return `ukava` to the `incomingDemons` argument. What we want is the set difference between `coins` and `deposit.Amount`.

x/incentive/keeper/rewards_savings.go
```diff
@@ -38,30 +38,54 @@ func (k Keeper) AccumulateSavingsRewards(ctx sdk.Context, rewardPeriod types.Mul
...
 	claim, found := k.GetSavingsClaim(ctx, deposit.Depositor)
 	if !found {
 		return
 	}
 
-	// Source shares for savings deposits is the deposit amount
-	for _, coin := range deposit.Amount {
-		claim = k.synchronizeSingleSavingsReward(ctx, claim, coin.Denom, coin.Amount.ToDec())
+	// Set the reward factor on claim to the global reward factor for each incoming denom
+	for _, denom := range incomingDenoms {
```
karzak, line 71: Without the `setDiff` fix, this is not safe, because an existing `ukava` deposit will have its reward index set to the global reward index before `synchronizeSingleSavingsReward` is called.

karzak (approved):
Working as expected in `kvtool` with fix applied, please review my fix and make sure it makes sense to you. Otherwise, LGTM.

PR #1211, opened 2022-04-19
Add Ethereum bridge module

drklee3:
Initial v0.1.1 version of bridge module added


contrib/devnet/init-new-chain.sh
```diff
@@ -59,3 +75,30 @@ sed -in-place='' 's/aphoton/akava/g' $DATA/config/genesis.json
...
 # Zero out the total supply so it gets recalculated during InitGenesis
 jq '.app_state.bank.supply = []' $DATA/config/genesis.json|sponge $DATA/config/genesis.json
+
+# Set relayer to devnet relayer address
+jq '.app_state.bridge.params.relayer = "kava15tmj37vh7ch504px9fcfglmvx6y9m70646ev8t"' $DATA/config/genesis.json | sponge $DATA/config/genesis.json
+
+# Set enabled erc20 tokens to match local geth testnet
+jq '.app_state.bridge.params.enabled_erc20_tokens = [
+    {
+        address: "0x6098c27D41ec6dc280c2200A737D443b0AaA2E8F",
```
DracoLi, line 85: How are the contract addresses obtained? are these auto-deployed by the bridge?

drklee3: Yeah they are deployed by an [init script in the kava-bridge](https://github.com/Kava-Labs/kava-bridge/blob/main/contract/scripts/init_dev_env.ts) which is run when using `make start-geth` in the bridge. kvtool with the --geth flag also uses the same state with the same contracts. There's a handful of other addresses on the bridge readme for other contracts that are also deployed

It uses the [default accounts in hardhat](https://hardhat.org/hardhat-network/reference/#accounts) so the contract addresses should be deterministic and stay the same as long as the order of contract deployment is the same and the hardhat account config isn't changed.

DracoLi: :+1: thx for the details

app/app.go
```diff
@@ -775,6 +820,7 @@ func NewApp(
 		validatorvestingtypes.ModuleName,
 		evmutiltypes.ModuleName,
 		savingstypes.ModuleName,
+		bridgetypes.ModuleName,
```
rhuairahrighairidh, line 823: These `SetOrder_` methods are a bit of a footgun.

All the invariants run in crisis genesis, but here the bridge state won't have been loaded from genesis yet (as it appears later in the list). So the bridge invariants run on an empty state.

Don't think its a strict problem for bridge as empty state passes invariants, but worth fixing in case that changes.

Also applies to savings and evmutil. Do you want to move these up in the list as well?

rhuairahrighairidh (changes requested):

PR #1218, opened 2022-04-21
[sc-1734] Add v17 migration logic and tests for x/auction

galxy25:
- Add proto definitions that can unmarshal kava-9 auction state
- Add proto definitions that can marshal kava-2022-10 state (in a backwards compatible manner that doesn't re-use previous field numbers)
- Remove go.mod for migrate so tests will run
- Add makefile target for running migration cmd tests
- Updated tests and docs for the above
- Update default duration for auction's forward bid duration to 24 hours
- Delete  v16 migration code (for Kava 9 upgrade)
- Add rollback doc for v17 migration / kava-2022-10 upgrade

Related PRs
- https://github.com/Kava-Labs/kava/pull/1158


app/app.go
```diff
@@ -25,6 +25,9 @@ import (
 	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
```
galxy25, line 25: not sure where this commit is coming from, I created the current branch off of master which [already has these changes]()https://github.com/Kava-Labs/kava/commit/de0e1646321acd9dd1547c3c6174db8f59540238 in it 🤔

migrate/v0_17/kava.go
```diff
@@ -2,20 +2,24 @@ package v0_17
```
galxy25, line 2: non auto-generated code changes in this file

migrate/v0_17/migrate_test.go
```diff
@@ -10,6 +10,7 @@ import (
```
galxy25, line 10: non auto-generated code changes in this file

x/auction/legacy/v0_17/migrate.go
```diff
@@ -0,0 +1,42 @@
+package v0_17
```
galxy25, line 1: non auto-generated code changes in this file

x/auction/legacy/v0_17/migrate.go
```diff
@@ -0,0 +1,42 @@
...
+	"strings"
+
+	codectypes "github.com/cosmos/cosmos-sdk/codec/types"
+
+	v016auction "github.com/kava-labs/kava/x/auction/legacy/v0_16"
+	v017auction "github.com/kava-labs/kava/x/auction/types"
+)
+
+func migrateAuction(auction *codectypes.Any) *codectypes.Any {
+	auction.TypeUrl = strings.Replace(auction.TypeUrl, "v1beta1", "v1beta2",-1)
```
drklee3, line 13: v17 version of the old v16 tests would be good, maybe also one that makes sure the encoded bytes doesn't change between versions.

If the encoded byte data did change, another way could be using unpacking, converting to the new v17 type, then re-packing it as any.

galxy25: In this case the encoded bytes _will_ change since the type url is being updated. I'm still a protobuf novice, do you have time tomorrow to pair on the unpacking / any approach?

PR #1221, opened 2022-04-22
Audit app.go

rhuairahrighairidh:
I checked over app.go to try and catch any mistakes that made it through various merges.

Besides changes listed in commits, I checked how we setup:
- ibc v3 modules against gaia v7 app.go
- ethermint modules against evmos v3 app.go
- authz against gaia v7 app.go


nddeluca (approved):
looks good - read through a few times in detail to ensure all modules are defined before they are referenced

PR #1222, opened 2022-04-22
[sc-1736] Add x/incentive module migrations

galxy25:


migrate/v0_17/testdata/genesis-v16.json
```diff
@@ -1,2280 +1,2725 @@
 {
-  "genesis_time": "2022-01-19T16:00:00Z",
```
galxy25, line 2: The reformatting noise in this diff for these two files is the result of running vscode "Prettify JSON", would be happy to update my vscode config to match what the rest of the team is using!

nddeluca: We use the `jq` default of two spaces

nddeluca: Can pipe through jq to fix formatting also

PR #1228, opened 2022-04-25
Reset bep3 swaps for zero height

rhuairahrighairidh:
Bep3 swaps need to be modified to ensure safety of funds during upgrades due to the height reset.

This code has been copied over from previous migrations.

The completion height of completed swaps is also reset so they're cleaned up on time.


x/bep3/legacy/v0_17/migrate.go
```diff
@@ -0,0 +1,57 @@
...
+		case types.SWAP_DIRECTION_OUTGOING:
+			// Open outgoing swaps should be extended to allow enough time to claim after the chain launches.
+			// They cannot be expired as there could be an open/claimed bnb swap.
+			swap.ExpireHeight = 1 + 24686 // default timeout used when sending swaps from kava
+		case types.SWAP_DIRECTION_UNSPECIFIED:
+		default:
+			panic(fmt.Sprintf("unknown bep3 swap direction '%s'", dir))
+		}
+	case types.SWAP_STATUS_EXPIRED:
+		// Once a swap is marked expired the expire height is ignored. However reset to 1 to be sure.
```
galxy25, line 31: very helpful comments!

galxy25 (approved):

PR #1233, opened 2022-04-28
feat: add evm-alpha testnet migrations for new modules

karzak:
Modules added

* savings
* bridge
* authz

Note I'm merging into `kd-v0.17.0-alpha.3`, as these migrations are only needed for testnet and shouldn't be included in `master` branch. `kd-v0.17.0-alpha.3` is tagged from the current `v0.17.0-alpha.3` release

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -16,7 +16,6 @@ import (
 	"github.com/cosmos/cosmos-sdk/server/api"
 	"github.com/cosmos/cosmos-sdk/server/config"
 	servertypes "github.com/cosmos/cosmos-sdk/server/types"
-	store "github.com/cosmos/cosmos-sdk/store/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/types/module"
 	"github.com/cosmos/cosmos-sdk/version"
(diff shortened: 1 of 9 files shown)


app/app.go
```diff
@@ -890,6 +892,30 @@ func NewApp(
 		}
 	}
 
+	app.upgradeKeeper.SetUpgradeHandler(
+		upgradeName,
+		func(ctx sdk.Context, _ upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
+			delete(fromVM, authz.ModuleName)
+			delete(fromVM, bridgetypes.ModuleName)
+			delete(fromVM, savingstypes.ModuleName)
```
rhuairahrighairidh, line 900: suggestion: if the `app.upgradeKeeper.SetModuleVersionMap(ctx, app.mm.GetVersionMap())` line is removed in `InitChainer` below you shouldn't need to delete these module names.
On the current testnet, the store holds the existing module versions, so after updating to this binary these new modules will already be missing.

rhuairahrighairidh (approved):
Haven't tested, but looks like it should work

app/app.go
```diff
@@ -883,6 +884,19 @@ func NewApp(
...
 
+	app.upgradeKeeper.SetUpgradeHandler(
+		upgradeName,
+		func(ctx sdk.Context, _ upgradetypes.Plan, _ module.VersionMap) (module.VersionMap, error) {
+			fromVM := app.mm.GetVersionMap()
+			fromVM[incentivetypes.ModuleName] = 1
+			fromVM[auctiontypes.ModuleName] = 1
+			delete(fromVM, authz.ModuleName)
+			delete(fromVM, bridgetypes.ModuleName)
+			delete(fromVM, savingstypes.ModuleName)
```
rhuairahrighairidh, line 895:
```suggestion
		func(ctx sdk.Context, _ upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
```

Not necessary for testnet, but I think you shouldn't need to manually set these if you read `fromVM` from the func arg and you remove line 30 (`app.upgradeKeeper.SetModuleVersionMap(ctx, app.mm.GetVersionMap())`).

rhuairahrighairidh (approved):

PR #1234, opened 2022-05-02
Update committee permissions for kava-10

rhuairahrighairidh:
Adds missing permissions in the kava-10 migration.
- Stability Committee:
  - AKT hard
  - UST cdp
  - Auction bid duration
- Hard Gov:
  - AKT hard

I checked all the permissions against the mainnet params for all modules and these were the only items missing.

This also adds Get/Set methods to the AllowedParamsChanges slice type. Might be a bit overkill, but it should make modifying permissions a little easier in future migrations.

### DIFF
--- a/migrate/v0_17/testdata/genesis-v16.json
+++ b/migrate/v0_17/testdata/genesis-v16.json
@@ -1562,6 +1562,51 @@
                       },
                       {
                         "key": "market_id",
+                        "val": "kava:usd",
+                        "allowed_subparam_attr_changes": ["active"]
+                      },
+                      {
+                        "key": "market_id",
+                        "val": "kava:usd:30",
+                        "allowed_subparam_attr_changes": ["active"]
+                      },
+                      {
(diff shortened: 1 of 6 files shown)


migrate/v0_17/testdata/genesis-v16.json
```diff
@@ -1762,7 +1807,7 @@
                   },
                   {
                     "subspace": "incentive",
-                    "key": "HardDelegatorRewardPeriods",
```
DracoLi, line 1765: oh interesting, when was this changed on mainnet?

DracoLi (approved):
:+1:, the `Get` and `Set` additions are great

migrate/v0_17/testdata/genesis-v16.json
rhuairahrighairidh: hmm not sure, the name changed in the incentive module sometime year

PR #1238, opened 2022-05-03
Fix unsafe-reset-all command

rhuairahrighairidh:
Between v0.45.1 and v0.45.3 of the sdk the `unsafe-reset-all` cmd broke. Without the `--home` flag it would ignore the app's default home folder (`~/.kava`).

This turned out to be a problem with tendermint's config setup during startup. It reads config values from environment variables, but without any prefix. So tendermint's 'home' config was populated with the users's $HOME env var.

It's fixed on sdk v0.46+ so I backported the fix to v0.45.3 on our fork: https://github.com/Kava-Labs/cosmos-sdk/pull/113.


cmd/kava/cmd/root.go
```diff
@@ -39,7 +42,7 @@ func NewRootCmd() *cobra.Command {
 		WithAccountRetriever(types.AccountRetriever{}).
 		WithHomeDir(app.DefaultNodeHome).
 		WithKeyringOptions(hd.EthSecp256k1Option()).
-		WithViper("") // TODO this sets the env prefix
+		WithViper(EnvPrefix)
```
rhuairahrighairidh, line 45: This is not part of the fix, just clearing up the TODO.

nddeluca:
Looks good and I verified diff between upstream v0.45.3 and v0.45.3-kava.1 is just the backport

nddeluca (approved):

PR #1244, opened 2022-05-06
Upgrade to cosmos v0.45.4 & tm-db@master for cosmos/gorocksdb

faddat:
Tm-db now references cosmos/gorocksdb, but only in the master banch.

I do recommend that you use master, and there are a number of other updates pending.

As you asked in Slack-- yes, there are newer versions of rocksdb available, specifically 7.1.0 and there's a new wrapper library as well, which deprecates cosmos/gorocksdb.  The wrapper of choice in the go ecosystem is now linxGnu/grocksdb.


faddat:
FYI:  db performance will matter a lot once the evm is on, here is some context:

https://github.com/tendermint/tm-db/pull/241

https://github.com/tendermint/tm-db/pull/245

https://github.com/tendermint/tm-db/pull/239

All of the above plus:

https://github.com/tharsis/evmos/pull/542

https://github.com/tendermint/tm-db/pull/237

At the end of the day all of these relate to this:

https://github.com/tharsis/evmos/issues/562

nddeluca:
Sweet, thanks for links to those issues -- will diffidently look into them and keep an eye on performance with evm, probably even start testing some of them after launch

nddeluca (approved):

PR #1248, opened 2022-05-07
gofumpt

faddat:
This PR formats the code nicely.

I find that 

```bash
go install mvdan.cc/gofumpt@latest
gofumpt -l -w .
```

makes things look neat and tidy.


nddeluca:
Thanks nice changed, I also ran gofumpt and got the same output

PR #1253, opened 2022-05-18
Filter out the evm denom from min gas prices in the cosmos-sdk ante

nddeluca:
This is a non-invasive fix for the cosmos ante handler chain to ignore the evm min gas price set in app.toml.

Before this change setting `0ukava;1000000000akava` results in cosmos-sdk transactions requiring an akava fee that can not be satisfied.

This change modifies the sdk context before fees are checked in order to remove `akava` from the gas price list, resulting in the previous behavior of setting "0ukava" in app.toml.

This still allows unchanged behavior for requiring fees (0.001ukava) or multiple fees that are not akava.

### DIFF
--- a/app/ante/min_gas_filter.go
+++ b/app/ante/min_gas_filter.go
@@ -7,20 +7,24 @@ import (
 
 var _ sdk.AnteDecorator = EvmMinGasFilter{}
 
+// EVMKeeper specifies the interface that EvmMinGasFilter requires
 type EVMKeeper interface {
 	GetParams(ctx sdk.Context) evmtypes.Params
 }
 
+// EvmMinGasFilter filters out the EvmDenom min gas price and calls the next ante handle with an updated context
 type EvmMinGasFilter struct {
 	evmKeeper EVMKeeper
 }
 
+// NewEvmMinGasFilter takes an EVMKeeper and returns a new min gas filter for it's EvmDenom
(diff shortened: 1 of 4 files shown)


app/ante/min_gas_filter.go
```diff
@@ -0,0 +1,40 @@
...
+func NewEvmMinGasFilter(evmKeeper EVMKeeper) EvmMinGasFilter {
+	return EvmMinGasFilter{
+		evmKeeper: evmKeeper,
+	}
+}
+
+func (emgf EvmMinGasFilter) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
+	evmDenom := emgf.evmKeeper.GetParams(ctx).EvmDenom
+
+	if ctx.MinGasPrices().AmountOf(evmDenom).IsPositive() {
```
nddeluca, line 27: @rhuairahrighairidh is this the best method to filter a DecCoins?  `Add` is probably inefficient, but does result in a valid sorted DecCoins at the end.

rhuairahrighairidh: I don't know of any much better method. I've used `Add` as well for those reasons.

rhuairahrighairidh (approved):

PR #1256, opened 2022-05-25
Update migrate.md

iammelea:
from  
kvd export --for-zero-height --height 1610471 > export-genesis.json
to 
kava export --for-zero-height --height 1610471 > export-genesis.json


nddeluca:
thank you @iammelea

PR #1259, opened 2022-05-27
Register EthAccount with the legacy amino codec

nddeluca:
Added amino registration of eth account in https://github.com/Kava-Labs/ethermint/pull/8


rhuairahrighairidh (approved):

DracoLi (approved):

PR #1261, opened 2022-05-31
Use base account for new accounts.

rhuairahrighairidh:
Current situation: any new accounts are `EthAccount`s
Consequences:
- grpc account queries return `EthAccount`, causing problems for wallets (most expect a base account)
- claiming incentive rewards doesn't work as it doesn't handle `EthAccount`s
- sending cosmos txs is ok, `EthAccount`s can contain normal pubkey types.
- sending evm txs is ok
- contract accounts are fine, they should be EthAccount anyway to hold the code hash.

Intended situtation: any new account should be `BaseAccount`
Consequences:
- grpc account queries return `BaseAccount` as in kava-9
- claiming incentive rewards is ok
- sending cosmos txs is ok, works as in kava-9
- sending eth txs is ok as `BaseAccount`s can hold ethsecp256k1 pubkeys and their corresponding addresses. CodeHash is filled in with emptyHash when the account is loaded from state.
- creating contract accounts is ok as the evm module converts `BaseAccount` to `EthAccount` as needed to store the non empty codehash

How to migrate between these:
- #1260 replaces the default account proto func to return `BaseAccount`s after the upgrade
- But any new account created in kava-10 is an `EthAccount`. If it's a contract account that's good. But user accounts should be `BaseAccount`s for both cosmos and evm users.
- This PR adds an antehandler to convert an `EthAccount` to a `BaseAccount` when someone sends a tx from it. This is only implemented in the cosmos antehandler, but could be added to the evm antehandler if there's time.

### DIFF
--- a/app/ante/ante_test.go
+++ b/app/ante/ante_test.go
@@ -276,9 +276,10 @@ func TestAppAnteHandler_ConvertEthAccount(t *testing.T) {
 
 	tApp := app.NewTestApp()
 
-	tApp = tApp.InitializeFromGenesisStatesWithTimeAndChainID(
+	tApp = tApp.InitializeFromGenesisStatesWithTimeAndChainIDAndHeight(
 		time.Date(1998, 1, 1, 0, 0, 0, 0, time.UTC),
 		chainID,
+		app.FixDefaultAccountUpgradeHeight-2,
 		app.NewAuthBankGenesisBuilder().
 			WithAccounts(&ethermint.EthAccount{
 				BaseAccount: authtypes.NewBaseAccount(
@@ -295,7 +296,6 @@ func TestAppAnteHandler_ConvertEthAccount(t *testing.T) {
 			}).
 			BuildMarshalled(encodingConfig.Marshaler),
 	)
-	accKeeper := tApp.GetAccountKeeper()
 
 	stdTx, err := helpers.GenTx(
 		encodingConfig.TxConfig,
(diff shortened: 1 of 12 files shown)


app/ante/convert_account.go
```diff
@@ -0,0 +1,60 @@
...
+
+	return next(ctx, tx, simulate)
+}
+
+func convertAccount(ctx sdk.Context, ak AccountKeeper, address sdk.AccAddress) error {
+	acc, err := authante.GetSignerAcc(ctx, ak, address)
+	if err != nil {
+		return err
+	}
+	ethAcc, isEthAcc := acc.(*ethermint.EthAccount)
```
nddeluca, line 52: Should we also check the public key type here? If the public key is set, then eth secp256k1 key shouldn't be converted

rhuairahrighairidh: I believe we don't need to worry about pubkeys. But I'm doing some manual testing just now to make sure I understand everything

app/ante/ante_test.go
```diff
@@ -264,3 +268,85 @@ func TestAppAnteHandler_RejectMsgsInAuthz(t *testing.T) {
...
+	require.NoError(t, err)
+	txBytes, err = encodingConfig.TxConfig.TxEncoder()(stdTx)
+	require.NoError(t, err)
+
+	// Ensure CheckTx passes
+	// Check account converted at upgrade height
+	checkAccountIsConverted(t, testAddresses[0], tApp, txBytes, (*authtypes.BaseAccount)(nil))
+}
+
+func checkAccountIsConverted(t *testing.T, fromAddress sdk.AccAddress, tApp app.TestApp, txBytes []byte, expectedAccountType interface{}) {
```
nddeluca, line 341: nice like the extracted function

app/ante/convert_account.go
```diff
@@ -0,0 +1,60 @@
...
+	ak AccountKeeper
+}
+
+func NewConvertEthAccounts(ak AccountKeeper) ConvertEthAccounts {
+	return ConvertEthAccounts{
+		ak: ak,
+	}
+}
+
+func (cea ConvertEthAccounts) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
```
nddeluca, line 32: Should this run on checkTx and simulate?

PR #1266, opened 2022-06-02
Update migrate.md

JusTTeam:
Add genesis hash instead of [PLACEHOLDER]


pirtleshell (approved):
can confirm `3bc9829faf3beae2892ff1dfb7158f41a3f0cff303b4798777a559250a4dc815` is the correct hash of the genesis file.

PR #1273, opened 2022-07-02
liquid staking

denalimarsh:
This PR introduces the `liquidstaking` module. It allows users to convert a staking delegation into liquid coins, and back the way from liquid coins to a staking delegation. Each derivative token is specific to the validator it was minted from.

Some audit activities are still in process (including spec), but can be done after merge.


proto/kava/liquidstaking/v1beta1/tx.proto
```diff
@@ -0,0 +1,25 @@
...
+
+  // MintDerivative defines a method for minting deriviatives
+  rpc MintDerivative(MsgMintDerivative) returns (MsgMintDerivativeResponse);
+}
+
+// MsgMintDerivative defines the Msg/MintDerivative request type.
+message MsgMintDerivative {
+  string sender    = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+  string validator = 2;
+  bytes  shares = 3 [(gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec", (gogoproto.nullable) = false];
```
rhuairahrighairidh, line 21: I was thinking through whether we should specify shares vs ukava in msgs. I think specifying ukava is slightly better. eg I'd change `shares` to `repeated Coin amount`.

User's think about their staked balance in terms of ukava. So they'd specify units of ukava when converting to derivatives. These ukava units need to be converted to share units somewhere (either front end or backend). Backend is probably a better place for any subtle math.

When we mint derivatives I think rounding the share amounts to 6 digits is slightly better than multiplying by 10^18. 18 digit precision coins are a bit of a pain: have to be more careful math funcs don't overflow, and they're harder to display in UIs. Downside is up to 1ukava loss when someone burns their derivatives. But even at 1000x the price that's still on the order of gas fees.

x/liquidstaking/module.go
```diff
@@ -0,0 +1,201 @@
...
+
+// // RegisterStoreDecoder registers a decoder for price feed module's types
+// func (AppModuleBasic) RegisterStoreDecoder(sdr sdk.StoreDecoderRegistry) {
+// 	sdr[StoreKey] = simulation.DecodeStore
+// }
+
+// // WeightedOperations returns the all the price feed module operations with their respective weights.
+// func (am AppModule) WeightedOperations(simState module.SimulationState) []sim.WeightedOperation {
+// 	return simulation.WeightedOperations(simState.AppParams, simState.Cdc, am.accountKeeper, am.keeper)
+// }
```
rhuairahrighairidh, line 201: nit: probably won't create simulations so I'd remove the comment

rhuairahrighairidh (changes requested):

x/liquidstaking/module.go
```diff
@@ -0,0 +1,201 @@
+package liquidstaking
```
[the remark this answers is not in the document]
denalimarsh: added in https://github.com/Kava-Labs/kava/pull/1273/commits/18c16a7f03cea9b2fe7cd0b074fbd86d7b2b4b07

x/liquidstaking/module.go
denalimarsh: Included in https://github.com/Kava-Labs/kava/pull/1273/commits/00d8f042c26fbb12132b43ca7bd31fbd25b87cd4

rhuairahrighairidh (approved):
looks good, I've merged the other PRs into this branch to use as a staging branch while we figure out the intricacies of the staking module. Once those are done we can merge into master.

drklee3 (commented):
just some minor things

x/liquid/types/msg.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidAddress, err.Error())
+	}
+
+	_, err = sdk.ValAddressFromBech32(msg.Validator)
+	if err != nil {
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidAddress, err.Error())
+	}
+
+	if msg.Amount.IsNil() || !msg.Amount.IsValid() {
```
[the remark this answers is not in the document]
rhuairahrighairidh: I added the zero checks, might as well catch errors earlier.
The validator and bkava denom in the msg is a bit redundant, but I'll leave that tidy up to a future PR for expediancy as I'd need to test the parse method more.

PR #1274, opened 2022-07-12
Add basic empty Earn module

drklee3:
- AppModule stuff
- Empty proto tx/query services to be filled later
- Module in app.go


rhuairahrighairidh (approved):
looks good, I'll leave more general thoughts on #1278

PR #1277, opened 2022-07-15
Add basic Earn module vault deposit/withdraw

drklee3:
Adds Deposit and Withdraw messages that only send coins to the module account and withdraw.

Strategies are currently unimplemented and only include an empty instance that does nothing.


rhuairahrighairidh (approved):
looks good, I'll leave more general thoughts on #1278

PR #1279, opened 2022-07-22
Add Earn grpc query service and cli query commands

drklee3:


x/earn/keeper/grpc_query.go
```diff
@@ -33,3 +39,281 @@ func (s queryServer) Params(ctx context.Context, req *types.QueryParamsRequest)
...
+			Value:          value,
+		})
+
+		return &types.QueryDepositsResponse{
+			Deposits:   deposits,
+			Pagination: nil,
+		}, nil
+	}
+
+	// 4. All accounts, all vaults
```
rhuairahrighairidh, line 226: suggestion: I've found when I want to add comments dividing up a function into sections, it usually means the function is too big and doing things at several layers of abstraction. The sections can often be pulled out into a functions with similar names to the comments.

rhuairahrighairidh (approved):
looks good

x/earn/keeper/grpc_query.go
drklee3: Will update in follow up refactor

PR #1281, opened 2022-07-27
feat: add governance prop to support earn deposits from community pool

karzak:
Adds `CommunityPoolDepositProposal` type, which supports deposits from the community pool to earn vaults via governance proposal.


proto/kava/earn/v1beta1/proposal.proto
```diff
@@ -0,0 +1,38 @@
...
+import "cosmos/base/v1beta1/coin.proto";
+
+option go_package = "github.com/kava-labs/kava/x/earn/types";
+
+// CommunityPoolDepositProposal deposits from the community pool into an earn vault
+message CommunityPoolDepositProposal {
+  option (gogoproto.goproto_stringer) = false;
+  option (gogoproto.goproto_getters)  = false;
+
+  string                       title          = 1 [(gogoproto.moretags) = "yaml:\"title\""];
```
drklee3, line 14: I believe we don't need the yaml tags anymore, the `gogoproto.moretag` options can be removed

drklee3 (approved):

x/earn/keeper/proposal_handler.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/kava-labs/kava/x/earn/types"
+)
+
+// HandleCommunityPoolDepositProposal is a handler for executing a passed community pool deposit proposal
+func HandleCommunityPoolDepositProposal(ctx sdk.Context, k Keeper, p *types.CommunityPoolDepositProposal) error {
+	// deposit from community pool address (kava1jv65s3grqf6v6jl3dp4t6c9t9rk99cd8m2splc)
+	return k.Deposit(ctx, k.accountKeeper.GetModuleAddress("distribution"), p.Amount)
```
rhuairahrighairidh, line 12: I beleive we'll need to update the community pool object stored in `distribution` like in `DistributeFromFeePool`: https://github.com/cosmos/cosmos-sdk/blob/v0.45.6/x/distribution/keeper/fee_pool.go.

The distribution module account holds the community pool plus pending staking rewards, so if we don't update it we could deposit too much and prevent user's from being able to claim.

x/earn/types/proposal.go
```diff
@@ -0,0 +1,66 @@
...
+func (cdp CommunityPoolDepositProposal) String() string {
+
+	var b strings.Builder
+	b.WriteString(fmt.Sprintf(`Community Pool Multi Spend Proposal:
+  Title:       %s
+  Description: %s
+  Amount:   %s
+`, cdp.Title, cdp.Description, cdp.Amount))
+	return b.String()
+}
```
rhuairahrighairidh, line 57: nit: I think `String` methods aren't necessary with proto as they're generated

karzak:
closing in favor of #1297

PR #1283, opened 2022-08-02
Update vaults to use sdk.Dec shares

drklee3:
* Fixes share issuing for deposits after share value increases.
* User can withdraw with sdk.Coin amount and receive the requested amount.
* In some cases account value may be 1 coin less than deposited. Deposit 100, but account value is would be 99. This should be at max 1 ucoin, e.g. 1ukava or 0.000001 kava.


x/earn/keeper/vault_share.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	if totalValue.Amount.IsZero() {
+		return types.VaultShare{}, fmt.Errorf("total value of vault is zero")
+	}
+
+	// sharePrice := totalTokens / shareCount
+	// issuedShares = assetAmount / sharePrice
+	// issuedShares := assetAmount / (totalTokens / shareCount)
+	//               = assetAmount * (shareCount / totalTokens)
```
DracoLi, line 30: very useful explanation. small suggestion: I think we should update the vars here to match the vars used in the function.

x/earn/keeper/vault_share.go
```diff
@@ -0,0 +1,80 @@
...
+	//
+	// multiply by reciprocal  of sharePrice to avoid two divisions and limit
+	// rounding to one time
+
+	// Per share is not used here as it loses decimal values and can cause a 0
+	// share count.
+	// Division is done at the last step as decimals are truncated then.
+	// For example:
+	// 100 * 100 / 101   == 10000 / 101 == 99
+	// 100 * (100 / 101) == 100 * 0     == 0
```
DracoLi, line 40: if asset amount is 1, this will result in no shares. is that expected?

x/earn/keeper/vault_share.go
DracoLi: totalTokens should be totalValue right?

x/earn/keeper/vault_share_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+			beforeConvert: func() {
+				suite.addTotalShareAndValue(vaultDenom, sdk.NewDec(1000), sdk.NewInt(1001))
+			},
+			giveAmount: sdk.NewCoin(vaultDenom, sdk.NewInt(100)),
+			// 100 * 100 / 101 = 99.0099something
+			wantShares: types.NewVaultShare(vaultDenom, sdk.NewDec(100).MulInt64(1000).QuoInt64(1001)),
+		},
+		{
+			name: "multi step flow 1:1",
```
DracoLi, line 64: is this suppose to be a different test? Seems like the first 1:1 test

rhuairahrighairidh (changes requested):
I think it's probably worth adding some tests for the `VaultShares` methods. If the current code is directly cloned from `DecCoins` it's probably correct. But tests are useful to catch regressions if we add features, or updgrade the sdk version.

rhuairahrighairidh (approved):

PR #1287, opened 2022-08-04
Update to go v1.19

Sdpierret:
Following  the release of Go v1.19

- Update some modules to latest release without breaking change
- Run `make test` without error


rhuairahrighairidh:
Thanks for submitting PR. However to avoid requiring users to update to the latest version of go I'm going to leave it as is.

PR #1291, opened 2022-08-16
Kava 11 migrations setup

DracoLi:
- adds kava 11 migration handler
- adds earn module
- removes bridge module
- add migrations for evmutil & committee

Not meant for merging to master


app/upgrades.go
```diff
@@ -1,3 +1,188 @@
...
+			Oracles:    oracles,
+			Active:     true,
+		},
+		{
+			MarketID:   "dai:usd:30",
+			BaseAsset:  "dai",
+			QuoteAsset: "usd",
+			Oracles:    oracles,
+			Active:     true,
+		},
```
drklee3, line 177: Duplicate market IDs with the TWAP :30 ones, fails when setting params.

`panic: value from ParamSetPair is invalid: duplicated market usdt:usd:30`

DracoLi: :+1:, fixed the bad markets

nddeluca:
Replaced by #1359

PR #1293, opened 2022-08-24
Add specific `/vaults/bkava` and `/deposits` query handler to get aggregated bkava amounts

drklee3:
Note that the `TotalValue` response field is a plain sum of different derivative amounts which may differ in true ukava value.

### DIFF
--- a/x/earn/keeper/grpc_query.go
+++ b/x/earn/keeper/grpc_query.go
@@ -193,21 +193,32 @@ func (s queryServer) getAccountVaultDeposit(
 		return nil, status.Error(codes.NotFound, "No deposit found for owner")
 	}
 
-	if shareRecord.Shares.AmountOf(req.Denom).IsZero() {
-		return nil, status.Error(codes.NotFound, fmt.Sprintf("No deposit for denom %s found for owner", req.Denom))
-	}
+	var value sdk.Coin
+	if req.Denom == "bkava" {
+		totalAccountValue, err := getAccountTotalValue(ctx, s.keeper, depositor, shareRecord.Shares)
+		if err != nil {
+			return nil, err
+		}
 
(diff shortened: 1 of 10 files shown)


x/earn/keeper/grpc_query.go
```diff
@@ -266,8 +369,65 @@ func (s queryServer) getVaultAllDeposits(
...
+	}
+
+	return &types.QueryDepositsResponse{
+		Deposits:   deposits,
+		Pagination: pageRes,
+	}, nil
+}
+
+// get1AccountAllDeposits returns deposits for all vaults for a specific account
+func (s queryServer) get1AccountAllDeposits(
```
DracoLi, line 430: the number in the func is a bit hard to see. Thoughts on changing it to `getOneAccountAllDeposits` or maybe just `getAccountAllDeposits`?

x/earn/keeper/grpc_query.go
```diff
@@ -266,8 +369,65 @@ func (s queryServer) getVaultAllDeposits(
...
 
-// getAccountAllDeposits returns deposits for all vaults for a specific account
-func (s queryServer) getAccountAllDeposits(
+// getBkavaVaultAllDeposits returns all deposits for the aggregated bkava vault
+func (s queryServer) getBkavaVaultAllDeposits(
+	ctx sdk.Context,
+	req *types.QueryDepositsRequest,
+) (*types.QueryDepositsResponse, error) {
+	deposits := []types.DepositResponse{}
+	store := prefix.NewStore(ctx.KVStore(s.keeper.key), types.VaultShareRecordKeyPrefix)
```
DracoLi, line 378: will this have any performance concerns for these queries since we are iterating over all deposit shares every time?

rhuairahrighairidh (approved):

PR #1300, opened 2022-09-08
Support per-module incentive claim denom multipliers

drklee3:
Allows setting two sets of different "ukava" reward multipliers.

For example, one with 0 lockup only for the earn module, and one that applies for existing rewards:

```json
"claim_multipliers": [
  {
    "denom": "ukava",
    "multipliers": [
      {
        "name": "small",
        "months_lockup": "1",
        "factor": "0.200000000000000000"
      },
      {
        "name": "large",
        "months_lockup": "12",
        "factor": "1.000000000000000000"
      }
    ]
  },
  {
    "denom": "ukava",
    "multipliers": [
      {
        "name": "large",
        "months_lockup": "0",
        "factor": "1.000000000000000000"
      }
    ],
   "module_name": "earn"
  }
]
```


rhuairahrighairidh (approved):
looks good. It would be nice to have an integration test for earn claims (probably duplicate the msg_server_x_test.go tests), but I've added that as a card so we can merge this sooner.

PR #1308, opened 2022-09-19
Add liquid staking reward redistribution via incentive

drklee3:


rhuairahrighairidh (changes requested):
I walked through the logic several times trying to think of ways to break it and couldn't come up with much.
Although I think `GetTotalDerivativeSupply` and `GetDerivativeSupply` will need to return the ukava amount rather than bkava. If you got intentionally slashed on a small stake a few times you could get a permanently better reward rate than most other validators I think. The lower share price sort of multiplies a validator's ukava to get a higher proportion of rewards.

x/incentive/keeper/rewards_earn.go
```diff
@@ -78,14 +77,17 @@ func (k Keeper) accumulateEarnBkavaRewards(ctx sdk.Context, rewardPeriod types.M
...
 			bkavaVaultsDenoms[vaultDenom] = true
 		}
 
 		return false
 	})
 
-	totalBkavaSupply := k.liquidKeeper.GetTotalDerivativeSupply(ctx)
+	totalBkavaValue, err := k.liquidKeeper.GetTotalDerivativeValue(ctx)
+	if err != nil {
+		panic(err)
```
drklee3, line 89: are panics ok in the BeginBlocker? Or should this just log and return early

rhuairahrighairidh: In other modules we've returned errors, then panicked in abci.go if an error is returned.
I'd probably do that (ie panic rather than log and skip), then if there's problems in testnet we'll know about them.

rhuairahrighairidh (approved):
looks good

PR #1325, opened 2022-09-30
Add Earn deposits `value_in_staked_tokens` query parameter

drklee3:
Adding the query param `value_in_staked_tokens=true` to `/kava/earn/v1beta1/deposits` will return a underlying ukava values for all separate bkava vaults. Currently specifying `denom=bkava` will return a *total* sum of all bkava deposits without a way to determine per-vault/denom ukava value.

### Example response

GET `/kava/earn/v1beta1/deposits?depositor=kava1ffv7nhd3z6sych2qpqkk03ec6hzkmufy0r2s4c&value_in_staked_tokens=true`

```json
{
  "deposits": [
    {
      "depositor": "kava1ffv7nhd3z6sych2qpqkk03ec6hzkmufy0r2s4c",
      "shares": [
        {
          "denom": "bkava-kavavaloper1ffv7nhd3z6sych2qpqkk03ec6hzkmufyz4scd0",
          "amount": "1000.000000000000000000"
        }
      ],
      "value": [
        {
          "denom": "ukava",
          "amount": "1000"
        }
      ]
    }
  ],
  "pagination": null
}
```


x/earn/keeper/grpc_query.go
```diff
@@ -397,6 +397,28 @@ func (s queryServer) getOneAccountAllDeposits(
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
+	if req.ValueInStakedTokens {
```
nddeluca, line 400: We *might* want to consider reducing this to a single depsoit address address or adding a hard pagination limit.  A N+(1 to many) query type situation with many a thousand or more deposits would get pretty slow and likely stall block processing

drklee3: This should be limited to only 1 account, the base query handler for Deposits requires a Depositor address

https://github.com/Kava-Labs/kava/blob/b40d923c13c06e3161a5f96c48f1cb218142956c/x/earn/keeper/grpc_query.go#L247-L249

x/earn/testutil/suite.go
```diff
@@ -357,6 +368,67 @@ func (suite *Suite) SavingsDepositAmountEqual(expected sdk.Coins) {
 	)
 }
 
+// ----------------------------------------------------------------------------
+// Staking
+
+// CreateNewUnbondedValidator creates a new validator in the staking module.
+// New validators are unbonded until the end blocker is run.
+func (suite *Suite) CreateNewUnbondedValidator(addr sdk.ValAddress, selfDelegation sdk.Int) stakingtypes.Validator {
```
nddeluca, line 376: Like the helplers 👍

nddeluca (approved):

x/earn/testutil/suite.go
drklee3: Copied from liquid 😛

PR #1326, opened 2022-10-03
Add migration to convert EOA EthAccounts to BaseAccounts

drklee3:
List of account addresses created from exported genesis file.

```bash
cat genesis.json | jq '.app_state.auth.accounts | map(select(.code_hash == "0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470")) | [.[].base_account.address]' > eth_eoa_addresses.json
```

EthAccounts with an "empty" codehash: 33,143 out of 2,217,435 total accounts (1.5% of all accounts)


app/upgrades_test.go
```diff
@@ -94,3 +98,64 @@ func (suite *UpgradeTestSuite) TestUpdateSavingsParams() {
...
 	)
 }
+
+func (suite *UpgradeTestSuite) TestConvertEOAsToBaseAccount() {
+	ak := suite.App.GetAccountKeeper()
+
+	accCount := 0
+
+	// Add all accounts as EthAccount
+	app.IterateEOAAddresses(func(addr string) {
```
DracoLi, line 108: this test is running through all 33k records right? How long does this take?

drklee3: Yep it goes through them all, the test consistently finishes in under 1s for me

galxy25 (approved):

galxy25:
@drklee3 is this filtering out all module accounts (if any)?

DracoLi (approved):

drklee3:
> @drklee3 is this filtering out all module accounts (if any)?

The address list should only include `EthAccounts` and there's an additional type check to skip modifying accounts that aren't `EthAccount`s

PR #1327, opened 2022-10-04
Add Vested Delegated API endpoint

rhuairahrighairidh:
Adds a `liquid/delegated_balance` endoint to return a user's total staked balance, split by how much is vesting/vested.
It's intended for use in determining how much of a user's existing delegations can be converted to derivatives; so it ignores tokens in unbonding delegations as they cannot be converted.

Also fixes kavadist proto file linting.


karzak (approved):
Tested against vesting account with `kvtool` - looks good

nddeluca (approved):

PR #1331, opened 2022-10-05
Fix high gas usage for coin conversions

nddeluca:
Uses an infinite gas meter during evm gas estimation and apply message, then uses the evm response to charge gas.

### DIFF
--- a/x/evmutil/keeper/evm.go
+++ b/x/evmutil/keeper/evm.go
@@ -89,6 +89,8 @@ func (k Keeper) CallEVMWithData(
 		return nil, err
 	}
 
+	ethGasContext := ctx.WithGasMeter(sdk.NewInfiniteGasMeter())
+
 	// EstimateGas applies the transaction against current block state to get
 	// optimal gas value. Since this is done right before the ApplyMessage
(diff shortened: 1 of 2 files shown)


x/evmutil/keeper/evm.go
```diff
@@ -118,11 +120,13 @@ func (k Keeper) CallEVMWithData(
 		true,                  // checkNonce
 	)
 
-	res, err := k.evmKeeper.ApplyMessage(ctx, msg, evmtypes.NewNoOpTracer(), true)
+	res, err := k.evmKeeper.ApplyMessage(ethGasContext, msg, evmtypes.NewNoOpTracer(), true)
 	if err != nil {
 		return nil, err
 	}
 
+	ctx.GasMeter().ConsumeGas(res.GasUsed, "evm gas consumed")
```
nddeluca, line 128: @rhuairahrighairidh @DracoLi possible we should consume gas after the res.Failed() check?

rhuairahrighairidh (approved):
Looks good, spent a while trying to find issues and didn't come up with anything.

PR #1332, opened 2022-10-05
kavadist module account migration

karzak:
Instantiates new kavadist module account (kava1g655pg22nnhcsku4v3pe7mz9tzqj37rlej0jmn) as part of migration. As a security measure, checks that the account doesn't already exist as a non-module account and transfers the balance to the community pool.

### DIFF
--- a/app/upgrades.go
+++ b/app/upgrades.go
@@ -3,6 +3,7 @@ package app
 import (
 	_ "embed"
 	"encoding/json"
+	"fmt"
 
 	storetypes "github.com/cosmos/cosmos-sdk/store/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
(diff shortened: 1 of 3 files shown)


app/upgrades.go
```diff
@@ -136,6 +141,32 @@ func ConvertEOAsToBaseAccount(ctx sdk.Context, accountKeeper authkeeper.AccountK
 	})
 }
 
+func AddKavadistFundAccount(ctx sdk.Context, accountKeeper authkeeper.AccountKeeper, bankKeeper bankKeeper.Keeper, distKeeper distrkeeper.Keeper) {
+	maccAddr := accountKeeper.GetModuleAddress(kavadisttypes.FundModuleAccount)
+
+	accountI := accountKeeper.GetAccount(ctx, maccAddr)
+	// if account already exists and is a module account, return
```
galxy25, line 148: assuming only we/the chain can create new module accounts (e.g. not something that can be done over the api or cli)?

galxy25 (approved):

app/upgrades.go
karzak: Correct, usually the are instatiated at `initGenesis`, but since this module already exists, we need to add it manually in the migration.

drklee3 (approved):
lgtm, tested and confirmed it replaces an existing baseaccount with the module account

nddeluca (approved):

rhuairahrighairidh (approved):

PR #1333, opened 2022-10-05
increment the account number for kavadist mod account

rhuairahrighairidh:
This copies over the module account creation code from `accountkeeper.GetModuleAccount`.
It increments the global account sequence number and sets the latest on the module account.


karzak:
cherry picked into #1332

PR #1339, opened 2022-10-07
Remove embedded json EOA EthAccounts to BaseAccount conversion

drklee3:
Follow up of #1326

Removes the need of the embed package and just uses a go file.


nddeluca (approved):
nice, lgtm

PR #1347, opened 2022-10-11
fix: Use correct share values for earn incentive hooks

drklee3:


x/incentive/keeper/msg_server_earn_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+	err = suite.DeliverEarnMsgDeposit(userAddr2, c(bkavaDenom1, 99e9), earntypes.STRATEGY_TYPE_SAVINGS)
+	suite.Require().NoError(err)
+	err = suite.DeliverEarnMsgDeposit(userAddr2, c(bkavaDenom2, 99e9), earntypes.STRATEGY_TYPE_SAVINGS)
+	suite.Require().NoError(err)
+
+	// Accumulate some staking rewards
+	// _ = suite.App.EndBlocker(suite.Ctx, abci.RequestEndBlock{})
+	// suite.Ctx = suite.Ctx.WithBlockTime(time.Now().Add(1 * time.Hour))
+	// suite.App.BeginBlocker(suite.Ctx, abci.RequestBeginBlock{})
```
rhuairahrighairidh, line 94: nit: I'd maybe use the app begin end blockers so test matches production more closely, rather than calling the keeper methods to distribution etc. Was there problems with using endblockers?

rhuairahrighairidh (approved):
fixes look good, probably worth adding more integration tests but that can be for future PRs

also suggestion: if you setup the mocks with `earnHooks := mocks.NewEarnHooks(suite.T())` it has nicer error messages (doesn't use panics), and auto registers the `AssertExpectations` call. more info: https://github.com/vektra/mockery#mock-constructors

PR #1349, opened 2022-10-12
Add tally handler tests

rhuairahrighairidh:
Adds some simple tests for the new tally handler.

Might add a test with more complex state.


nddeluca (approved):
nice, like the helpers.   Tests look easy to read & modify

PR #1352, opened 2022-10-16
dragonberry

faddat:
- dragonberry
- state sync script
- lint fixes
- lint fixes
- save place
- refactor to resemble evmos so that iavl fast node will work
- encoding error left
- wrong encodingconfig type and use initcmd in genuntilcli

This PR is incomplete, but is most of the way there.  Submitting now, because time is rather of the 
essence.


nddeluca:
Resolved by #1353

PR #1355, opened 2022-10-18
Update to cosmos 0.45.9

nddeluca:


cmd/kava/cmd/app.go
```diff
@@ -77,6 +77,11 @@ func (ac appCreator) newApp(
 		panic(fmt.Sprintf("could not get authorized address from config: %v", err))
 	}
 
+	iavlDisableFastNode := appOpts.Get(server.FlagIAVLFastNode)
+	if iavlDisableFastNode == nil {
+		iavlDisableFastNode = true
```
karzak, line 82: Do we also want to override if `iavlDisableFastNode == false`?

karzak (approved):

cmd/kava/cmd/app.go
nddeluca: No, that should only happen if validators make the choice to put it in their config files.  On v0.45.9, the default value is true when the app.toml file is generated on commit (and that's adjustable, evmos modifies the default config value to false).

We just want to ensure that validators who do not generate a new app.toml do not inadvertently turn it on since omission (nil) casts to false.

nddeluca: If set to false, then there is a long delay at startup and it can be hours for slow nodes with a lot of state.  In addition, if fast node is enabled, it can not be disabled without a reset atm due to a bug in iavl.

PR #1358, opened 2022-10-20
Bump ibc to v3.3.0

nddeluca:
Use the latest v3 of IBC in order to allow base denoms containing slashes


nddeluca:
will be including migration on the release branch

PR #1362, opened 2022-10-24
Update Protos

nddeluca:
- Update proto generation for go 1.18
- Update third party proto files for new deps
- Update Swagger
- Re-gen proto files


nddeluca:
Will backport to release/v0.19.x as well

PR #1364, opened 2022-10-24
Fix: install protoc-gen-doc

rhuairahrighairidh:
After the recent update to proto generation https://github.com/Kava-Labs/kava/pull/1362, `Make proto-all` didn't work for me locally.
Looks like the newly added `protoc-gen-doc` tool wasn't being installed.
I think the problem is with go 1.18 `go get` not longer builds or installs anything (mentioned at the bottom of this section: https://go.dev/ref/mod#go-get).
I added a `go install ...` line to install the package. Without a version suffix this uses the local go.mod for decide dependency versions. If we don't care about that I think we can remove the `go get` line and add a version suffix to `go install`.

### DIFF
--- a/scripts/protocgen.sh
+++ b/scripts/protocgen.sh
@@ -15,6 +15,7 @@ protoc_gen_gocosmos() {
 
 protoc_gen_doc() {
   go get -u github.com/pseudomuto/protoc-gen-doc/cmd/protoc-gen-doc 2>/dev/null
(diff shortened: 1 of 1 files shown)


galxy25 (approved):
lgtm, `make proto-all` works for me locally running on

```
go version go1.19.1 darwin/arm64
```

nddeluca:
Thanks @rhuairahrighairidh, I think I was running a cached image that I already called install on

nddeluca (approved):

PR #1365, opened 2022-10-24
Spike: combine incentive claim types

rhuairahrighairidh:
Spike into deplicated the claim types in incentive.

summary of changes:
- new claim type
- replaced `Get/SetXClaim` with a general `Get/SetClaim` (only for swap and delegation, others TODO)
- replaced `Get/SetXRewardIndexes` with general `Get/SetGlobalIndexes` (only for swap and delegation, others TODO)
- removed redundant sync test

A potential downside: when all rewards use the same claim type, it would be harder to change specific rewards. eg adding some new feature to swap rewards, but not any others.


x/incentive/types/claims.go
```diff
@@ -8,6 +8,19 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
+// RewardType is the id of some rewards
+type RewardType int
```
galxy25, line 12: @rhuairahrighairidh regarding

```
A potential downside: when all rewards use the same claim type, it would be harder to change specific rewards. eg adding some new feature to swap rewards, but not any others.
```

could this be used in the future to enable targeted behavior per reward type? E.g. allowing calling code to filter out non swap claims?

PR #1368, opened 2022-10-25
Update ethermint for account type fix

nddeluca:


nddeluca:
Port of ethermint side fix from #1367

PR #1375, opened 2022-10-31
Add incentive claim state methods

drklee3:


x/incentive/types/keys.go
```diff
@@ -40,3 +42,42 @@ var (
...
+func createKey(bytes ...[]byte) (r []byte) {
+	for _, b := range bytes {
+		r = append(r, b...)
+	}
+	return
+}
+
+func getKeyPrefix(dataTypePrefix []byte, claimType ClaimType) []byte {
+	b := make([]byte, 4)
+	binary.LittleEndian.PutUint32(b, uint32(claimType))
```
rhuairahrighairidh, line 63: any reason for little endian over big endian?

rhuairahrighairidh (commented):
looks good, left a question on keys
I think we don't need migrations as the store changes are backwards compatible.

x/incentive/types/keys.go
drklee3: hmm no reason in particular, would there be a benefit for one over the other?

rhuairahrighairidh: In this case I think there's no difference. But if we were creating an key for `int32`s it would create an odd sort order as the last byte would be first in the key. Also casting `claimType` (`int32` underlying) to `uint32` orders negative numbers after positive.

I'd probably standardize on big endian, just to avoid our future selves using it when a sort order is needed.

drklee3: ah right sort order is a good point, will update it accordingly

rhuairahrighairidh (approved):

PR #1377, opened 2022-11-01
Add source adapter interface definition

rhuairahrighairidh:
Adds the interface definition for `SourceAdapter`s - the abstraction connecting the keeper sync and accumulate methods to external modules.

I added a spike of use in swap and earn here: #1378

Only difference to spike doc is `GetShares` requries a list of `sourceID`s, returning the shares for all of them at once. Most external modules store user's shares for different denoms in the same object (eg earn `VaultShareRecord`). So if the method was `GetShares(ctx, owner, sourceID)` it would need to fetch and unmarshal the whole object every time it's called. In the case of earn, this would be called for every bkava denom.

Another option would be `GetShares(ctx, owner) map[string]sdk.Dec` that returns all shares owned by a user. But this kind of has the opposite problem of forcing the adapter to fetch all shares when only one may be needed. eg in the case of swap, a users's pool shares are stored in separate objects. Also it returns a map, which introduces indeterminism if iterated over incorrectly (it could return a determinsitic type, but that's more work to create).

Open to thoughts on naming. `SourceShareFetcher` might be a bit clearer, but this object may be expanded to include hooks as in #1370.
I've been trying standardize on "source" being a single rewarded activity, like a single earn deposit, or single swap pool deposit.

### DIFF
--- a/x/incentive/types/adapter.go
+++ b/x/incentive/types/adapter.go
@@ -1,0 +1,17 @@
+package types
+
+import sdk "github.com/cosmos/cosmos-sdk/types"
+
+// SourceAdapter provides source shares from an external module.
+type SourceAdapter interface {
+	// GetShares returns source shares owned by one address.
+	//
+	// For example, the shares a user owns in the kava:usdx and bnb:usdx swap pools.
+	// It returns the shares for several sources at once, in the same order as the sourceIDs. Specifying no sourceIDS will return no shares.
+	GetShares(ctx sdk.Context, owner sdk.AccAddress, sourceIDs []string) []sdk.Dec
+
+	// GetTotalShares returns the sum of all shares for a source (across all users).
+	//
+	// For example, the total number of shares in the kava:usdx swap pool for all users.
+	GetTotalShares(ctx sdk.Context, sourceID string) sdk.Dec
+}
--- a/x/incentive/types/adapter.go
+++ b/x/incentive/types/adapter.go
@@ -2,16 +2,16 @@ package types
 
 import sdk "github.com/cosmos/cosmos-sdk/types"
 
-// SourceAdapter provides source shares from an external module.
+// SourceAdapter queries source shares from an external module.
 type SourceAdapter interface {
-	// GetShares returns source shares owned by one address.
+	// OwnerSharesBySource returns source shares owned by one address.
 	//
 	// For example, the shares a user owns in the kava:usdx and bnb:usdx swap pools.
 	// It returns the shares for several sources at once, in the same order as the sourceIDs. Specifying no sourceIDS will return no shares.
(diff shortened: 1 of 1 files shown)


x/incentive/types/adapter.go
```diff
@@ -0,0 +1,17 @@
...
+
+import sdk "github.com/cosmos/cosmos-sdk/types"
+
+// SourceAdapter provides source shares from an external module.
+type SourceAdapter interface {
+	// GetShares returns source shares owned by one address.
+	//
+	// For example, the shares a user owns in the kava:usdx and bnb:usdx swap pools.
+	// It returns the shares for several sources at once, in the same order as the sourceIDs. Specifying no sourceIDS will return no shares.
+	GetShares(ctx sdk.Context, owner sdk.AccAddress, sourceIDs []string) []sdk.Dec
```
nddeluca, line 11: Might be a good idea to prefix this -- helps for clarity when not being compared to TotalShares and scope code searches (grep, ctags, etc).

Maybe `GetOwnerShares` or `OwnerSharesBySource`, etc?

x/incentive/types/adapter.go
```diff
@@ -0,0 +1,17 @@
...
+
+	// GetTotalShares returns the sum of all shares for a source (across all users).
+	//
+	// For example, the total number of shares in the kava:usdx swap pool for all users.
+	GetTotalShares(ctx sdk.Context, sourceID string) sdk.Dec
```
nddeluca, line 16: Could also name for clarity that this is scoped by source.  `TotalSharesBySource`, etc to not be confused with a total of all sources

x/incentive/types/adapter.go
drklee3, line 11: Thoughts on making this return a map instead? Returning a slice that must be the same order seems a bit more easier to make errors compared to dealing with iterating over a map deterministically.

e.g. in the spike using it, each adapter implementation must be careful to return the exact order vs. if returning a map, we may be able to localize the deterministic iteration to 1 method.

https://github.com/Kava-Labs/kava/blob/49e8e4b1c4378858765ae70ddbf33896c5a89bd4/x/incentive/keeper/rewards_swap.go#L115-L117

rhuairahrighairidh: Yeah we could try it out. I don't have a strong opinon, both aren't great.
Good point that new adapters will be written more frequently than code consuming the interface.
We'll have to be careful to check each value exists in the map when reading from it, otherwise it'll return `sdk.Dec{}` which panics whenever it's used.
If we're only using it one method a map is probably ok. We've had several indeterminsm bugs from map iteration, and I think Juno went down partially due to it as well. Seems to be really easy to forget about.

rhuairahrighairidh: I normally try and rely on the containing thing (`SourceAdapter` in this case) and argument types to add context to avoid realy long names, but these do seem clearer

PR #1379, opened 2022-11-02
Add AccrualTime type and state methods

drklee3:
Adds a new proto type `AccrualTime` which contains the same fields as `AccumulationTime` but with an additional `ClaimType` field. We can get the same info from the store key, but it requires additional byte slice parsing that may be difficult to update later on.


x/incentive/keeper/keeper.go
```diff
@@ -981,3 +981,89 @@ func (k Keeper) GetAllClaims(ctx sdk.Context) types.Claims {
...
+
+// GetRewardAccrualTime fetches the last time rewards were accrued for the
+// specified ClaimType and subKey.
+func (k Keeper) GetRewardAccrualTime(
+	ctx sdk.Context,
+	claimType types.ClaimType,
+	subKey string,
+) (time.Time, bool) {
+	store := prefix.NewStore(ctx.KVStore(k.key), types.GetPreviousRewardAccrualTimeKeyPrefix(claimType))
+	b := store.Get([]byte(subKey))
```
rhuairahrighairidh, line 993: nit: I'd maybe use a function to convert the subKey into a byte key. Then there's just one place that defines the conversion.

x/incentive/keeper/keeper.go
```diff
@@ -981,3 +981,89 @@ func (k Keeper) GetAllClaims(ctx sdk.Context) types.Claims {
...
+
+		if cb(accrualTime.CollateralType, accrualTime.PreviousAccumulationTime) {
+			break
+		}
+	}
+}
+
+// IterateAllRewardAccrualTimes iterates over all reward accrual times of any
+// claimType and performs a callback function.
+func (k Keeper) IterateAllRewardAccrualTimes(
```
rhuairahrighairidh, line 1040: naming suggestion: It can be clearer to have the `IterateX` method iterate over all `X` and `IterateXByY` methods for iterating over `X` in groups. Makes it easier to add additional Ys, like `IterateClaimsByOwner`, `IterateClaimsByClaimType` (not that that applies for `AccrualTimes`)

rhuairahrighairidh (approved):

x/incentive/keeper/keeper.go
drklee3: good idea for later iteration groups

PR #1382, opened 2022-11-03
remove all yaml tags from proto definitions

pirtleshell:
the yaml tags are not needed as they are generated automatically from the json tags.

additionally fixed proto formatting error requiring doc comments on enum values (for x/incentive claim type enum).


nddeluca (approved):
👍

PR #1383, opened 2022-11-04
Add Initialize/Synchronize Claim methods

drklee3:
This does not contain `GetSynchronizedClaim`, tests require at least one implementation of sourceadapter


rhuairahrighairidh (approved):
looks good, I'll add my tests in another PR

PR #1386, opened 2022-11-08
Add GetSynchronizedClaim and swap adapter

drklee3:


x/incentive/keeper/rewards.go
```diff
@@ -86,3 +87,42 @@ func (k *Keeper) synchronizeClaim(
...
+		sortedSourceIDs = append(sortedSourceIDs, sourceID)
+	}
+
+	// Sort source IDs to ensure deterministic order of claim syncs
+	sort.Strings(sortedSourceIDs)
+
+	// Synchronize claim for each source ID
+	for _, sourceID := range sortedSourceIDs {
+		claim = k.synchronizeClaim(ctx, claim, sourceID, owner, accShares[sourceID])
+	}
```
rhuairahrighairidh, line 125: suggestion: I think it could be marginally safer to iterate through `sourceIDs` rather than the sorted keys. If an adapter has a bug and doesn't return all provided sourceIDs, this could silently skip a sync. Whereas if `sourceIDs` was iterated through and the map didn't contain the key this could panic and make the issues visible.

rhuairahrighairidh (approved):

PR #1389, opened 2022-11-10
mint community pool inflation in x/kavamint

pirtleshell:
Implements and tests `x/kavamint`'s `CommunityPoolInflation` parameter. Inflation is compounded for the number of seconds since the last block, and then coins are minted and transferred to the community pool (the module account of `x/community`)

Slightly modified staking reward implementation to ensure newly minted coins don't contribute the basis. (ie. if we mint coins and _then_ find the total supply, we'll be inflating more than expected because the total supply now includes the newly minted tokens.

### DIFF
--- a/x/kavamint/abci.go
+++ b/x/kavamint/abci.go
@@ -1,8 +1,11 @@
 package kavamint
 
 import (
+	"fmt"
+
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/kava-labs/kava/x/kavamint/keeper"
+	"github.com/kava-labs/kava/x/kavamint/types"
 )
 
 // BeginBlocker mints & distributes new tokens for the previous block.
@@ -11,9 +14,10 @@ func BeginBlocker(ctx sdk.Context, k keeper.Keeper) {
 	if !found {
 		previousBlockTime = ctx.BlockTime()
 	}
+	secondsPassed := ctx.BlockTime().Sub(previousBlockTime).Seconds()
 	// calculate totals before any minting is done to prevent new mints affecting the values
-	totalBonded := k.TotalBondedTokens(ctx)
 	totalSupply := k.TotalSupply(ctx)
+	totalBonded := k.TotalBondedTokens(ctx)
 
 	// ------------- Staking Rewards -------------
 	stakingRewardCoins, err := k.AccumulateStakingRewards(ctx, totalBonded, previousBlockTime)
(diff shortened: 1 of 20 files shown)


x/kavamint/abci.go
```diff
@@ -11,9 +11,12 @@ func BeginBlocker(ctx sdk.Context, k keeper.Keeper) {
 	if !found {
 		previousBlockTime = ctx.BlockTime()
 	}
+	// calculate totals before any minting is done to prevent new mints affecting the values
```
galxy25, line 14: does the ordering of the keepers in app.go ensure that x/mint's begin blocker is called before the staking modules?

x/kavamint/abci.go
```diff
@@ -29,8 +32,20 @@ func BeginBlocker(ctx sdk.Context, k keeper.Keeper) {
 	}
 
 	// ------------- Community Pool -------------
-	// TODO: mint tokens for community pool
-	// TODO: send tokens to community pool
+	communityPoolInflation, err := k.AccumulateCommunityPoolInflation(ctx, totalSupply, previousBlockTime)
+	if err != nil {
+		panic(err)
```
galxy25, line 37: nit: might be useful to include the totalSupply and previousBlockTime in this error message?

galxy25 (approved):
could see updating the spec further to supplement the code samples with a written description of what the desired behavior for the mint module is and adding more edge case tests but otherwise lgtm

x/kavamint/abci.go
```diff
@@ -11,9 +11,12 @@ func BeginBlocker(ctx sdk.Context, k keeper.Keeper) {
 	if !found {
 		previousBlockTime = ctx.BlockTime()
 	}
+	// calculate totals before any minting is done to prevent new mints affecting the values
+	totalBonded := k.TotalBondedTokens(ctx)
+	totalSupply := k.TotalSupply(ctx)
 
 	// ------------- Staking Rewards -------------
-	stakingRewardCoins, err := k.AccumulateStakingRewards(ctx, previousBlockTime)
+	stakingRewardCoins, err := k.AccumulateStakingRewards(ctx, totalBonded, previousBlockTime)
```
galxy25, line 19: an orthogonal concern to this pr, but I kinda think it makes since for the first begin blocker to calculate the time since the last block and pass that to all modules so that is consistent across modules

galxy25: (perhaps by setting it as value in the context...)

pirtleshell: i've refactored to consolidate `AccumulateStakingRewards` and `AccumulateCommunityPoolInflation` into a single `AccumulateInflation(ctx, rate, basis, secondsSinceLastMint)` method.

Now `secondsPassed` is calculated in the begin blocker and the individual inflation calculations look like:
```go
	stakingRewardCoins, err := k.AccumulateInflation(
		ctx, params.StakingRewardsApy, totalBonded, secondsPassed,
	)

	communityPoolInflation, err := k.AccumulateInflation(
		ctx, params.CommunityPoolInflation, totalSupply, secondsPassed,
	)
```

x/kavamint/abci.go
pirtleshell: yes, the ordering of the begin blockers matters. the kavamint's begin blocker must be registered before distribution so that the fee pool has the newly minted tokens to distribute. indicated the requirement in a comment where the begin blocker is registered.

https://github.com/Kava-Labs/kava/blob/18c41c2ef15fa59af9a8d363302dd8ad03b73747/app/app.go#L783-L785

PR #1391, opened 2022-11-15
Fix earn /vaults/{denom} query to allow denoms with slashes

drklee3:
This captures multiple path segments for earn /vaults/{denom=**} query.

Url percent-encoding does not seem to work so this is necessary.

Current endpoint failure:

https://api.data.kava.io/kava/earn/v1beta1/vaults/erc20/multichain/dai


rhuairahrighairidh (approved):
looks good
might need to do this for other endpoints in the future (hard/total_deposited has the same problem with ibc denoms)

PR #1395, opened 2022-11-18
Use different accumulator for earn

drklee3:
- Moves state methods to sub-package `store`
- New `accumulators` sub-package for different accumulators, Basic and Earn.


rhuairahrighairidh (approved):
I reviewed #1392, #1393, and #1396 together.

Keeping the bkava accumulation the same as master defintely helped review it, but I think it'd be good to simplify at some point. I imagine it's bit hard to onboard into right now. I have some ideas but that can be for future PRs.

nit: in `types/params.go` a line validating `RewardPeriods` could be added to `Params.Validate`

nit: looks like `accumulateBkavaEarnRewards` in `accumulators/earn.go` could use the adapters to get the total source shares

PR #1399, opened 2022-11-22
Refactor Buf Usage

nddeluca:
This removes custom protobuf scripts & docker image in favor of using buf directly.  In addition, clang-format is removed in favor of buf's built in formatting.

Tested `make proto-all` on several platforms, which make 3.81 and 4.3.

Includes CI checks to ensure all `make proto-all` tasks have been run.

There are also some makefile updates here that improve the versioning on master, in addition to defining more helper variables for tagging.


galxy25 (approved):

.github/workflows/proto.yml
```diff
@@ -0,0 +1,24 @@
+name: Protobuf Checks
```
galxy25, line 1: note to self to look into ways to re-use workflow jobs across workflows (e.g. being able to reference this proto check from the ci job versus having to copy and paste it into the ci workflow)

PR #1406, opened 2022-11-29
Add incentive migrations for earn rewards

drklee3:


x/incentive/migrations/v2/store.go
```diff
@@ -0,0 +1,142 @@
...
+	oldStore := prefix.NewStore(store, EarnClaimKeyPrefix)
+	iterator := sdk.KVStorePrefixIterator(oldStore, []byte{})
+	defer iterator.Close()
+	for ; iterator.Valid(); iterator.Next() {
+		var c types.EarnClaim
+		cdc.MustUnmarshal(iterator.Value(), &c)
+
+		if err := c.Validate(); err != nil {
+			return fmt.Errorf("invalid v1 EarnClaim: %w", err)
+		}
```
rhuairahrighairidh, line 44: Nice validations. I see what you mean about the tests not being very reassuring. Given mainnet state could be buggy this kind of defensive assumption checking is probably the way to go.

rhuairahrighairidh (changes requested):
The mainnet release (on releases/v0.19.x) has a migration added for the last uprade, so this will need to use consensus version 3. The upgrade module stores all the versions that should be running: https://api.kava.io/cosmos/upgrade/v1beta1/module_versions

rhuairahrighairidh (approved):

PR #1408, opened 2022-11-30
Update earn rewards claiming and hooks to use new methods

drklee3:


rhuairahrighairidh (approved):
looks good, we'll need to add the migration handler to `app.go` at some point, but the planned mainnet migration test will resurface that task.

PR #1414, opened 2022-12-02
Add savings incentives and fix hooks

drklee3:
- Fix savings incentive hooks
- Add savings hooks test
- Add new incentive savings adapter


x/savings/types/expected_keepers.go
```diff
@@ -29,7 +29,7 @@ type AccountKeeper interface {
 // SavingsHooks event hooks for other keepers to run code in response to Savings modifications
 type SavingsHooks interface {
 	AfterSavingsDepositCreated(ctx sdk.Context, deposit Deposit)
-	BeforeSavingsDepositModified(ctx sdk.Context, deposit Deposit, incomingDenoms []string)
+	BeforeSavingsDepositModified(ctx sdk.Context, deposit Deposit, newDenoms []string)
```
rhuairahrighairidh, line 32: nit: if we removed the `Deposit` type from the hook (and just pass address and coins), it would save implementers having to import that type. Keeps modules a little bit more decoupled

rhuairahrighairidh (approved):
The changes to sync/init look good. Nice addition of savings hooks tests.
nit: could the old savings sync code be removed?

drklee3:
Deleted old accumulator, leaving in types and keeper methods for now to be removed all together later

PR #1416, opened 2022-12-05
remove x/mint

pirtleshell:
x/mint's functionality has been replaced by x/kavamint.

This PR removes the x/mint module from the kava App.

Besides removing from module registration, the one place outside of tests that used x/mint was x/incentive's query for `GetStakingAPR`. this method is responsible for adding the staking reward contribution from incentive rewards to the base staking rate. originally it calculated the base rate from the inflation - community tax, adjusted to the bonded token ratio. now staking rate is set explicitly in the params of x/kavamint.

The tests of the incentive rewards needed to be slightly modified to adjust for the fact that minting is now done in the begin blocker of x/kavamint instead of x/mint.

All other changes are to update tests that were manually minting funds to accounts or using `simapp.Fund*` funcs. These were updated to use an existing test helper.


app/app.go
```diff
@@ -49,9 +49,6 @@ import (
 	"github.com/cosmos/cosmos-sdk/x/gov"
```
galxy25, line 49: do we need to add / update the kava v20 migration handler to delete the mint module?

x/earn/testutil/suite.go
```diff
@@ -205,7 +204,7 @@ func (suite *Suite) GetEvents() sdk.Events {
 // AddCoinsToModule adds coins to the earn module account
 func (suite *Suite) AddCoinsToModule(amount sdk.Coins) {
 	// Does not use suite.BankKeeper.MintCoins as module account would not have permission to mint
-	err := simapp.FundModuleAccount(suite.BankKeeper, suite.Ctx, types.ModuleName, amount)
+	err := suite.App.FundModuleAccount(suite.Ctx, types.ModuleName, amount)
```
galxy25, line 207: lots of warm and fuzzies, deleting code is nice but whats great is when you can simplify an interface, 👨🏽‍🍳 💋

x/incentive/types/expected_keepers.go
```diff
@@ -94,9 +93,9 @@ type AccountKeeper interface {
 	GetModuleAccount(ctx sdk.Context, name string) authtypes.ModuleAccountI
 }
 
-// MintKeeper defines the required methods needed by this modules keeper
-type MintKeeper interface {
-	GetMinter(ctx sdk.Context) (minter minttypes.Minter)
+// KavamintKeeper defines the required methods needed by this modules keeper
+type KavamintKeeper interface {
+	GetStakingApy(ctx sdk.Context) sdk.Dec
```
galxy25, line 98: nit: is GetStakingApy a method on the kavaminter? It's generally more generic / powerful to return a true interface here versus trying to make golang bend over to create a bloated java class object with getters and setters....

galxy25 (approved):

x/incentive/types/expected_keepers.go
pirtleshell: agreed. sadly though, getters like this appear to be the standard pattern for sharing values between modules (note that this interface is the kavamint keeper expected by x/incentive). i think the primary reason is because the actual value is dependent on context (block height to determine the correct parameter), not just some fixed value.

i only created getter-like methods on keepers for values that needed to be access by external modules like this one. the original x/mint GetMinter is a similar vein. as well as x/distribution GetCommunityTax used below.

it is a bit sad to write go like java :joy: :sob: but the pattern is strongly represented and is effective.

i suppose some benefits are more complex mocking powers for tests and the future allowance of `GetStakingApy` to be a more complex calculation. for now it's contents are indeed just `GetParams(ctx).StakingRewardsApy`...

x/incentive/keeper/querier.go
```diff
@@ -422,18 +422,8 @@ func queryGetAPYs(ctx sdk.Context, req abci.RequestQuery, k Keeper, legacyQuerie
...
-	bondedTokens := k.stakingKeeper.TotalBondedTokens(ctx)
-	circulatingSupply := k.bankKeeper.GetSupply(ctx, types.BondDenom)
-
-	// Staking APR = (Inflation Rate * (1 - Community Tax)) / (Bonded Tokens / Circulating Supply)
-	stakingAPR := inflationRate.
-		Mul(sdk.OneDec().Sub(communityTax)).
-		Quo(bondedTokens.ToDec().
-			Quo(circulatingSupply.Amount.ToDec()))
+	// Get staking APR
+	stakingAPR := k.kavamintKeeper.GetStakingApy(ctx)
```
nddeluca, line 426: Nice I like incentive not having to do the calculation here

x/incentive/types/expected_keepers.go
nddeluca: Naming aside, encapsulating the kavamint parameter structure is good here.  I'd rather add to the interface than couple external modules to the kava mint parameter structure.

x/kavamint/keeper/keeper.go
```diff
@@ -67,6 +67,12 @@ func (k Keeper) SetParams(ctx sdk.Context, params types.Params) {
 	k.paramSpace.SetParamSet(ctx, &params)
 }
 
+// GetStakingApy returns the APY minted for staking rewards
+func (k Keeper) GetStakingApy(ctx sdk.Context) sdk.Dec {
```
nddeluca, line 71: It's a simple method, but imo all public methods should have a unit test

nddeluca (approved):
just minor nits

PR #1419, opened 2022-12-07
Add hard supply/borrow adapter

drklee3:
Hooks will be updated in a later pr

### DIFF
--- a/x/incentive/keeper/adapters/hard_borrow/adapter.go
+++ b/x/incentive/keeper/adapters/hard_borrow/adapter.go
@@ -26,7 +26,16 @@ func (f SourceAdapter) TotalSharesBySource(ctx sdk.Context, sourceID string) sdk
 		return sdk.ZeroDec()
 	}
 
-	return coins.AmountOf(sourceID).ToDec()
+	totalBorrowed := coins.AmountOf(sourceID).ToDec()
+
+	interestFactor, found := f.keeper.GetBorrowInterestFactor(ctx, sourceID)
+	if !found {
+		// assume nothing has been borrowed so the factor starts at it's default value
+		interestFactor = sdk.OneDec()
+	}
+
+	// return borrowed/factor to get the "pre interest" value of the current total borrowed
+	return totalBorrowed.Quo(interestFactor)
 }
 
 func (f SourceAdapter) OwnerSharesBySource(
--- a/x/incentive/keeper/adapters/hard_borrow/adapter.go
+++ b/x/incentive/keeper/adapters/hard_borrow/adapter.go
@@ -1,6 +1,8 @@
 package hard_borrow
 
 import (
+	"fmt"
+
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
 	"github.com/kava-labs/kava/x/incentive/types"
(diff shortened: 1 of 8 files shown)


x/incentive/keeper/adapters/hard_borrow/adapter.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func (f SourceAdapter) TotalSharesBySource(ctx sdk.Context, sourceID string) sdk.Dec {
+	coins, found := f.keeper.GetBorrowedCoins(ctx)
+	if !found {
+		return sdk.ZeroDec()
+	}
+
+	return coins.AmountOf(sourceID).ToDec()
+}
```
rhuairahrighairidh, line 28: Hard has an additional complexity that the deposit amounts don't directly map to source shares.

A user's hard supplies and borrows technically increase every block due to interest, so they can't directly be used as source shares (otherwise we'd have to call sync every block). Similar to incentive the interest is not added to the store objects every block, it's calculated retroactively when a user deposits/withdraws/borrows/repays, so `Borrow.Amount` isn't the precise borrowed amount (like how the `Claim.Rewards` field isn't the actual claim's current reward balance).

The current code instead uses the 'normalized' borrow/deposit amounts as source shares:

https://github.com/Kava-Labs/kava/blob/aa5e9f63823575114ff7e3817aa9b299f3293cff/x/incentive/keeper/rewards_borrow.go#L39-L64
https://github.com/Kava-Labs/kava/blob/aa5e9f63823575114ff7e3817aa9b299f3293cff/x/incentive/keeper/rewards_borrow.go#L95-L96

x/incentive/keeper/adapters/hard_borrow/adapter_test.go
```diff
@@ -0,0 +1,285 @@
...
+
+	err = hardKeeper.Borrow(
+		suite.ctx,
+		suite.addrs[1],
+		sdk.NewCoins(
+			sdk.NewCoin(suite.denomA, sdk.NewInt(250)),
+		),
+	)
+	suite.NoError(err)
+
```
rhuairahrighairidh, line 168: I think it's worth testing the case where `NormalizedBorrow` differs from the borrow amount. If we don't use `NormalizedBorrow` in the adapter the tests won't catch it.

This case happens when the borrow.Index is greater than 1. Which happens when a borrow is modified after interest has accrued. Bit like how a claim's indexes increase when Sync is called after the global indexes have increased.

Should be able to trigger it by:
- borrowing some amount
- stepping through a few blocks to accrue interest on the borrow (interest accrues in hard begin blocker)
- modifying the borrow (borrow or repay) to sync it, therby updating the `Index` field

x/incentive/keeper/adapters/hard_borrow/adapter_test.go
```diff
@@ -0,0 +1,285 @@
...
+	err = hardKeeper.Borrow(
+		suite.ctx,
+		suite.addrs[0],
+		sdk.NewCoins(
+			sdk.NewCoin(suite.denomA, sdk.NewInt(100)),
+		),
+	)
+	suite.NoError(err)
+
```
rhuairahrighairidh, line 249: Similar to the owner shares test, I think we should test the case where `GetBorrowedCoins` differs from the total source share amount.

Should be able to tigger by:
- borrowing
- stepping through a few blocks to accumulate interest

At that point `GetBorrowedCoins` will include interest, but the `TotalSharesBySource` amount shouldn't.

rhuairahrighairidh (changes requested):

x/incentive/keeper/adapters/hard_borrow/adapter_test.go
```diff
@@ -0,0 +1,400 @@
...
+		name         string
+		giveSourceID string
+		wantShares   sdk.Dec
+	}{
+		{
+			"total shares",
+			suite.denomA,
+			// 40_000 + 25_000 + 1_000 + interest
+			// 66_000 + interest
+			sdk.NewDecWithPrec(66002, 0),
```
rhuairahrighairidh, line 384: We'll need to exclude interest from the total shares as well. Then the `normalized_owner_borrow / normalized_total_borrows` should be fraction of the borrows a user owns.
The old normalisation code: https://github.com/Kava-Labs/kava/blob/cb9e3da74aef12149a9c9a3746312aab304f46d2/x/incentive/keeper/rewards_borrow.go#L39-L64

PR #1420, opened 2022-12-07
add x/mint QueryServer for backwards compatibility

pirtleshell:
adds a replacement of `/comsos/mint` endpoints & queries. only `inflation` endpoint is implemented.

this endpoint is used to determine staking apr on third party platforms. to not break them `inflation` returns an adjusted value that will make the staking apr calculation correct.

the assumed calculation: 
$$StakingAPY = (Inflation - CommunityTax) * \frac{TotalSupply}{TotalBonded}$$

the community tax (`x/distribution` param) is set to zero. thus, the x/mint inflation endpoint returns
$$Inflation = StakingAPY * \frac{TotalBonded}{TotalSupply}$$

### DIFF
--- a/x/kavamint/keeper/grpc_query.go
+++ b/x/kavamint/keeper/grpc_query.go
@@ -3,7 +3,12 @@ package keeper
 import (
 	"context"
 
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	minttypes "github.com/cosmos/cosmos-sdk/x/mint/types"
+
 	"github.com/kava-labs/kava/x/kavamint/types"
 )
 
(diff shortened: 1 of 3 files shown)


galxy25 (approved):

nddeluca (approved):
nice addition 👍

x/kavamint/keeper/grpc_query.go
```diff
@@ -24,3 +29,57 @@ func (k Keeper) Inflation(c context.Context, _ *types.QueryInflationRequest) (*t
...
+	return MintQueryServer{kavamintKeeper}
+}
+
+var _ minttypes.QueryServer = MintQueryServer{}
+
+// Params is not implemented. There is no mint module.
+func (MintQueryServer) Params(
+	_ context.Context, _ *minttypes.QueryParamsRequest,
+) (*minttypes.QueryParamsResponse, error) {
+	return nil, status.Error(codes.Unimplemented, "x/mint has been replaced by x/kavamint")
```
nddeluca, line 51: Like the additional info pointing callers to the kavamint info.  This could also be extracted to a method since it's duplicated, but OK to leave since it's only used twice.  Likely not worth the extra layer of abstraction

PR #1422, opened 2022-12-08
Nd kavamint touchups

nddeluca:
draft for review -- will rebase into feat branch instead of merging


client/docs/config.json
```diff
@@ -216,6 +216,23 @@
         ]
       }
     },
+    {
+      "url": "./out/swagger/kava/kavamint/v1beta1/query.swagger.json",
```
nddeluca, line 220: This is required for the swagger to be generated

proto/kava/kavamint/v1beta1/kavamint.proto
```diff
@@ -1,19 +1,25 @@
...
 option go_package = "github.com/kava-labs/kava/x/kavamint/types";
 
 // Params wraps the governance parameters for the kavamint module
 message Params {
+  option (gogoproto.goproto_stringer) = false;
+  option (gogoproto.goproto_getters) = false;
+
   // yearly inflation of total token supply minted to the community pool.
-  bytes community_pool_inflation = 1 [
+  string community_pool_inflation = 1 [
```
nddeluca, line 15: sdk.Dec should be strings - https://github.com/cosmos/cosmos-sdk/issues/10863

proto/kava/kavamint/v1beta1/query.proto
```diff
@@ -11,7 +12,7 @@ option go_package = "github.com/kava-labs/kava/x/kavamint/types";
 service Query {
   // Params queries the parameters of x/kavamint module.
   rpc Params(QueryParamsRequest) returns (QueryParamsResponse) {
-    option (google.api.http).get = "/kava/kavamint/v1beta1/parameters";
+    option (google.api.http).get = "/kava/kavamint/v1beta1/params";
```
nddeluca, line 15: standard is `params`, though not all modules have followed it

x/kavamint/abci.go
```diff
@@ -12,8 +12,8 @@ import (
 func BeginBlocker(ctx sdk.Context, k keeper.KeeperI) {
 	params := k.GetParams(ctx)
 	// determine seconds since last mint
-	previousBlockTime, found := k.GetPreviousBlockTime(ctx)
-	if !found || previousBlockTime.IsZero() {
+	previousBlockTime := k.GetPreviousBlockTime(ctx)
```
nddeluca, line 15: We just need to protect the zero case -- found is redundant since we have zero value

x/kavamint/genesis.go
```diff
@@ -4,30 +4,35 @@ import (
...
+func InitGenesis(ctx sdk.Context, keeper keeper.Keeper, ak types.AccountKeeper, gs *types.GenesisState) {
+	// guard against invalid genesis
+	if err := gs.Validate(); err != nil {
+		panic(fmt.Sprintf("failed to validate %s genesis state: %s", types.ModuleName, err))
 	}
-	keeper.SetPreviousBlockTime(ctx, data.PreviousBlockTime)
 
-	if macc := ak.GetModuleAccount(ctx, types.ModuleName); macc == nil {
+	// get module account -- creates one with allowed permissions if it does not exist
+	macc := ak.GetModuleAccount(ctx, types.ModuleName)
```
nddeluca, line 20: This actually creates the module account, so the panic is just a double check incase that behavior changes

x/kavamint/genesis.go
```diff
@@ -4,30 +4,35 @@ import (
...
+	if macc == nil {
 		panic(fmt.Sprintf("%s module account has not been set", types.ModuleName))
 	}
+
+	// check module account has minter permissions
+	if !macc.HasPermission(authtypes.Minter) {
```
nddeluca, line 26: Good to check required permissions as well

x/kavamint/genesis.go
```diff
@@ -4,30 +4,35 @@ import (
...
 
 // ExportGenesis returns a GenesisState for a given context and keeper.
 func ExportGenesis(ctx sdk.Context, keeper keeper.Keeper) *types.GenesisState {
-	params := keeper.GetParams(ctx)
-	previousBlockTime, found := keeper.GetPreviousBlockTime(ctx)
-	if !found || previousBlockTime.IsZero() {
-		previousBlockTime = ctx.BlockTime()
-	}
-	return types.NewGenesisState(params, previousBlockTime)
+	return types.NewGenesisState(keeper.GetParams(ctx), keeper.GetPreviousBlockTime(ctx))
```
nddeluca, line 37: No modifications to genesis on export.

We test that an imported genesis always results in the same exported genesis -- best to make modifications outside of abci

x/kavamint/keeper/keeper.go
```diff
@@ -152,21 +146,30 @@ func (k Keeper) CumulativeInflation(ctx sdk.Context) sdk.Dec {
...
 	if err := blockTime.UnmarshalBinary(b); err != nil {
 		panic(err)
 	}
-	return blockTime, true
+
+	return
 }
 
 // SetPreviousBlockTime set the time of the previous block
 func (k Keeper) SetPreviousBlockTime(ctx sdk.Context, blockTime time.Time) {
```
nddeluca, line 165: We delete the key if the previous block time is blank

x/kavamint/types/genesis.go
```diff
@@ -1,10 +1,14 @@
 package types
 
 import (
-	fmt "fmt"
 	time "time"
 )
 
+var (
+	// DefaultPreviousBlockTime represents a time that is unset -- no previous block has occured
+	DefaultPreviousBlockTime = time.Time{}
```
nddeluca, line 9: change of default to assume no previous block and no accumulation on the first height after initgenesis

@pirtleshell we can set a previous block time explicity in the upgrade handler if we want

x/kavamint/types/genesis.go
```diff
@@ -24,11 +28,5 @@ func DefaultGenesisState() *GenesisState {
...
 // expected invariants holds.
 func (gs GenesisState) Validate() error {
-	if err := gs.Params.Validate(); err != nil {
-		return err
-	}
-	if gs.PreviousBlockTime.IsZero() {
-		return fmt.Errorf("previous block time not set")
-	}
-	return nil
+	return gs.Params.Validate()
```
nddeluca, line 31: Unset is a valid case for new chains since there are no previous blocks

x/kavamint/types/genesis_test.go
```diff
@@ -1,95 +1,149 @@
...
+				assert.Nil(t, err)
 			} else {
-				require.Error(t, err)
+				assert.EqualError(t, err, tc.expectedErr)
 			}
 		})
 	}
 }
+
+func TestGenesis_JSONEncoding(t *testing.T) {
```
nddeluca, line 109: These json tests help serve as examples and prevent regressions in protobuf encoding/tags

x/kavamint/types/params.go
```diff
@@ -56,29 +54,34 @@ func (p *Params) Validate() error {
 	return validateStakingRewardsApy(p.StakingRewardsApy)
 }
 
+// String implements fmt.Stringer
+func (p Params) String() string {
+	return fmt.Sprintf(`Params:
+	CommunityPoolInflation: %s
```
nddeluca, line 60: Just a helpful change -- protobuf Stringer() doesn't show sdk.Dec with a decimal place

x/kavamint/types/params.go
```diff
@@ -56,29 +54,34 @@ func (p *Params) Validate() error {
...
-// validateRateWithinBounds ensure that the given rate falls within the allowed bounds: [0, MaxMintingRate]
-func validateRateWithinBounds(rate sdk.Dec) error {
-	if rate.BigInt().Sign() == -1 {
-		return fmt.Errorf("rate must be >= 0")
-	}
-	if MaxMintingRate.LT(rate) {
-		return fmt.Errorf("rate out of bounds. the max allowed rate is %s", MaxMintingRate)
+// validateRate ensures rate is properly initialized (non-nil), not negative, and not greater than the max rate
+func validateRate(rate sdk.Dec) error {
+	if rate.IsNil() || rate.IsNegative() || rate.GT(MaxMintingRate) {
```
nddeluca, line 83: It's important to check that sdk.Dec values are not nill as they will panic if somehow set

x/kavamint/keeper/querier.go
```diff
@@ -1,48 +0,0 @@
-package keeper
-
```
nddeluca, line 2: Removed -- the legacy querier paths are not needed for new modules

x/kavamint/types/params.go
```diff
@@ -2,36 +2,33 @@ package types
...
+	// default inflation values are zero
 	DefaultCommunityPoolInflation = sdk.ZeroDec()
 	DefaultStakingRewardsApy      = sdk.ZeroDec()
 
-	// rates larger than 17,650% are out of bounds
-	// this is due to the necessary conversion of yearly rate to per second rate
-	// TODO consider lowering max rate. when it's this large the precision is very bad.
-	MaxMintingRate = sdk.NewDecWithPrec(1765, 1)
+	// MaxMintingRate returns the per second rate equivalent to 10,000% per year
+	MaxMintingRate = sdk.MustNewDecFromStr("0.000000146028999310")
```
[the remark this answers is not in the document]
nddeluca: Fixed to be the yearly value -- I got confused with the per second rates used in other modules

x/kavamint/types/genesis_test.go
```diff
@@ -1,95 +1,149 @@
...
 			types.NewGenesisState(
 				types.NewParams(sdk.ZeroDec(), sdk.ZeroDec()),
 				time.Now(),
 			),
-			true,
+			"",
 		},
 		{
-			"invalid - no time set",
+			"vali - no time set",
```
pirtleshell, line 54:
```suggestion
			"valid - no time set",
```

x/kavamint/types/keys.go
```diff
@@ -12,11 +12,7 @@ const (
...
 	QuerierRoute = StoreKey
-
-	// Query endpoints supported by kavamint
-	QueryParameters = "parameters"
-	QueryInflation  = "inflation"
 )
 
 // PreviousBlockTimeKey is the store key for the previous block time
-var PreviousBlockTimeKey = []byte{0x00}
+var PreviousBlockTimeKey = []byte{0x01}
```
pirtleshell, line 18: start w/ 1 to prevent confusion with nil value?

x/kavamint/genesis.go
nddeluca: I wrote it both ways, but thought this was cleaner and easier to read without the permission check nested.

nddeluca: Could go either way on it, basically just increased the scope of `macc` to use below

x/kavamint/types/keys.go
nddeluca: hmm I think I fat fingered a key here, didn't mean to modify it

PR #1423, opened 2022-12-09
Proto & types for x/community lend proposals

pirtleshell:
Adds the proto and types for two new gov proposals:
`CommunityPoolLendDepositProposal`: will deposit community pool funds into a lend position
`CommunityPoolLendWithdrawProposal`: will withdraw community pool funds from a lend position

Separating this from the actual registration & implementation for smaller, more easily reviewed PRs.


nddeluca (approved):
lgtm, nice test coverage. I like the ErrorContains method -- much easier to prevent false positives if code is refactored and the test case becomes invalid for another reason.

PR #1424, opened 2022-12-12
Add MsgEVMCall to x/evmutil

DracoLi:
Same as https://github.com/Kava-Labs/kava/pull/1405 but with a cleaned-up git history.


galxy25:
# tldr;
	
@DracoLi @rhuairahrighairidh @nddeluca Can we lock down the behavior of this feature to only what is explicitly needed (using community pool funds) with support (via governance modifiable whitelists) for future use cases to only be enabled once they are defined and audited.

# Requested Changes
- Limit via api and protocol level validations what addresses can be used for the authority or from_address fields in CommunityPoolProposal messages 
- Limit via api and protocol level validations what contracts can be used for the to fields in CommunityPoolProposal messages related to evm contracts
- For the above validations, support a whitelist (modifiable via governance proposal) for 
	- what EVM contracts can be called
	- what addresses can be used to move funds from

The above requests are based on the below research, open to any feedback on where I missed / misunderstood something.

## Governance Background
	
- Anyone can submit a governance proposal
- Proposal will be enacted if
	- at least 40% of possible votes have been cast
	- at least 50% of cast votes are yes
	- less than 34% of cast votes are “no with veto”
	- voting period has ended
	- all transactions for messages contained in the proposal execute without error
	- https://github.com/cosmos/cosmos-sdk/tree/main/x/gov#readme

## Prior Feature Work

- Funds from the community pool can be sent to arbitrary kava addresses (e.g. for paying out Kava Rise rewards)
	- https://kava-labs.atlassian.net/wiki/spaces/PROD/pages/1187545089/Kava+Rise+Payout+Process
	- https://www.mintscan.io/kava/proposals/106

## New Product Requirements

- Funds from the community pool can be used to interact with arbitrary evm contracts via a governance proposal
- Funds from the community pool can be used to make arbitrary cosmos sdk calls via a governance proposal

## Feature Functionality (as implemented or proposed)
	
- Funds from any account can be used to interact with arbitrary evm contracts via a governance proposal
- Funds from any account can be used to make arbitrary cosmos sdk calls via a governance proposal
- https://kava-labs.atlassian.net/wiki/spaces/ENG/pages/1154940936/Spike+Enable+contract+calls+via+CommunityPool+Proposal

## Attack Surface

- Either via an inadvertent error, or from a purposefully malicious proposer, non community pool funds can be accessed via a governance proposal
- Undefined or undesired behavior as a result of unforeseen interactions between a chain of message calls

## Defense Mechanisms
- For kava mainnet, the voting period is 1 week and kava labs has veto proof voting power (i.e. if all kava labs voting power is a yes for a proposal, it will always pass): this enables kava labs to prevent a malicious or flawed proposal (e.g. typo in the from or to address of a bank send message) AS LONG AS the malicious or flawed proposal is spotted and voted no on by kava labs within 7 days of the proposal being submitted.

While the above is a big mitigating factor, humans make errors, and if we can do things to prevent those errors in the first place or minimize the magnitude of an error, that's a delicious recipe for more secure production software. Following the model of least privilege and making explicit the assumptions and desired power of this feature I think is warranted given the potential scope of the feature: ANY account can submit a proposal that can call ANY contract, ANY sdk message, using funds from ANY account. All those *'s could one leave one seeing stars, and not the cosmic kind 😅

image

app/app.go
```diff
@@ -501,6 +501,7 @@ func NewApp(
 		evmutilSubspace,
 		app.bankKeeper,
 		app.accountKeeper,
+		authtypes.NewModuleAddress(communitytypes.ModuleAccountName).String(),
```
galxy25, line 504: does this mean the evm module will only be able to modify funds from the community pool? Can this only ever be one address / is it possible to modify this authorization as part of a governance proposal?

proto/kava/evmutil/v1beta1/tx.proto
```diff
@@ -50,3 +53,28 @@ message MsgConvertERC20ToCoin {
...
+  string data = 3;
+
+  // amount defines the integer value of the transaction amount.
+  string amount = 4 [
+    (cosmos_proto.scalar) = "cosmos.Int",
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+    (gogoproto.nullable) = false
+  ];
+
+  // authority is the address of the account that must be the signer.
```
galxy25, line 75: Proposal submitted
Proposal contains a CommunityPoolProposal message
CommunityPoolProposal message contains an evm call 

So for these messages the signer should always be the community pool address? 

If it's always the same address, do we need to / why should we make the user always supply this?

If it's always the same address, are we validating at the protocol and api layer that this address is the expected one (i.e. the community module)?

x/evmutil/keeper/erc20.go
```diff
@@ -19,34 +19,32 @@ const (
 	erc20BalanceOfMethod = "balanceOf"
 )
 
-// DeployTestMintableERC20Contract deploys an ERC20 contract on the EVM as the
+// DeployTestERC20Contract deploys an ERC20 contract on the EVM as the
```
galxy25, line 22: curious why "Test" is in the function name?

x/evmutil/keeper/evm.go
```diff
@@ -48,7 +48,7 @@ func (k Keeper) CallEVM(
 		)
 	}
 
-	resp, err := k.CallEVMWithData(ctx, from, &contract, data)
+	resp, err := k.CallEVMWithData(ctx, from, &contract, data, big.NewInt(0))
```
galxy25, line 51: here we hardcode this to 0 because this evm call doesn't involve sending any funds but just triggering some action on the contract?

x/evmutil/keeper/evm.go
```diff
@@ -122,11 +123,15 @@ func (k Keeper) CallEVMWithData(
 
 	res, err := k.evmKeeper.ApplyMessage(ethGasContext, msg, evmtypes.NewNoOpTracer(), true)
 	if err != nil {
+		if len(res.Ret) > 0 {
```
galxy25, line 126: for troubleshooting scenerios might be useful to debug log the params passed to `ApplyMessage` in this error branch

x/evmutil/keeper/keeper.go
```diff
@@ -20,6 +20,10 @@ type Keeper struct {
 	bankKeeper    types.BankKeeper
 	evmKeeper     types.EvmKeeper
 	accountKeeper types.AccountKeeper
+
+	// the address capable of executing a MsgEVMCall message. Typically, this
+	// should be the x/community module account.
+	authority string
```
galxy25, line 26: making this a slice now could be useful for future proofing if / when we want to expand the list of authorities

x/evmutil/keeper/keeper.go
```diff
@@ -40,9 +45,15 @@ func NewKeeper(
 		paramSubspace: params,
 		bankKeeper:    bk,
 		accountKeeper: ak,
+		authority:     authority,
 	}
 }
 
+// GetAuthority returns the x/evmutil module's authority.
```
galxy25, line 52: is this message used anywhere? Typically for golang you want to avoid java styler getter and setters and only interface the behavior of the desired type...e.g. there might be a method on the interface or concrete type "IsValidAuthority" to allow callers to let the keeper determine validity without exposing its state. Or if the state does need to get exposed, that can more directly be stored in a public field on the struct that can be accessed directly

x/evmutil/keeper/msg_server.go
```diff
@@ -102,3 +106,66 @@ func (s msgServer) ConvertERC20ToCoin(
 
 	return &types.MsgConvertERC20ToCoinResponse{}, nil
 }
+
+// MsgEVMCall executes the msg data again the go-ethereum EVM.
```
galxy25, line 110: `against the`

x/evmutil/keeper/msg_server.go
```diff
@@ -102,3 +106,66 @@ func (s msgServer) ConvertERC20ToCoin(
...
+func (k msgServer) EVMCall(goCtx context.Context, msg *types.MsgEVMCall) (*types.MsgEVMCallResponse, error) {
+	expMsgAuthority := k.keeper.GetAuthority()
```
galxy25, line 112: could validate this at the api layer to prevent submission of flawed proposals earlier (in addition to this protocol level check)

and ah now I see where GetAuthority is being called! Definitely would recommend making the code more idiomatic and well encapsulated by creating a "ValidateEVMCall" on the keeper for this and the below validations

x/evmutil/keeper/msg_server.go
```diff
@@ -102,3 +106,66 @@ func (s msgServer) ConvertERC20ToCoin(
...
+	}
+	_, err = k.keeper.CallEVMWithData(
+		ctx,
+		authorityEvmAddr,
+		&toAddr,
+		data,
+		amt,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("evm call failed: %w", err)
```
galxy25, line 152: for troubleshooting might be useful to debug log the params passed to `CallEVMWithData` above here

x/evmutil/spec/03_messages.md
```diff
@@ -63,3 +63,40 @@ message MsgConvertCoinToERC20 {
...
+  ];
+
+  // authority is the address of the account that must be the signer.
+  string authority = 5 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+}
+```
+
+### State Changes
+
+- EVM is called from the msg authority's 0x address with the provided data and amount. The msg authority must be the module's expected authority bech32 address. By default, this is the x/community module.
```
galxy25, line 102: saying "by default" makes me wonder if it can be changed from the default (and if so how)?

x/evmutil/types/msg.go
```diff
@@ -144,3 +154,94 @@ func (msg MsgConvertERC20ToCoin) Route() string {
...
+
+// Route implements Msg
+func (msg MsgEVMCall) Route() string { return types.RouterKey }
+
+// Type implements Msg
+func (msg MsgEVMCall) Type() string { return sdk.MsgTypeURL(&msg) }
+
+// ValidateBasic implements Msg
+func (msg MsgEVMCall) ValidateBasic() error {
+	if _, err := sdk.AccAddressFromBech32(msg.Authority); err != nil {
```
galxy25, line 166: here we could validate the the authority address is in the whitelist of addresses

galxy25 (changes requested):
see comments. Given the scope of the feature, and the time we have until it is scheduled to be launched, I think slowing down the implementation by making validations and configuration of what accounts, contracts and messages can be used as part of this feature will add significant future proofing and security wins.

proto/kava/evmutil/v1beta1/tx.proto
DracoLi: Answer to these are yes and is the intended behaviour. This method of handling government msgs aligns with the current expectations of cosmo-sdk. See https://github.com/cosmos/cosmos-sdk/blob/main/x/gov/keeper/keeper.go#L45

x/evmutil/keeper/msg_server.go
DracoLi: I don't think we should move `ValidateEVMCall` to the keeper. The context matter and this being here is more suited for this context. `ValidateEVMCall` in the EVMUtil keeper doesn't make sense to me. In addition, it aligns with the other implementations of governance messages which you can see [here](https://github.com/cosmos/cosmos-sdk/blob/main/x/gov/keeper/msg_server.go#L192)

x/evmutil/spec/03_messages.md
DracoLi: this is set in app.go

x/evmutil/types/msg.go
DracoLi: this is already validated in `msg_sever`. Basic validation cannot get data from state so its not possible to validate whitelists here.

x/evmutil/keeper/msg_server.go
DracoLi: 👍

x/evmutil/keeper/evm.go
DracoLi: i'll add the log to the msg_server since it will cover other errors from evm as well and avoids polluting logs from the evm calls from other places.

galxy25:
@DracoLi please ping me when ready for review after the discussed updates are implemented

x/evmutil/keeper/erc20.go
DracoLi: Refactored in https://github.com/Kava-Labs/kava/pull/1448

galxy25 (approved):

galxy25 (approved):

pirtleshell:
we'll remember you fondly, `MsgEVMCall`. maybe we'll see you again someday 👋

PR #1425, opened 2022-12-12
Add x/community Lend proposals

pirtleshell:
Implements two new proposals:
* `CommunityPoolDepositProposal` - deposits community pool funds into lend position
* `CommunityPoolWithdrawProposal` - withdraws community pool funds from lend position

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -607,6 +607,7 @@ func NewApp(
 		communitytypes.ModuleAccountName,
 	)
 
+	// x/community's deposit/withdraw to lend proposals depend on hard keeper.
 	app.communityKeeper = communitykeeper.NewKeeper(
 		app.accountKeeper,
 		app.bankKeeper,
--- a/app/app.go
+++ b/app/app.go
@@ -508,20 +508,6 @@ func NewApp(
 
 	app.evmutilKeeper.SetEvmKeeper(app.evmKeeper)
 
-	app.communityKeeper = communitykeeper.NewKeeper(
-		app.accountKeeper,
-		app.bankKeeper,
-	)
-	app.kavadistKeeper = kavadistkeeper.NewKeeper(
-		appCodec,
-		keys[kavadisttypes.StoreKey],
-		kavadistSubspace,
-		app.bankKeeper,
-		app.accountKeeper,
-		app.communityKeeper,
-		app.loadBlockedMaccAddrs(),
-	)
-
 	app.transferKeeper = ibctransferkeeper.NewKeeper(
 		appCodec,
 		keys[ibctransfertypes.StoreKey],
@@ -620,6 +606,22 @@ func NewApp(
 		&savingsKeeper,
 		communitytypes.ModuleAccountName,
 	)
+
+	app.communityKeeper = communitykeeper.NewKeeper(
+		app.accountKeeper,
+		app.bankKeeper,
+		hardKeeper,
+	)
+	app.kavadistKeeper = kavadistkeeper.NewKeeper(
+		appCodec,
+		keys[kavadisttypes.StoreKey],
+		kavadistSubspace,
+		app.bankKeeper,
+		app.accountKeeper,
+		app.communityKeeper,
+		app.loadBlockedMaccAddrs(),
+	)
+
 	app.kavamintKeeper = kavamintkeeper.NewKeeper(
 		appCodec,
 		keys[kavaminttypes.StoreKey],
@@ -694,6 +696,7 @@ func NewApp(
 		AddRoute(distrtypes.RouterKey, distr.NewCommunityPoolSpendProposalHandler(app.distrKeeper)).
 		AddRoute(kavadisttypes.RouterKey, kavadist.NewCommunityPoolMultiSpendProposalHandler(app.kavadistKeeper)).
 		AddRoute(earntypes.RouterKey, earn.NewCommunityPoolProposalHandler(app.earnKeeper)).
+		AddRoute(communitytypes.RouterKey, community.NewCommunityPoolProposalHandler(app.communityKeeper)).
 		AddRoute(committeetypes.RouterKey, committee.NewProposalHandler(app.committeeKeeper))
 	app.govKeeper = govkeeper.NewKeeper(
 		appCodec,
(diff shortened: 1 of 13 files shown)


app/app.go
```diff
@@ -620,6 +606,22 @@ func NewApp(
 		&savingsKeeper,
 		communitytypes.ModuleAccountName,
 	)
+
```
galxy25, line 609: nit: code comment for future readers on the required state / dependency order in case of future updates to that state / related state functions

app/app.go
```diff
@@ -1046,14 +1049,22 @@ func (app *App) RegisterTendermintService(clientCtx client.Context) {
...
+		app.accountKeeper.GetModuleAddress(kavadisttypes.FundModuleAccount).String(): true,
+		// community
+		app.accountKeeper.GetModuleAddress(communitytypes.ModuleAccountName).String(): true,
+	}
 
 	for addr := range modAccAddrs {
-		// Set the kavadist and earn module account address as unblocked
-		if addr == kavadistMaccAddr.String() || addr == earnMaccAddr.String() || addr == liquidMaccAddr.String() || addr == kavadistFundMaccAddr.String() {
+		// Set allowed module accounts as unblocked
+		if allowedMaccs[addr] {
```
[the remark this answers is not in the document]
galxy25: for ultimate safety recommend using the two value map entry test 

https://stackoverflow.com/questions/2050391/how-to-check-if-a-map-contains-a-key-in-go

```golang
if val, ok := myMap["foo"]; ok {
    //do something here
}

```

x/community/handler.go
```diff
@@ -0,0 +1,24 @@
...
+// NewCommunityPoolProposalHandler handles x/community proposals.
+func NewCommunityPoolProposalHandler(k keeper.Keeper) govtypes.Handler {
+	return func(ctx sdk.Context, content govtypes.Content) error {
+		switch c := content.(type) {
+		case *types.CommunityPoolLendDepositProposal:
+			return keeper.HandleCommunityPoolLendDepositProposal(ctx, k, c)
+		case *types.CommunityPoolLendWithdrawProposal:
+			return keeper.HandleCommunityPoolLendWithdrawProposal(ctx, k, c)
+		default:
+			return sdkerrors.Wrapf(sdkerrors.ErrUnknownRequest, "unrecognized community proposal content type: %T", c)
```
galxy25, line 21: would be useful to have an integration test for this flow to make sure the app doesn't panic, though less a concern if this handler is only called on a per request basis

x/community/testutil/main.go
```diff
@@ -30,6 +33,10 @@ func (suite *Suite) SetupTest() {
 	suite.App = tApp
 	suite.Ctx = ctx
 	suite.Keeper = tApp.GetCommunityKeeper()
+	communityPoolAddress := tApp.GetAccountKeeper().GetModuleAddress(types.ModuleAccountName)
+	// hello, greppers!
+	suite.Equal("kava17d2wax0zhjrrecvaszuyxdf5wcu5a0p4qlx3t5", communityPoolAddress.String())
```
galxy25, line 38: (I know this is copy pasted code but...can this value come from a constant that lives outside of test land?)

galxy25 (approved):

x/community/keeper/proposal_handler_test.go
```diff
@@ -0,0 +1,312 @@
...
+		proposals          []*types.CommunityPoolLendWithdrawProposal
+		expectedErr        string
+		expectedWithdrawal sdk.Coins
+	}{
+		{
+			// in the week it would take a proposal to pass, the position would have grown
+			// to withdraw the entire position, it'd be safest to set a very high withdraw
+			name:           "valid - requesting withdrawal of more than total will withdraw all",
+			initialDeposit: ukava(1e9),
+			proposals: []*types.CommunityPoolLendWithdrawProposal{
```
nddeluca, line 199: Nice test case

x/community/keeper/proposal_handler_test.go
```diff
@@ -0,0 +1,312 @@
...
+			initialDeposit: ukava(1e9).Add(usdx(1e9)...),
+			proposals: []*types.CommunityPoolLendWithdrawProposal{
+				{Amount: ukava(1e9)},
+				{Amount: usdx(1e9)},
+			},
+			expectedErr:        "",
+			expectedWithdrawal: ukava(1e9).Add(usdx(1e9)...),
+		},
+		{
+			name:           "valid - multiple proposals, partial withdrawal",
```
nddeluca, line 243: Really liking the number of different partial, denom, etc tests

x/community/keeper/proposal_handler_test.go
```diff
@@ -0,0 +1,312 @@
...
+			expectedErr:        "deposit not found",
+			expectedWithdrawal: sdk.NewCoins(),
+		},
+		{
+			name:           "invalid - nonexistent position, has deposits of different denom",
+			initialDeposit: ukava(1e8),
+			proposals: []*types.CommunityPoolLendWithdrawProposal{
+				{Amount: usdx(1e8)},
+			},
+			expectedErr:        "no coins of this type deposited",
```
nddeluca, line 267: Good error testing 👍

x/community/types/keys.go
```diff
@@ -7,6 +7,9 @@ const (
 	// ModuleAccountName is the name of the module's account
 	ModuleAccountName = ModuleName
 
+	// RouterKey is the top-level router key for the module
+	RouterKey = ModuleName
```
nddeluca, line 11: I think we can also now add this to module.go now that we have one

nddeluca (approved):

app/app.go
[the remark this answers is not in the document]
pirtleshell: i'm purposefully checking the value of the map. this correctly results in  `allowedMaccs[address]` returning false for both the case that it doesn't exist in the map, or it was improperly setup to be `false`.

should i check `ok && val`?

x/community/testutil/main.go
pirtleshell: this isn't copypasta, it was just me keeping the address in source control for the inevitable search. originally only added it to get the address for documentation. i left it b/c it felt useful.
the constant is calculated based off the `types.ModuleAccountName`. i don't think it makes sense to keep the string itself as a constant outside the tests.
as a test, this catches regressions in the module account name, or breaking changes in how macc addresses are determined.

PR #1430, opened 2022-12-15
Override original community-pool query service

pirtleshell:
`x/distribution` includes a query for `community-pool` that points to a balance that no longer represents community pool funds.

by registering our own distribution query service that overrides the community pool query (and forwards all other requests to the vanilla query server), we can return the expected balance. this eases third party integration with the following downside:  historical values of the original community pool can't be queried
to overcome this, we will add a `legacy-community-pool` query to `x/community` in a future PR.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -90,6 +90,7 @@ import (
 
 	"github.com/kava-labs/kava/app/ante"
 	kavaparams "github.com/kava-labs/kava/app/params"
+	kavadistrquery "github.com/kava-labs/kava/app/query/distribution"
 	"github.com/kava-labs/kava/x/auction"
 	auctionkeeper "github.com/kava-labs/kava/x/auction/keeper"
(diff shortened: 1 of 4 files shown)


app/app.go
```diff
@@ -971,6 +972,28 @@ func NewApp(
...
+		// skip registration of distribution services
+		if module.Name() == distrtypes.ModuleName {
+			continue
+		}
+		module.RegisterServices(cfg)
+	}
+
+	// register ditribution services except query server
+	distrtypes.RegisterMsgServer(cfg.MsgServer(), distrkeeper.NewMsgServerImpl(app.distrKeeper))
+	cfg.RegisterMigration(
```
DracoLi, line 987: we should add a note here that this needs to be updated whenever we update cosmos SDK since new migrations may be updated.

nddeluca (approved):
nice replacement

PR #1447, opened 2023-01-05
add community module to store upgrades

pirtleshell:
* fixes the mainnet upgrade handler (adds x/community to store upgrades)
* adds new v0.20.0-alpha.1 upgrade handler to make same change to public testnet

**NOTE: i'm not currently convinced this fixes the pruning nodes of public testnet. x/community has no store and so i would not expect it to be required in the store upgrades.** making this PR in case it is what is needed for the fix.

### DIFF
--- a/app/upgrades.go
+++ b/app/upgrades.go
@@ -17,7 +17,7 @@ import (
 
 const (
 	MainnetUpgradeName = "v0.20.0"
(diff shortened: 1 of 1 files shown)


pirtleshell:
not actually a problem and not a necessary upgrade.

PR #1448, opened 2023-01-09
Community Proposal with arbitrary messages support

DracoLi:
Implement new community pool proposal with the following features:
- Ability to include any Cosmos SDK messages
- Whitelist for allowed SDK messages to include in proposal
- Tests for using MsgEVMCall in community proposal for interacting with EVM
- Refactor EVM contracts used in x/evmutil. It is now moved into its own module
- CLI for submitting new community pool proposal


x/community/client/proposal_handler.go
```diff
@@ -11,16 +11,22 @@ import (
...
 		notImplementedRestHandler(types.ProposalTypeCommunityPoolLendDeposit),
 	)
+	// Deprecated: Use CommunityPoolProposal instead
 	LendWithdrawProposalHandler = govclient.NewProposalHandler(
 		cli.NewCmdSubmitCommunityPoolLendWithdrawProposal,
 		notImplementedRestHandler(types.ProposalTypeCommunityPoolLendDeposit),
 	)
+	ProposalHandler = govclient.NewProposalHandler(
+		cli.GetCmdSubmitCommunityPoolProposal,
+		notImplementedRestHandler(types.ProposalTypeCommunityPool),
```
galxy25, line 28: since there is a cli command for this, I assume this is implemented in the json RPC endpoint? Is there a reason why this isn't implemented in the rest or other non rpc transport apis (e.g. websocket or grpc), and where do we track / document which methods are implemented for a given transport?

x/community/client/utils/utils.go
```diff
@@ -37,3 +38,48 @@ func ParseCommunityPoolLendWithdrawProposal(
...
+
+		msgs[i] = msg
+	}
+
+	deposit, err := sdk.ParseCoinsNormalized(jsonProposal.Deposit)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	communityPoolProposal, err := types.NewCommunityPoolProposal(jsonProposal.Title, jsonProposal.Description, msgs)
```
galxy25, line 79: for best user experience / principle of least surprise recommend validating at proposal time here that the message not only parses - is syntactially correct, but also is semantically correct -e.g. that the authority is in the list of whitelisted addresses, and that the type of the message to be called is in the whitelist of messages that the community pool is authorized to execute.

x/community/genesis.go
```diff
@@ -11,9 +11,11 @@ import (
...
 
 // InitGenesis initializes the community module account
-func InitGenesis(ctx sdk.Context, k keeper.Keeper, ak types.AccountKeeper) {
+func InitGenesis(ctx sdk.Context, k keeper.Keeper, ak types.AccountKeeper, gs *types.GenesisState) {
 	// check if the module account exists
 	if moduleAcc := ak.GetModuleAccount(ctx, types.ModuleAccountName); moduleAcc == nil {
 		panic(fmt.Sprintf("%s module account has not been set", types.ModuleAccountName))
 	}
+
+	k.SetParams(ctx, gs.Params)
```
galxy25, line 20: assuming this will need a migration for public testnet and mainnet, @nddeluca is this an example of a "consensus" level migration that should be written and tested in the current PR?

x/community/keeper/grpc_query.go
```diff
@@ -26,6 +26,13 @@ func (s queryServer) Balance(c context.Context, _ *types.QueryBalanceRequest) (*
 	}, nil
 }
 
+func (s queryServer) Params(c context.Context, _ *types.QueryParamsRequest) (*types.QueryParamsResponse, error) {
```
galxy25, line 29: missing doc string for public method

x/community/keeper/keeper.go
```diff
@@ -31,12 +37,17 @@ func NewKeeper(ak types.AccountKeeper, bk types.BankKeeper, dk types.Distributio
 		panic("legacy community pool address not found")
 	}
 
-	return Keeper{
-		bankKeeper:    bk,
-		distrKeeper:   dk,
-		hardKeeper:    hk,
-		moduleAddress: addr,
+	if !params.HasKeyTable() {
```
galxy25, line 40: what scenarios would trigger this if block to happen / is that worth documenting for future developers with a comment?

x/community/keeper/proposal_handler.go
```diff
@@ -15,3 +19,62 @@ func HandleCommunityPoolLendDepositProposal(ctx sdk.Context, k Keeper, p *types.
 func HandleCommunityPoolLendWithdrawProposal(ctx sdk.Context, k Keeper, p *types.CommunityPoolLendWithdrawProposal) error {
 	return k.hardKeeper.Withdraw(ctx, k.moduleAddress, p.Amount)
 }
+
+// HandleCommunityPoolProposal is a handler for executing a passed community pool proposal
+func HandleCommunityPoolProposal(ctx sdk.Context, k Keeper, p *types.CommunityPoolProposal) error {
+	logger := k.Logger(ctx)
+
+	// attempt to execute all messages within the passed proposal
```
galxy25, line 27: 🥇 great comment that focuses on the why over the what

x/community/keeper/proposal_handler.go
```diff
@@ -15,3 +19,62 @@ func HandleCommunityPoolLendDepositProposal(ctx sdk.Context, k Keeper, p *types.
...
+	for idx, msg := range msgs {
+		if !isMsgEnabled(msg, enabledUrls) {
+			return logAndReturnProposalMsgError(logger, idx, msg, "msg not enabled via params")
+		}
+
+		// assert that the community module account is the only signer of the messages
+		signers := msg.GetSigners()
+		if len(signers) != 1 || !signers[0].Equals(k.moduleAddress) {
+			return logAndReturnProposalMsgError(logger, idx, msg, "invalid signer")
+		}
```
galxy25, line 47: still feel like this logic would be more cleanly encapsulated as a method on the msg type, then for each msg we would call a `Validate` method and it can return the appropriate error

(and this is another reason / value from validating the message signers are correct on submission, the earlier we can eject garbage / bad data from the system the more reliable and secure the system will be)

x/community/types/errors.go
```diff
@@ -0,0 +1,10 @@
+package types
+
+import (
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+var (
+	ErrInvalidProposal      = sdkerrors.Register(ModuleName, 1, "invalid community pool proposal")
```
galxy25, line 8: suggest additional errors for cases where authority is invalid or message to execute is not in the whitelist

x/community/keeper/proposal_handler.go
```diff
@@ -15,3 +19,62 @@ func HandleCommunityPoolLendDepositProposal(ctx sdk.Context, k Keeper, p *types.
...
+
+	enabledUrls := k.GetParams(ctx).EnabledProposalMsgUrls
+	for idx, msg := range msgs {
+		if !isMsgEnabled(msg, enabledUrls) {
+			return logAndReturnProposalMsgError(logger, idx, msg, "msg not enabled via params")
+		}
+
+		// assert that the community module account is the only signer of the messages
+		signers := msg.GetSigners()
+		if len(signers) != 1 || !signers[0].Equals(k.moduleAddress) {
```
galxy25, line 45: thought we talked about making this extensible for when this proposal message is used for infrastracture pool funds but it looks like here we aren't supporting being able to initialize the keeper with a slice of whitelisted signer addresses?

x/evmutil/contract/CustomERC20.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.4;
+
+import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+contract MyToken is ERC20, Ownable {
```
galxy25, line 7: are there any unit testing frameworks for solidity?

x/evmutil/keeper/keeper.go
```diff
@@ -49,6 +49,10 @@ func NewKeeper(
 	}
 }
 
+func (k Keeper) GetAccountKeeper() types.AccountKeeper {
```
galxy25, line 52: missing doc string for public method

galxy25 (approved):
approved pending resolution of comments

x/community/client/proposal_handler.go
DracoLi: this is not grpc but legacy rest, which is now deprecated.

x/community/genesis.go
DracoLi: yes, we will need a module level migration to set the default params and we can add some tests for this. I'll add it to the PR.

x/community/keeper/proposal_handler.go
DracoLi: this is intentional since the context here matters. This signer validation is only applicable here on the community proposal submission so I think its clearer to do that here. On a msg level, the msg can validate the signer only if the msgs should be ran for a specific signer (which is the case for only the EVMCallMsg).

x/community/keeper/proposal_handler.go
DracoLi: thats for the `EvmCallMsg`, which is done [here](https://github.com/Kava-Labs/kava/pull/1424/commits/022f5d7d82b616d994a127891be88d25ea4eab8d). On a proposal level, it needs to be module specific.

PR #1452, opened 2023-01-24
replace x/kavamint with x/mint

pirtleshell:
like the old times, only new.

* replaces x/kavamint with x/mint. updates all non-kavamint tests to use & expect x/mint.
* updated all non-kavamint tests & verified they pass via `go test $(go list ./... | grep -v /x/kavamint)`
* verified that newly init'd chain runs

Will delete x/kavamint code & regenerate proto files in follow-up PR.


nddeluca (approved):
looks good, OK to ignore the proto errors for backwards compatibility since there are no messages in kavamint.

App.go dependencies look OK too.

PR #1453, opened 2023-01-24
delete all x/kavamint code

pirtleshell:


app/test_common.go
```diff
@@ -118,9 +117,6 @@ func (tApp TestApp) GetFeeMarketKeeper() feemarketkeeper.Keeper { return tApp.fe
 func (tApp TestApp) GetLiquidKeeper() liquidkeeper.Keeper       { return tApp.liquidKeeper }
 func (tApp TestApp) GetEarnKeeper() earnkeeper.Keeper           { return tApp.earnKeeper }
 func (tApp TestApp) GetRouterKeeper() routerkeeper.Keeper       { return tApp.routerKeeper }
-
-// TODO remove me.
-func (tApp TestApp) GetKavamintKeeper() kavamintkeeper.Keeper   { return kavamintkeeper.Keeper{} }
```
pirtleshell, line 123: this is the only removal outside of x/kavamint module & proto code. put it as a placeholder in #1452 so that project could build without errors coming from the kavamint tests that use this testapp getter.

PR #1459, opened 2023-01-26
remove legacy-community-pool query

pirtleshell:
removes `legacy-community-pool` query. the community pool is no longer being migrated to a module account, so the balance of the community pool can still be queried via the distribution module.


nddeluca (approved):
lgtm, OK to ignore the proto-lint failure.  Those breaking changes look OK

PR #1466, opened 2023-02-02
Nd fix mint module testnet store loading

nddeluca:


pirtleshell (approved):
ran acceptance testing on local chains:

**testnet upgrade handler**: on a chain that was manually upgraded from 0.19 -> 0.20 -> 0.21
* confirmed deletion of iavl tree versions for x/mint before upgrade
* confirmed all state changes in upgrade handler (distribution param change, community pool fund transfer before vs after upgrade, mint params)
* confirmed lend deposit/withdrawl via community proposal

Confirmed all of the above on a chain with `default` pruning and on a chain with `everything` pruning. Confirmed we can stop & start the chain with new binary.

**mainnet upgrade handler**: on a chain that was manually upgraded from 0.19 -> 0.21
* confirmed no unexpected iavl tree changes (all x/mint verisons remain intact)
* confirmed lend deposit/withdraw proposals
* confirmed new binary runs for >100 blocks with `everything` pruning

👏 :shipit:

PR #1475, opened 2023-02-22
add initial E2E testing setup

pirtleshell:
* adds `tests/e2e` which includes a docker setup for running a local chain & running a suite of tests against it
* tests initialize an unauthenticated evm client & some kava query clients
* support creating new signing clients & awaiting transactions broadcast with them
* includes preliminary example tests that highlight querying kava chain, querying the evm, and creating a new funded account

The local chain that is run is based on a chain generated by kvtool. Future work should include automating the generation of the node configuration so we can support things like running multiple chains or running a single chain with multiple validators.


go.mod
```diff
@@ -12,12 +12,14 @@ require (
 	github.com/golang/protobuf v1.5.2
 	github.com/gorilla/mux v1.8.0
 	github.com/grpc-ecosystem/grpc-gateway v1.16.0
+	github.com/kava-labs/go-tools v0.0.0-20221224222255-39c4be283202
```
nddeluca, line 15: We should vendor the signer and drop the circular go-tools dependency

go.mod
```diff
@@ -38,14 +43,15 @@ require (
...
+	github.com/cenkalti/backoff/v4 v4.2.0 // indirect
 	github.com/cespare/xxhash v1.1.0 // indirect
 	github.com/cespare/xxhash/v2 v2.1.2 // indirect
 	github.com/coinbase/rosetta-sdk-go v0.7.0 // indirect
 	github.com/confio/ics23/go v0.7.0 // indirect
+	github.com/containerd/continuity v0.3.0 // indirect
 	github.com/cosmos/btcutil v1.0.4 // indirect
 	github.com/cosmos/go-bip39 v1.0.0 // indirect
-	github.com/cosmos/iavl v0.19.4 // indirect
+	github.com/cosmos/iavl v0.19.3 // indirect
```
nddeluca, line 54: iavl downgrades are bad

tests/e2e/generated/kava-1/config/app.toml
```diff
@@ -0,0 +1,272 @@
...
+# For more information, see https://github.com/toml-lang/toml
+
+###############################################################################
+###                           Base Configuration                            ###
+###############################################################################
+
+# The minimum gas prices a validator is willing to accept for processing a
+# transaction. A transaction's fees must meet the minimum of any denomination
+# specified in this config (e.g. 0.25token1;0.0001token2).
+minimum-gas-prices = "0.001ukava"
```
nddeluca, line 11: Lets also add an akava min fee for the evm side

nddeluca: minimum-gas-prices = "0.001ukava;1000000000akava"

tests/e2e/generated/kava-1/config/app.toml
```diff
@@ -0,0 +1,272 @@
...
+
+# default: the last 100 states are kept in addition to every 500th state; pruning at 10 block intervals
+# nothing: all historic states will be saved, nothing will be deleted (i.e. archiving node)
+# everything: all saved states will be deleted, storing only the current and previous state; pruning at 10 block intervals
+# custom: allow pruning options to be manually specified through 'pruning-keep-recent', 'pruning-keep-every', and 'pruning-interval'
+pruning = "default"
```
nddeluca, line 17: Let's prune nothing here?

go.mod
pirtleshell: good catch. was probably pulled in from go-tools. should be resolved now. now we're only bringing in new packages (all from dockertest package) except for a minor version bump in cenkalti/backoff ([diff](https://github.com/cenkalti/backoff/compare/v4.1.1...v4.1.3))

PR #1477, opened 2023-02-24
Upgrade to Cosmos v0.46

drklee3:
Version upgrades:
* cosmos v0.45.10 -> v0.46.10
* ethermint v0.14.0 -> v0.21.0
* ibc v3.4.0 -> v6.1.0

Note that Cosmos and Ethermint forked dependencies use a pseudo-version targeting the latest commits on the branches, and should be updated to use a tagged version when review of those dependecies are completed.


cmd/kava/cmd/root.go
```diff
@@ -81,6 +88,7 @@ func NewRootCmd() *cobra.Command {
 // addSubCmds registers all the sub commands used by kava.
 func addSubCmds(rootCmd *cobra.Command, encodingConfig params.EncodingConfig, defaultNodeHome string) {
 	rootCmd.AddCommand(
+		rpc.StatusCommand(),
```
drklee3, line 91: Replaced our own status command with the cosmos rpc status command - not sure why we had a custom one in the first place.

cmd/kava/cmd/root.go
```diff
@@ -69,7 +71,12 @@ func NewRootCmd() *cobra.Command {
 
 			customAppTemplate, customAppConfig := servercfg.AppConfig("ukava")
 
-			return server.InterceptConfigsPreRunHandler(cmd, customAppTemplate, customAppConfig)
+			return server.InterceptConfigsPreRunHandler(
+				cmd,
+				customAppTemplate,
+				customAppConfig,
+				tmcfg.DefaultConfig(),
```
drklee3, line 78: Includes tendermint config option here, unsure if we want any modifications to it.

cmd/kava/cmd/root.go
rhuairahrighairidh: This was to fix a mild regression in the sdk where the json keys became capitalized in the v40 upgrade: https://github.com/Kava-Labs/kava/pull/1055

I'd keep our version. Might be able to upstream it to cosmos

cmd/kava/cmd/root.go
```diff
@@ -85,6 +86,10 @@ func NewRootCmd() *cobra.Command {
 	return rootCmd
 }
 
+func addModuleInitFlags(startCmd *cobra.Command) {
+	crisis.AddModuleInitFlags(startCmd)
+}
+
```
rhuairahrighairidh, line 92: any reason to add this func when `appCreator.addStartCmdFlags` already exists?

I don't have a strong preference which to keep, but probably shouldn't have both.

x/incentive/keeper/rewards_earn.go
```diff
@@ -179,7 +182,18 @@ func (k Keeper) collectDerivativeStakingRewards(ctx sdk.Context, collateralType
 		// otherwise there's no validator or delegation yet
 		rewards = nil
 	}
-	return sdk.NewDecCoinsFromCoins(rewards...)
+
+	// Bug with NewDecCoinsFromCoins when calling passing 0 amount Coin, see
+	// https://github.com/cosmos/cosmos-sdk/pull/12903
+	// Fix is in Cosmos-SDK v0.47.0
```
rhuairahrighairidh, line 185: Do we need to compensate for this bug in our other uses of `sdk.NewDecCoinsFromCoins`?

x/incentive/keeper/msg_server_earn_test.go
```diff
@@ -214,10 +214,24 @@ func (suite *HandlerTestSuite) TestEarnLiquidClaim() {
...
-		TruncateInt()
-	suite.BalanceEquals(userAddr2, preClaimBal2.Add(sdk.NewCoin("ukava", stakingRewards2)))
+		RoundInt()
+
+	suite.BalanceInEpsilon(
+		userAddr2,
+		preClaimBal2.Add(sdk.NewCoin("ukava", stakingRewards2)),
+		// Highest precision to allow 1ukava margin of error
+		// 820778117815 vs 820778117814
+		1e-11,
```
rhuairahrighairidh, line 224: Do you know where this change in balance came from?

x/issuance/types/msg_test.go
rhuairahrighairidh, line 1: Swap uses colons in pool id's. Could a coin with a colon cause problems with swap?

app/ante/ante.go
```diff
@@ -163,7 +164,7 @@ func newEthAnteHandler(options HandlerOptions) sdk.AnteHandler {
 		evmante.NewEthMempoolFeeDecorator(options.EvmKeeper),   // Check eth effective gas price against minimal-gas-prices
 		evmante.NewEthValidateBasicDecorator(options.EvmKeeper),
 		evmante.NewEthSigVerificationDecorator(options.EvmKeeper),
-		evmante.NewEthAccountVerificationDecorator(options.AccountKeeper, options.BankKeeper, options.EvmKeeper),
+		evmante.NewEthAccountVerificationDecorator(options.AccountKeeper, options.EvmKeeper),
 		evmante.NewEthGasConsumeDecorator(options.EvmKeeper, options.MaxTxGasWanted),
 		evmante.NewCanTransferDecorator(options.EvmKeeper),
```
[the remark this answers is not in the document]
drklee3: Do you mean reordered like this? After as in higher in the list since it runs the next ones first right 🤔 

```go
		evmante.NewCanTransferDecorator(options.EvmKeeper),
		evmante.NewEthAccountVerificationDecorator(options.AccountKeeper, options.EvmKeeper),
		evmante.NewEthGasConsumeDecorator(options.EvmKeeper, options.MaxTxGasWanted),
```

app/tally_handler_test.go
```diff
@@ -190,6 +191,7 @@ func (suite *tallyHandlerSuite) TestTallyOutcomes() {
 		proposal := suite.createProposal()
 
 		// no stake
+		suite.app.DeleteGenesisValidator(suite.T(), suite.ctx)
```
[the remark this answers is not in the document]
drklee3: Total bonded tokens becomes non-zero, which returns true for `burns`.

Could be removed + expected value changed, but thought it would be more clear if we have a clean no-validator state for the test.

cmd/kava/cmd/root.go
drklee3: ah missed there's already a func for it, removed the new one

x/incentive/keeper/rewards_earn.go
drklee3: I looked through our other uses, mostly in tests and with incentive `RewardsPerSecond` - I don't believe these coins would have a zero amount, but it isn't validated. I can add some extra validation to make sure the coins in `RewardsPerSecond` are non-zero

x/committee/types/permissions.go
```diff
@@ -17,11 +17,11 @@ import (
 func init() {
 	// CommitteeChange/Delete proposals are registered on gov's ModuleCdc (see proposal.go).
 	// But since these proposals contain Permissions, these types also need registering:
-	govtypes.ModuleCdc.RegisterInterface((*Permission)(nil), nil)
-	govtypes.RegisterProposalTypeCodec(GodPermission{}, "kava/GodPermission")
-	govtypes.RegisterProposalTypeCodec(TextPermission{}, "kava/TextPermission")
-	govtypes.RegisterProposalTypeCodec(SoftwareUpgradePermission{}, "kava/SoftwareUpgradePermission")
-	govtypes.RegisterProposalTypeCodec(ParamsChangePermission{}, "kava/ParamsChangePermission")
+	govv1beta1.ModuleCdc.RegisterInterface((*Permission)(nil), nil)
+	govv1beta1.ModuleCdc.Amino.RegisterConcrete(GodPermission{}, "kava/GodPermission", nil)
```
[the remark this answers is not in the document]
drklee3: hmm seems like it only accepts `sdk.Msg` types which the permission types aren't - I can update the other `RegisterConcrete` uses for the `MsgSubmitProposal` and `MsgVote` types to use it.

Might also be able to replace the init chunk with just calling `RegisterLegacyAminoCodec(govv1beta1.ModuleCdc.LegacyAmino)`

app/test_common.go
```diff
@@ -185,9 +358,57 @@ func (tApp TestApp) InitializeFromGenesisStatesWithTimeAndChainIDAndHeight(genTi
...
+	require.True(t, found, "genesis validator not found")
+
+	delegations := sk.GetValidatorDelegations(ctx, genVal.GetOperator())
+	for _, delegation := range delegations {
+		_, err := sk.Undelegate(ctx, delegation.GetDelegatorAddr(), genVal.GetOperator(), delegation.Shares)
+		require.NoError(t, err)
+	}
+}
+
+func (tApp TestApp) ResetBankState(t *testing.T, ctx sdk.Context) {
```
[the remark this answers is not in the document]
drklee3: Yeah it only clears from the initial genesis account bank balances - when tests want to test without the initial validator + initial validator staked tokens. Updated it to be more clear, but maybe removed later if we decide to work around the initial validator requirement

x/incentive/keeper/msg_server_earn_test.go
drklee3: Haven't pinned it 100% down to where exactly, but the total ukava supply changed because of the additional initial validator, and there is bankers rounding when claiming which could cause changes of 1 - along with the `stakingRewards2` in the test being calculated by 99% of the total directly instead of through the incentive shares

x/issuance/types/msg_test.go
drklee3: Yup seems like pools can be created fine which would cause errors afterwards, does adding extra validation in swap params to disallow colons in denoms sound good?

drklee3: added validation

https://github.com/Kava-Labs/kava/pull/1477/files#diff-049c0515998154124b7e34bff956489ac2b22a07823b9c649dd1d48f5d6bc10eR108-R117

x/committee/types/committee.go
```diff
@@ -24,10 +24,7 @@ const (
 func init() {
 	// CommitteeChange/Delete proposals are registered on gov's ModuleCdc (see proposal.go).
 	// But since these proposals contain Committees, these types also need registering:
-	govtypes.ModuleCdc.RegisterInterface((*Committee)(nil), nil)
-	govtypes.RegisterProposalTypeCodec(BaseCommittee{}, "kava/BaseCommittee")
```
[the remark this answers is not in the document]
drklee3: `RegisterProposalTypeCodec` method got removed, but yes seems like those weren't being registered and still need to be - Registered them along with the other types in 89d0fde0e38649980ff6ae962b04e35d1fbfff61

go.mod
```diff
@@ -3,155 +3,193 @@ module github.com/kava-labs/kava
 go 1.18
 
 require (
-	github.com/cosmos/cosmos-proto v1.0.0-alpha6
-	github.com/cosmos/cosmos-sdk v0.45.10
-	github.com/cosmos/ibc-go/v3 v3.4.0
-	github.com/ethereum/go-ethereum v1.10.16
+	cosmossdk.io/math v1.0.0-beta.6.0.20230216172121-959ce49135e4
```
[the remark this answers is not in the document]
drklee3: Is it possible to use `github.com/cosmos/cosmos-sdk` instead of `cosmossdk.io/math` set in their `go.mod`?

I get this error when trying to use `github.com/cosmos/cosmos-sdk/math`

```text
go: downloading github.com/cosmos/cosmos-sdk/math v1.0.0-rc.0
go: github.com/cosmos/cosmos-sdk/math@v1.0.0-rc.0: parsing go.mod:
	module declares its path as: cosmossdk.io/math
	        but was required as: github.com/cosmos/cosmos-sdk/math
```

cmd/kava/cmd/root.go
pirtleshell: can confirm that this binary now marshals the fields like `NodeInfo`, `SyncInfo`, and `ValidatorInfo`, instead of in snake_case.
took a look through some of our automation and there are a few places we pipe `kava status` to `jq` and access fields. this change breaks those.
though usually we curl the `/status` endpoint directly which this does not affect.

interestingly, it looks like the sdk's `status` command doesn't write to stdout, which breaks piping to `jq` altogether. in any case this is a breaking change we may not want to have to work through the consequences of in this release. worth adding a ticket to unvendor the status command in a future release, i think.

go.mod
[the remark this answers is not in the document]
pirtleshell: ah, yup. ok that makes sense. easiest to just keep it as is then because they define that as their module path in go.mod.

safe to disregard, but worth the mental note.

app/ante/ante.go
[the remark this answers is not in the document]
drklee3: ahh I see now to be before `NewEthGasConsumeDecorator`, updated now ty for the poke

cmd/kava/cmd/root.go
drklee3: Updated to a custom one with snake_case - seems like it is outputting to stdout

x/evmutil/client/cli/evm.go
```diff
@@ -27,7 +27,12 @@ func CanSignEthTx(ctx client.Context) error {
...
 	}
 
-	if keyInfo.GetAlgo() != hd.EthSecp256k1Type {
+	pubKey, err := keyInfo.GetPubKey()
+	if err != nil {
+		return err
+	}
+
+	if pubKey.Type() != string(hd.EthSecp256k1Type) {
 		return fmt.Errorf("from address does not support %v", hd.EthSecp256k1Type)
```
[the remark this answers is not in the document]
drklee3: ah I read it in the different way that seemed correct, "from address" referring to the input address instead of the requirement address type. But I can see how it is understood the other way, updated to be more clear without the ambiguity 😃

cmd/kava/cmd/root.go
rhuairahrighairidh: I don't think we make any standard changes to the default file.

x/incentive/types/params.go
```diff
@@ -207,6 +207,11 @@ func (rp RewardPeriod) Validate() error {
 	if !rp.RewardsPerSecond.IsValid() {
 		return fmt.Errorf("invalid reward amount: %s", rp.RewardsPerSecond)
 	}
+
+	if rp.RewardsPerSecond.Amount.IsZero() {
+		return fmt.Errorf("reward amount cannot be zero: %v", rp.RewardsPerSecond)
+	}
+
```
rhuairahrighairidh, line 214: nit: what was the reason for this change?

rhuairahrighairidh (approved):
had a final look through, found a couple of small things I added to #1522 , but otherwise looks good for internal testnet 🚀

x/incentive/types/params.go
drklee3: It is passed to `sdk.NewDecCoinsFromCoins` in the accumulator which would have the sdk bug in the method if it has a 0 amount

PR #1503, opened 2023-03-16
Add incentive grpc query service

drklee3:


x/incentive/keeper/grpc_query.go
```diff
@@ -0,0 +1,331 @@
...
+		addr, err := sdk.AccAddressFromBech32(req.Owner)
+		if err != nil {
+			return nil, status.Errorf(codes.InvalidArgument, "invalid address: %s", err)
+		}
+
+		owner = addr
+	}
+
+	if err := s.queryRewards(sdkCtx, &res, owner, hasOwner, req.RewardType); err != nil {
+		return nil, err
```
galxy25, line 67: for debugging purposes might be helpful to return the address and reward type that we used to query rewards from

x/incentive/keeper/grpc_query.go
```diff
@@ -0,0 +1,331 @@
...
+	}
+
+	if !req.Unsynchronized {
+		if err := s.synchronizeRewards(sdkCtx, &res); err != nil {
```
galxy25, line 71: curious if this query request result in state changes?

x/incentive/keeper/grpc_query.go
```diff
@@ -0,0 +1,331 @@
...
+
+	// Incentive only APYs
+	for _, param := range params.EarnRewardPeriods {
+		// Skip bkava as it's calculated earlier with staking rewards
+		if param.CollateralType == liquidtypes.DefaultDerivativeDenom {
+			continue
+		}
+
+		// Value in the vault in the same denom as CollateralType
+		vaultTotalValue, err := s.keeper.earnKeeper.GetVaultTotalValue(sdkCtx, param.CollateralType)
```
galxy25, line 171: do we need to do any bounds checking (e.g. greater than 0) on the return value before using it below (or does this method return an error in that case?)

x/incentive/keeper/grpc_query.go
```diff
@@ -0,0 +1,331 @@
...
+	rewardType string,
+) error {
+	rewardType = strings.ToLower(rewardType)
+	isAllRewards := rewardType == ""
+
+	if !rewardTypeIsValid(rewardType) {
+		return status.Errorf(codes.InvalidArgument, "invalid reward type: %s", rewardType)
+	}
+
+	if isAllRewards || rewardType == "usdx_minting" {
```
galxy25, line 204: nit: suggest extracting these strings out to constant values

x/incentive/keeper/grpc_query.go
galxy25: on further reading looks like this method only results in simulated state changes 😅

x/incentive/module.go
```diff
@@ -136,7 +136,7 @@ func (AppModule) QuerierRoute() string {
 // RegisterServices registers module services.
 func (am AppModule) RegisterServices(cfg module.Configurator) {
 	types.RegisterMsgServer(cfg.MsgServer(), keeper.NewMsgServerImpl(am.keeper))
-	// TODO: types.RegisterQueryServer(cfg.QueryServer(), keeper.NewQueryServerImpl(am.keeper, am.accountKeeper, am.bankKeeper))
```
galxy25, line 139: `#TODONE` 👏🏽

galxy25 (approved):
minor nits and comments

x/incentive/keeper/grpc_query.go
```diff
@@ -0,0 +1,331 @@
...
+	}
+
+	return &res, nil
+}
+
+func (s queryServer) RewardFactors(
+	ctx context.Context,
+	req *types.QueryRewardFactorsRequest,
+) (*types.QueryRewardFactorsResponse, error) {
+	if req == nil {
```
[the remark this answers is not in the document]
drklee3: I added it to be consistent with cosmos since they also do these checks, haven't checked if actually necessary

x/incentive/keeper/grpc_query.go
drklee3: yup the `GetAPYFromMultiRewardPeriod` will return 0 for apy if the total value is 0

x/incentive/keeper/grpc_query.go
drklee3: yup it returns both now in `queryRewards`

PR #1528, opened 2023-04-04
fix e2e test in CI pipeline

pirtleshell:
* updates docker container base image to go 1.19
* updates kvtool to have cosmos v46 support in master template
* add a .tool-versions file for automated go version management


.tool-versions
```diff
@@ -0,0 +1 @@
+golang 1.19
```
galxy25, line 1: any reason not to update to latest (1.20)?

galxy25 (approved):

.tool-versions
pirtleshell: just matching the version in go.mod which itself is pinned to the [version used by cosmos-sdk](https://github.com/cosmos/cosmos-sdk/blob/v0.46.11/go.mod#L1). historically, the go version that kava is built with was important so i figured i'd keep parity with the repo itself

PR #1530, opened 2023-04-04
Update deprecated Cosmos methods

drklee3:
Main changes:
- `"cosmossdk.io/math"` for `Int` types
- `cosmossdk.io/errors"` for error methods


galxy25 (approved):
spot checked the first 40 files, going to let the ci take care of the rest 😁

PR #1533, opened 2023-04-05
Use kava-labs/cometbft-db fork for rocksdb support

drklee3:


galxy25:
@drklee3 can you rebase off of ls-support-rocksdb-cosmos-v46 so the ci checks will run against this pr and verify if this fixes the issue

.github/scripts/install-rocksdb.sh
```diff
@@ -9,7 +9,7 @@ git clone https://github.com/facebook/rocksdb.git /home/runner/rocksdb
 
 cd /home/runner/rocksdb
 
-git checkout v7.7.3
+git checkout v7.9.2
```
galxy25, line 12: this should be pulled from the environment and set to v7.10.2 (another benefit to rebasing off of https://github.com/Kava-Labs/kava/pull/1532/files)

galxy25:
merged into `ls-support-rocksdb-cosmos-v46` manually

PR #1539, opened 2023-04-06
feat: block new vesting create messages

drklee3:


galxy25 (approved):

galxy25:
maybe worth a changelog entry?

pirtleshell (approved):

drklee3:
> maybe worth a changelog entry?

Since the messages weren't available before, and won't be allowed after upgrade there isn't really any effective changes with these messages, maybe better mentioned as a small note to any cosmos v0.46 upgrades docs when added?

PR #1542, opened 2023-04-06
feat: re-add validator-vesting legacy rest endpoints

drklee3:
## Description
Adds the legacy `/vesting/*` rest endpoints back, along with the required cosmos methods.

## Checklist

### DIFF
--- a/client/rest/rest.go
+++ b/client/rest/rest.go
@@ -1,0 +1,449 @@
+// Package rest provides HTTP types and primitives for REST
+// requests validation and responses handling.
+package rest
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strconv"
+	"strings"
+
+	"github.com/tendermint/tendermint/types"
+
+	"github.com/cosmos/cosmos-sdk/client"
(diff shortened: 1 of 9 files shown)


client/rest/rest.go
```diff
@@ -0,0 +1,449 @@
...
+		WriteErrorResponse(w, http.StatusUnauthorized, fmt.Sprintf("invalid from address: %s", br.From))
+		return false
+	}
+
+	return true
+}
+
+// ReadRESTReq reads and unmarshals a Request's body to the the BaseReq struct.
+// Writes an error response to ResponseWriter and returns false if errors occurred.
+func ReadRESTReq(w http.ResponseWriter, r *http.Request, cdc *codec.LegacyAmino, req interface{}) bool {
```
galxy25, line 137: `to the BaseReq struct`
Idiomatic go is to return error vs bool to signal error and have callers check if nil -> https://go.dev/doc/effective_go#multiple-returns

galxy25 (approved):

client/rest/rest.go
drklee3: The rest files are just the removed files from cosmos temporarily copied while we support the endpoints, I'll just leave it as is for now since we may be removing it later again 😅

PR #1544, opened 2023-04-08
Update go.mod replace directives

drklee3:
## Description

These updates follow the Cosmos SDK replacements.

* Remove replace for `github.com/confio/ics23/go`, v0.9.0 includes dragonberry security patch
* Update `github.com/cosmos/keyring` to v1.2.0.
* Add `github.com/dgrijalva/jwt-go` replacement for security issue.

There is one replacement in Cosmos SDK that is *not* included, unclear the reasoning why this was added:
* `github.com/jhump/protoreflect => github.com/jhump/protoreflect v1.9.0`

Additional version change to address minor security alert

* github.com/hashicorp/go-getter to v1.7.1, see #1527

## Checklist


galxy25 (approved):

rhuairahrighairidh (approved):
Good catch with the go-getter upgrade.
I had a look were the jhump/protoreflect replace came from. It was pretty unclear why it was added. But it's only used in a cosmos-sdk test so seems fine to ignore.

PR #1545, opened 2023-04-11
v22 upgrade handler

DracoLi:
- add v22 upgrade handler with logic to add authz grants and move community funds
- Depends on changes from https://github.com/Kava-Labs/ethermint/pull/18

## Checklist
 - [x] Use new ethermint tag after its merged. https://github.com/Kava-Labs/ethermint/pull/18

### DIFF
--- a/app/upgrades.go
+++ b/app/upgrades.go
@@ -86,7 +86,7 @@ func GrantGovCommunityPoolMessages(
 ) {
 	communityAddr := accountKeeper.GetModuleAddress(communitytypes.ModuleName)
 	govAddr := accountKeeper.GetModuleAddress(govtypes.ModuleName)
-	allowedMsgs := getCommunityPoolAllowedMsgs()
+	allowedMsgs := GetCommunityPoolAllowedMsgs()
 	for _, msg := range allowedMsgs {
 		auth := authz.NewGenericAuthorization(msg)
 		if err := authzKeeper.SaveGrant(ctx, govAddr, communityAddr, auth, nil); err != nil {
@@ -129,7 +129,7 @@ func FundCommunityPoolModule(
 	distKeeper.SetFeePool(ctx, feePool)
 }
 
-func getCommunityPoolAllowedMsgs() []string {
+func GetCommunityPoolAllowedMsgs() []string {
 	return []string{
 		"/cosmos.bank.v1beta1.MsgSend",
 		"/cosmos.bank.v1beta1.MsgMultiSend",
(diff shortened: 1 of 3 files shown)


app/upgrades.go
```diff
@@ -1,3 +1,158 @@
...
+	govAddr := accountKeeper.GetModuleAddress(govtypes.ModuleName)
+	allowedMsgs := []string{
+		"/cosmos.bank.v1beta1.MsgSend",
+		"/cosmos.bank.v1beta1.MsgMultiSend",
+		"/cosmos.staking.v1beta1.MsgDelegate",
+		"/cosmos.staking.v1beta1.MsgBeginRedelegate",
+		"/cosmos.staking.v1beta1.MsgUndelegate",
+		"/cosmos.staking.v1beta1.MsgCancelUnbondingDelegation",
+		"/cosmos.gov.v1.MsgVote",
+		"/cosmos.gov.v1beta1.MsgVote",
```
rhuairahrighairidh, line 97: do we have an expected use case for voting? The community pool vote would have to created by another gov proposal, which feels kind of confusing. If it's not a requirement I'd probably remove the vote msgs to keep things simpler.

app/upgrades.go
```diff
@@ -1,3 +1,158 @@
...
+
+// GrantGovCommunityPoolMessages grants x/gov module account access to submit x/authz messages from the community pool module account.
+func GrantGovCommunityPoolMessages(
+	ctx sdk.Context,
+	authzKeeper authzkeeper.Keeper,
+	accountKeeper authkeeper.AccountKeeper,
+) {
+	communityAddr := accountKeeper.GetModuleAddress(communitytypes.ModuleName)
+	govAddr := accountKeeper.GetModuleAddress(govtypes.ModuleName)
+	allowedMsgs := []string{
```
pirtleshell, line 89: i'd consider pulling this list out into a variable that could be imported by an e2e test that does acceptance to check that each message has actually been granted.

pirtleshell (approved):
migrations look good. just had some thoughts on messages that should potentially also be granted.

approved pending resolution of @rhuairahrighairidh 's comment on MsgVote grants.

remember to update the base of this PR to the to-be-cut release branch

app/upgrades.go
```diff
@@ -1,3 +1,160 @@
...
+	}
+
+	// make sure x/distribution knows that there're no funds in the community pool.
+	// we keep the leftover decimal change in the account to ensure all funds are accounted for.
+	feePool := distKeeper.GetFeePool(ctx)
+	feePool.CommunityPool = leftoverDust
+	distKeeper.SetFeePool(ctx, feePool)
+}
+
+func getCommunityPoolAllowedMsgs() []string {
```
pirtleshell, line 132: this could just be a variable, though a func works. but i would make it public so it can be imported into e2e tests

```go
var V22UpgradeCommunityModuleGrants = []string{}
```

drklee3 (approved):

PR #1548, opened 2023-04-12
fix: always use latest seed script when seeding chains

galxy25:
## Description

fix: always use latest seed script when seeding chains

## Checklist

### DIFF
--- a/.github/workflows/cd-seed-chain.yml
+++ b/.github/workflows/cd-seed-chain.yml
@@ -76,8 +76,13 @@ jobs:
       - name: compile default erc20 contracts
         run: make compile-contracts
         working-directory: kava-bridge
+      - name: download seed script from master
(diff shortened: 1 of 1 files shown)


.github/workflows/cd-seed-chain.yml
```diff
@@ -76,8 +76,11 @@ jobs:
       - name: compile default erc20 contracts
         run: make compile-contracts
         working-directory: kava-bridge
+      - name: download seed script from master
```
drklee3, line 79: maybe we can checkout back to master after the kava install instead of downloading, in case we need other files later on?

drklee3 (approved):

.github/workflows/cd-seed-chain.yml
galxy25: I like it, thats a better idea / more future proof

galxy25: ended up going with the original way since checking out the kava repo in the github action messed up the install paths for all the hardhat tools used by the seed script

PR #1550, opened 2023-04-12
bug: fix invalid accumulation time genesis validation

pirtleshell:
## Description
Mainnet genesis exports fail validation checks due to incentive accumulation times being more than a year ago. This PR removes the faulty validation, replacing it instead with a check that ensures the value is set. This way, no matter how much time passes, an exported genesis will still have a valid accumulation time value while still preventing misconfiguration from rewarding from the beginning of the unix epoch.

**before**:
```
panic: found accumulation time '2021-06-10 16:43:11.679705 +0000 UTC' more than '8760h0m0s' behind genesis time '2023-04-12 16:55:11.803644562 +0000 UTC'
```

**after**:
```
File at export-4355333.json is a valid genesis file
```

## Checklist

### DIFF
--- a/x/incentive/genesis.go
+++ b/x/incentive/genesis.go
@@ -14,7 +14,7 @@ const year = 365 * 24 * time.Hour
 
 // EarliestValidAccumulationTime is how far behind the genesis time an accumulation time can be for it to be valid.
 // It's a safety check to ensure rewards aren't accidentally accumulated for many years on the first block (eg since Jan 1970).
-var EarliestValidAccumulationTime time.Duration = year
+var EarliestValidAccumulationTime time.Duration = 5 * year
 
 // InitGenesis initializes the store state from a genesis state.
 func InitGenesis(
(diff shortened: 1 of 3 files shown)


galxy25 (approved):

x/incentive/genesis.go
```diff
@@ -14,7 +14,7 @@ const year = 365 * 24 * time.Hour
 
 // EarliestValidAccumulationTime is how far behind the genesis time an accumulation time can be for it to be valid.
 // It's a safety check to ensure rewards aren't accidentally accumulated for many years on the first block (eg since Jan 1970).
-var EarliestValidAccumulationTime time.Duration = year
+var EarliestValidAccumulationTime time.Duration = 5 * year
```
rhuairahrighairidh, line 17: nit: could be removed as not used anymore

rhuairahrighairidh (approved):

PR #1559, opened 2023-04-14
add liquidate msg spam filter that only runs on checkTx with 1e6 gas

nddeluca:


app/ante/ante.go
```diff
@@ -117,6 +117,7 @@ func newCosmosAnteHandler(options cosmosHandlerOptions) sdk.AnteHandler {
 	decorators := []sdk.AnteDecorator{}
 
 	decorators = append(decorators,
+		NewHardLiquidateSpamFilter(1000000),
```
galxy25, line 120: can we read this value from config file /environment variable?

app/ante/liquidate_spam_filter.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+	}
+
+	// continue if no hard liquidate messages are found
+	if !hasHardLiquidate {
+		return next(ctx, tx, simulate)
+	}
+
+	// reject transaction if greater than the set gas limit
+	if feeTx.GetGas() > f.gasLimit {
```
galxy25, line 55: I think we don't want to break on line 45 on the first hard message, but collect all of them in the block and check that none of them are above the gas limit

app/ante/ante.go
nddeluca: moved to a constant

app/ante/liquidate_spam_filter.go
```diff
@@ -0,0 +1,61 @@
...
+		return ctx, sdkerrors.Wrapf(sdkerrors.ErrUnauthorized, "%v", fmt.Errorf("liquidation gas %d over %d limit", feeTx.GetGas(), f.gasLimit))
+	}
```
pirtleshell, line 57: i'd add a check that `next`s txs that have `fee < gasLimit` before we iterate through all the messages of the tx. 

that way we dont need to read into the messages for all low fee txs that will pass regardless of whether they have har liquidate messages or not.

pirtleshell: current logic:
* filter for hard liquidate message tx
* reject if gas > gas limit

proposed logic:
* continue if gas <= gas limit
* filter for hard liquidate messages & reject if found

app/ante/liquidate_spam_filter.go
nddeluca: We can only check one tx a time here, and we are limiting any number of messages to 1,000,000 gas

app/ante/ante.go
galxy25: may not be needed for this specific change, but my thought was that if we wanted to change this limit without having to upgrade the kava binary, reading this value from config or the environment (with a fallback to this default value) would allow us to do that versus hard coding it

app/ante/liquidate_spam_filter.go
nddeluca: hmm seems like a good optimization will add

pirtleshell (approved):

app/ante/ante.go
nddeluca: ya is a good call but going to leave as an improvement since we won't apply this generally to mainnet releases atm.  I need to do more research on gas costs before we tag for all validators

rhuairahrighairidh (approved):
looks good, had a think through
- possibility of state breaking changes
- rechecktx / simulate stuff
- other signing modes

PR #1562, opened 2023-04-17
feat: add CommunityPoolLendWithdrawalPermission

pirtleshell:
## Description

* Adds committee permission for withdrawing community pool lend positions

## Checklist


CHANGELOG.md
```diff
@@ -44,6 +44,7 @@ Ref: https://keepachangelog.com/en/1.0.0/
 - (deps) [#1477] Migrate to CometBFT.
 - (x/incentive) [#1512] Add grpc query service.
 - (deps) [#1544] Bump confio/ics23/go to v0.9.0, cosmos/keyring to v1.2.0.
+- (x/committee) [#1562] Add CommunityPoolLendWithdrawalPermission
```
drklee3, line 47: don't forget to add the link reference at the bottom of the file 😃 should be easy to copy paste

pirtleshell: haha totally missed that these were linked below, lol. thanks!

galxy25 (approved):

x/committee/types/codec.go
```diff
@@ -47,7 +48,6 @@ func init() {
 	RegisterProposalTypeCodec(govv1beta1.TextProposal{}, "cosmos-sdk/TextProposal")
 	RegisterProposalTypeCodec(upgradetypes.SoftwareUpgradeProposal{}, "cosmos-sdk/SoftwareUpgradeProposal")
 	RegisterProposalTypeCodec(upgradetypes.CancelSoftwareUpgradeProposal{}, "cosmos-sdk/CancelSoftwareUpgradeProposal")
-
```
drklee3, line 50: Does `CommunityPoolLendWithdrawProposal` need to be registered here on the committee ModuleCdc?

karzak (approved):
LGTM, tested against a local kvtool branch and working as expected.

x/committee/types/codec.go
rhuairahrighairidh: yeah it should be because `MsgSubmitProposal.GetSignBytes` uses the codec and contains a Proposal interface. I put together an explainer here: https://kava-labs.atlassian.net/wiki/spaces/ENG/pages/1237483521/Legacy+Amino+Codec+Registration

pirtleshell: ah, i was under the assumption that because they were registered in their respective modules in our code we didnt need to register them here. sounds like that is not the case.

does that mean that other committee proposals originating from our modules also need to be added? which as of right now is just `kavadist.CommunityPoolMultiSpendProposal`

rhuairahrighairidh: yeah probably worth adding those. I think it's not strictly a problem, as everything here is behind governance, but good to be consistent

PR #1565, opened 2023-04-18
feat: add community cdp repay debt proposal

pirtleshell:
## Description

* adds `x/community` proposal `CommunityCDPRepayDebtProposal` which repays debts owned by the community module with community module funds

Follow up PR (#1566) adds corresponding committee permission for submission via committee.

**Note**: Does not include custom CLI command.

## Checklist


proto/kava/community/v1beta1/proposal.proto
```diff
@@ -31,3 +31,15 @@ message CommunityPoolLendWithdrawProposal {
     (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"
   ];
 }
+
+// CommunityCDPRepayDebtProposal repays a cdp debt position owned by the community module
+// This proposal exists primarily to allow committees to repay community module cdp debts.
+message CommunityCDPRepayDebtProposal {
```
drklee3, line 37: nit: should we make the naming consistent with the other ones?

```diff
- CommunityCDPRepayDebtProposal
+ CommunityPoolCDPRepayDebtProposal
```

pirtleshell: i originally had it named this way but purposefully renamed it because there isn't actually any interaction with x/distribution's community pool. the naming this way is to signal that its community module all the way down. i could be convinced to do otherwise, but i found myself getting confused when i realized that this proposal only interacts with x/community module account balance, not the true community pool.

i intended to distinguish it from the `CommunityPoolLend*Proposal`s because those actually use x/distribution's community pool funds and only use x/community as an intermediary. they have both been made obsolete by the authz grants we will perform in the k13 upgrade.

(i recognize that these are essentially one and the same pool of funds, all controlled by governance, so maybe not worth leaking the abstraction in this name?) what do you think?

drklee3: hmm yeah I think it's just a bit confusing with both the x/dist community pool and x/community balance. I thought of the community pool as a big one that includes both the x/dist pool and x/community balance (like the x/dist balance endpoint), but your reasoning makes sense and we can just stick with that since this gets into bike shedding now to figure out the best possible name 😄

app/app.go
```diff
@@ -635,8 +635,9 @@ func NewApp(
 	app.communityKeeper = communitykeeper.NewKeeper(
 		app.accountKeeper,
 		app.bankKeeper,
+		&cdpKeeper,
 		app.distrKeeper,
-		hardKeeper,
+		&hardKeeper,
```
pirtleshell, line 640: creating a public record of the fact that because this was originally passed as a copy, any positions it created never had claims instantiated (because the hooks do not get registered to the copy).
this may have consequences for the open lend position held by the community module.

pirtleshell: confirmed with @rhuairahrighairidh any fixes needed for this would be in a follow up pr. commencing merge!

PR #1573, opened 2023-04-21
feat: Proposal 128 - Change minimum commission to 5%

smartdev0328:
## Checklist


pirtleshell:
Thanks for your contribution. Sorry for the delay in getting dev time for it.

Rather than using antehandlers, we decided to enforce this through the MinValidatorCommission staking parameter. The next upgrade will update the staking param and migrate all existing validators that have a commission lower than 5%. See #1761

PR #1574, opened 2023-04-21
Add gov upgrade e2e tests

DracoLi:
## Description

- Adds e2e tests for gov authz proposals for x/community.
- Requires kvtool updates here https://github.com/Kava-Labs/kvtool/pull/156. Make sure to update your kvtool before running the tests.

### DIFF
--- a/tests/e2e/.env
+++ b/tests/e2e/.env
@@ -22,7 +22,7 @@ E2E_KAVA_UPGRADE_NAME=v0.22.0
 # If IBC tests are enabled this should be >50. Otherwise, this should be >10.
 E2E_KAVA_UPGRADE_HEIGHT=53
 # E2E_KAVA_UPGRADE_BASE_IMAGE_TAG is the tag of the docker image the chain should upgrade from (ie v0.21.0)
-E2E_KAVA_UPGRADE_BASE_IMAGE_TAG=v0.21.0
+E2E_KAVA_UPGRADE_BASE_IMAGE_TAG=v0.21.1
 
 # E2E_KAVA_ERC20_ADDRESS is the address of a pre-deployed ERC20 token.
 # The E2E_KAVA_FUNDED_ACCOUNT_MNEMONIC account should have a balance.


tests/util/addresses.go
```diff
@@ -12,3 +13,8 @@ func SdkToEvmAddress(addr sdk.AccAddress) common.Address {
 func EvmToSdkAddress(addr common.Address) sdk.AccAddress {
 	return sdk.AccAddress(addr.Bytes())
 }
+
+func RandomAccAddress() sdk.AccAddress {
+	_, _, addr := testdata.KeyTestPubAddr()
+	return addr
+}
```
pirtleshell, line 20: i think we should prefer using `app.RandomAddress()` rather than add another method for this.

tests/util/gov.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"context"
+	"errors"
+	"time"
+
+	govtypesv1 "github.com/cosmos/cosmos-sdk/x/gov/types/v1"
+)
+
+func WaitForProposalStatus(
```
pirtleshell, line 11: 🔥

pirtleshell (commented):
i think these tests could be added to master and not only run on the release branch. we should update the master template of kvtool to start with the necessary grants to the community module

the upgrade tests should start with a template like mainnet and then verify that the correct grants et al are added to the community module.

in other words, the functionality you have in this branch's tests should always work. only upgrade handler specific checks should need the upgrade info added to the env. when we cut the release branch we should pr changes to the e2e env so that all prs to that branch run against a post-upgrade chain

pirtleshell:
@DracoLi could you add `release/**` to the matching branches that the default CI runs against? if you add that, all PRs to the release branch will run e2e tests before merge (including this one)

https://github.com/Kava-Labs/kava/blob/95d9fb9aca30c4cb30fa553f87df1f99a8f34be1/.github/workflows/ci-pr.yml#L3-L7

pirtleshell (approved):

tests/util/gov.go
```diff
@@ -0,0 +1,42 @@
...
+	govClient govtypesv1.QueryClient,
+	proposalId uint64,
+	status govtypesv1.ProposalStatus,
+	timeout time.Duration,
```
rhuairahrighairidh, line 15: nit: looks like `timeout` arg isn't used

rhuairahrighairidh (approved):

PR #1581, opened 2023-04-27
Update KAVA.VERSION

downing034:
Point internal testnet to latest master


galxy25 (approved):

downing034:
@galxy25 I saw that protonet has KAVA.VERSION set to `master`  I assume this change is okay to make here as we want to be on the most recent commits _I think_

karzak:
I don't think we want this yet @galxy25  - internal testnet should remain on "latest" (ie v0.21.1) until just prior to a mainnet upgrade. That way, internal testnet will be the last development environment that runs the current mainnet version and can be used for current mainnet acceptance testing.

galxy25:
@downing034 @karzak if you want master to be deployed to internal testnet (to bring it in parity with protonet), this change is needed (after fixing the genesis to be valid https://github.com/Kava-Labs/kava/actions/runs/4822220653/jobs/8589162569?pr=1581) , if you don't then yeah don't merge.

It all depends on what you purpose you want to use internal testnet for.

Originally internal testnet was meant to allow frontend to integrate early in a non local environment with the latest protocol features as soon as they are ready for integration.

Now we have protonet to serve that use case, and recently internal testnet has been used to test out frontend features such as the proposals before deploying to mainnet. 

So again, if you want internal testnet to be similar to protonet, this is the pr for you, if you want it to be more similar to mainnet, this is not the pr for you.

karzak:
Thanks @galxy25 - closing this PR as the workflow going forward will be

1. Use protonet for testing FE again bleeding edge features
2. Keep internal testnet harmonized to deployed mainnet version until just before/just after mainnet upgrade.

PR #1582, opened 2023-04-28
Remove x/evm AllowedParamsChange from stability committee

drklee3:
## Description
Updates the existing migration to remove the x/evm AllowedParamsChange from the permission as the x/evm module no longer uses the params module.

Also adds a test to ensure it works as expected.

## Checklist
- ~~[ ] Changelog has been updated as necessary.~~

### DIFF
--- a/app/upgrades.go
+++ b/app/upgrades.go
@@ -51,13 +51,13 @@ func (app App) RegisterUpgradeHandlers() {
 
 			app.Logger().Info(fmt.Sprintf(
 				"adding lend & cdp committee permissions to stability committee (id=%d)",
-				TestnetStabilityCommitteeId,
+				MainnetStabilityCommitteeId,
 			))
-			AddNewPermissionsToStabilityCommittee(ctx, app.committeeKeeper, TestnetStabilityCommitteeId)
+			AddNewPermissionsToStabilityCommittee(ctx, app.committeeKeeper, MainnetStabilityCommitteeId)
 
 			app.Logger().Info(fmt.Sprintf(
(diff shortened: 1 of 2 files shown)


galxy25 (approved):

app/upgrades.go
```diff
@@ -150,23 +157,56 @@ func FundCommunityPoolModule(
...
 	// get committee
 	comm, found := committeeKeeper.GetCommittee(ctx, committeeId)
 	if !found {
 		panic(fmt.Sprintf("expected to find committee with id %d but found none", committeeId))
 	}
-	// set new permissions
 	perms := comm.GetPermissions()
+
+	// Remove the x/evm param permissions
+	for i, permission := range perms {
```
nddeluca, line 179: We should only make this update for mainnet or the `v0.22.0` upgrade.  Since we already migrated testnet, this would create a binary that can't be synced from the upgrade block.

nddeluca (changes requested):
Need to adjust to only modify the permission on mainnet

app/upgrades.go
nddeluca: We'll have to issue a gov proposal on the public testnet

nddeluca (approved):

app/upgrades.go
```diff
@@ -57,7 +57,7 @@ func (app App) RegisterUpgradeHandlers() {
 
 			app.Logger().Info(fmt.Sprintf(
 				"removing x/evm AllowedParamsChange from stability committee (id=%d)",
-				MainntStabilityCommitteeId,
```
nddeluca: Thanks for the fix 👍 I should have run make test before pushing

PR #1586, opened 2023-05-02
Add mainnet upgrade handler, clear testnet upgrade handler

drklee3:
## Description
- Add mainnet upgrade handler for v0.21.x -> v0.23.0
- Update testnet upgrade handler for v0.22.x -> v0.23.0

## Checklist
 - ~~[ ] Changelog has been updated as necessary.~~

### DIFF
--- a/app/upgrades.go
+++ b/app/upgrades.go
@@ -25,16 +25,18 @@ import (
 )
 
 const (
-	TestnetUpgradeName = "v0.22.0-alpha.0"
+	MainnetUpgradeName = "v0.23.0"
+	TestnetUpgradeName = "v0.23.0-alpha.0"
 
+	MainnetStabilityCommitteeId = uint64(1)
 	TestnetStabilityCommitteeId = uint64(1)
 )
 
 func (app App) RegisterUpgradeHandlers() {
-	// register upgrade handler for testnet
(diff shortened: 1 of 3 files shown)


app/upgrades.go
```diff
@@ -60,13 +68,28 @@ func (app App) RegisterUpgradeHandlers() {
 		},
 	)
 
+	// register upgrade handler for testnet. This only runs the module
+	// migrations, as testnet already ran the other upgrades in v0.22.0.
```
nddeluca, line 72: 👍

app/upgrades.go
```diff
@@ -25,16 +25,18 @@ import (
 )
 
 const (
-	TestnetUpgradeName = "v0.22.0-alpha.0"
+	MainnetUpgradeName = "v0.23.0"
+	TestnetUpgradeName = "v0.23.0-alpha.0"
 
+	MainnetStabilityCommitteeId = uint64(1)
 	TestnetStabilityCommitteeId = uint64(1)
```
nddeluca, line 32: No need to change, but I think we should use the commitee keeper to look up the committee with the correct description (there doesn't seem to be a name field), and not hardcode to values in the future.

nddeluca (approved):

PR #1599, opened 2023-05-25
fix(evmutil): register MsgConvertCosmosCoinToERC20 on amino

rhuairahrighairidh:
## Description
Noticed the new `MsgConvertCosmosCoinToERC20` wasn't registered on amino. So added a test to help catch this in the future by checking all msgs registered on the proto codec are on amino. It's not perfect, we should ideally check `msg.GetSignBytes` returns properly encoded msgs (it uses a local amino, not the app's amino). But it did catch a pre-existing issue where we hadn't setup the registration in `module.go`.

Also added this to the list of gotchas with link to explainer. It's unlikely to cause issues but can in very particular cases mess with tx signing.

## Checklist

### DIFF
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -2,11 +2,17 @@ package app
 
 import (
 	"encoding/json"
+	"fmt"
 	"os"
 	"sort"
 	"testing"
 	"time"
 
+	"github.com/cosmos/cosmos-sdk/codec"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/x/auth/migrations/legacytx"
+	vestingtypes "github.com/cosmos/cosmos-sdk/x/auth/vesting/types"
+	evmtypes "github.com/evmos/ethermint/x/evm/types"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	abci "github.com/tendermint/tendermint/abci/types"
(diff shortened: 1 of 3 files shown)


galxy25 (approved):

app/app_test.go
```diff
@@ -63,6 +69,64 @@ func TestExport(t *testing.T) {
...
+
+		// Only check legacy msgs for amino registration.
+		// Only legacy msg can be signed with amino.
+		_, ok := msg.(legacytx.LegacyMsg)
+		if !ok {
+			continue
+		}
+
+		// Check the msg is registered in amino by checking a repeat registration call panics.
+		panicValue, ok := catchPanic(func() {
```
pirtleshell, line 105: super clever way to check the messages are registered 👍 🥇 

nit: you can probably use [`require.PanicsWithError`](https://pkg.go.dev/github.com/stretchr/testify/require?utm_source=godoc#PanicsWithValue) (or `PanicsWithError`) to test this.

app/app_test.go
```diff
@@ -63,6 +69,64 @@ func TestExport(t *testing.T) {
...
+
+	lcdc := tApp.LegacyAmino()
+
+	// Use the proto codec as the canonical list of msg types.
+	protoCodec := tApp.AppCodec().(*codec.ProtoCodec)
+	msgProtoNames := protoCodec.InterfaceRegistry().ListImplementations(sdk.MsgInterfaceProtoName)
+
+	for i, msgName := range msgProtoNames {
+		// skip external unregistered msgs
+		if msgName == sdk.MsgTypeURL(&evmtypes.MsgEthereumTx{}) ||
```
pirtleshell, line 85: i'd love if this was a list called `unregisteredMsgs` or something similar

pirtleshell (approved):
i register this type in the upcoming PR with the handler implementation. 😄 maybe that's unorthodox!? but this is a great test guaranteeing we're registering messages as soon as they're introduced.

nddeluca (approved):

app/app_test.go
```diff
@@ -63,6 +69,64 @@ func TestExport(t *testing.T) {
 	assert.Len(t, exportedApp.Validators, 1) // no validators set in default genesis
 }
 
+// TestLegacyMsgAreAminoRegistered checks if all known msg types are registered on the app's amino codec.
+// It doesn't check if they are registered on the module codecs used for signature checking.
```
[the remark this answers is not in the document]
rhuairahrighairidh: yeah, I was thinking we could add tests to the modules to cover them. But if we replace them with a global codec as they've done in v0.46 (https://github.com/cosmos/cosmos-sdk/pull/11240) we could run this test on that global codec.

app/app_test.go
rhuairahrighairidh: thanks, I started with `PanicsWithValue/Error` but the panic returns an error msg containing the go type name of the msg (eg `"types.MsgPlaceBid"`). I couldn't think of a way to reconstruct it to do the comparison.

app/app_test.go
rhuairahrighairidh: agreed, it's a bit confusing, changed to `protoRegisteredMsgs` as at this point it's unclear if these msgs are all registered on amino or not

PR #1612, opened 2023-06-02
test(e2e): add tests for cosmos coin conversion

pirtleshell:
## Description
Adds e2e tests for cosmos coin to/from erc20 conversion.
Tests signing messages via SDK and EIP712.
Kvtool master template used by e2e tests has been updated to include a denom that has conversion enabled by default and EIP712 signing allowed.
When the upgrade handler is written it will use the v0.23 template which does not have these features enabled. In that way, we'll be able to e2e test that the upgrade handler properly initializes the evmutil param & adds the messages to allowed EIP712 messages.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,7 @@ build/bin
 # Ignore build dirs
 build/darwin
 build/linux
+
+# Go workspace files
+go.work
+go.work.sum
--- a/tests/e2e/e2e_convert_cosmos_coins_test.go
+++ b/tests/e2e/e2e_convert_cosmos_coins_test.go
@@ -1,0 +1,118 @@
+package e2e_test
+
+import (
+	"context"
+	"math/big"
+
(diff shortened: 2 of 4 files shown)


galxy25 (approved):

nddeluca (approved):
Nice the additional eip712 is great.

Would it be tracking additional e2e tests like:
- A user can't call mint & burn on the deployed erc20 contract
- A user can't bridge back more tokens than they own
- An e2e test exercising the erc20 approve permissions
- A more complex scenario where user A bridges to erc20, transfers those to user B, and that user B can bridge back to cosmos.

I know these are unit tested well, but regressions can occur during refactors/upgrades around these things and top level tests are nice to ensure not only this feature works, but it stays secure.

rhuairahrighairidh (approved):

PR #1620, opened 2023-06-08
patch barberry

sontrinh16:
## Description
Barberry patch for Kava. For more detail [https://forum.cosmos.network/t/cosmos-sdk-security-advisory-barberry/10825](url)

## Checklist


nddeluca:
ty @sontrinh16, we actually need to update our cosmos-sdk fork that is in the replace statement.  Also, we are unaffected so updating is not an immediate concern

PR #1624, opened 2023-06-09
fix(cli): Resolve problem with assert-invariants cmd

pirtleshell:
## Description
The `assert-invariants` command works by initializing a `TestApp` with the genesis file state, which runs the invariant checks and `InitGenesis`. When we added a validator to all `TestApp` instances in #1477, we began modifying state of the genesis being checked in `assert-invariants`.

This PR makes adding the validator optional so that `assert-invariants` checks only state in the actual genesis file.

Additionally, removes unused no-op `migrate` command and moves the `assert-invariants` command into the `cmd` directory.

## Checklist


nddeluca (approved):
looks good ty for the fix

PR #1627, opened 2023-06-15
test(e2e): refactor NodeRunner in prep for live networks

pirtleshell:
## Description
This PR refactors the e2e framework in anticipation of adding support for running the e2e suite against live networks.
There are no functional changes. The `KavaNodeRunner` has been refactored and renamed to `KvtoolRunner`. The setup of this node runner is dictated by a new `E2E_RUN_KVTOOL_NETWORKS` env variable.

An implementation of a new `NodeRunner` that connects to live networks to come in a separate PR.

## Checklist

### DIFF
--- a/tests/e2e/.env
+++ b/tests/e2e/.env
@@ -2,6 +2,10 @@
 # Should be funded with KAVA and have an ERC20 balance
 E2E_KAVA_FUNDED_ACCOUNT_MNEMONIC='tent fitness boat among census primary pipe nose dream glance cave turtle electric fabric jacket shaft easy myself genuine this sibling pulse word unfold'
 
+# E2E_RUN_KVTOOL_NETWORKS is the only currently supported option.
+# It triggers tests to be run against local networks spun up with kvtool.
+E2E_RUN_KVTOOL_NETWORKS=true
+
 # E2E_KVTOOL_KAVA_CONFIG_TEMPLATE is the kvtool template used to start the chain. See the `kava.configTemplate` flag in kvtool.
 # Note that the config tempalte must support overriding the docker image tag via the KAVA_TAG variable.
 E2E_KVTOOL_KAVA_CONFIG_TEMPLATE="master"
--- a/tests/e2e/runner/chain.go
+++ b/tests/e2e/runner/chain.go
@@ -59,7 +59,7 @@ func (c *Chains) Register(name string, chain *ChainDetails) error {
 // the Chain details are all hardcoded because they are currently fixed by kvtool.
 // someday they may be accepted as configurable parameters.
 var (
-	kavaChain = ChainDetails{
+	kvtoolKavaChain = ChainDetails{
 		RpcUrl:    "http://localhost:26657",
 		GrpcUrl:   "http://localhost:9090",
 		EvmRpcUrl: "http://localhost:8545",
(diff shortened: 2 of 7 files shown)


galxy25 (approved):
the best software is the most re-used software. Great to see all the investment put into local testing paying out dividends for automated integration testing of our production and production like environments.

tests/e2e/.env
```diff
@@ -2,6 +2,10 @@
 # Should be funded with KAVA and have an ERC20 balance
 E2E_KAVA_FUNDED_ACCOUNT_MNEMONIC='tent fitness boat among census primary pipe nose dream glance cave turtle electric fabric jacket shaft easy myself genuine this sibling pulse word unfold'
 
+# E2E_RUN_KVTOOL_NETWORKS is the only currently supported option.
+# It triggers tests to be run against local networks spun up with kvtool.
+E2E_RUN_KVTOOL_NETWORKS=true
```
nddeluca, line 7: Are you thinking that this will be a `E2E_NODE_NETWORK=kvtool|local|mainnet` type thing when more networks are added or use a config file `E2E_CHAIN_CONFIG_PATH=kvtool.json|mainnet.yaml|testnet.tomlsomeother.combo` or something else?

tests/e2e/runner/chain.go
```diff
@@ -60,23 +57,20 @@ func (c *Chains) Register(name string, chain *ChainDetails) error {
 }
 
 // the Chain details are all hardcoded because they are currently fixed by kvtool.
-// some day they may be configurable, at which point `runner` can determine the ports
-// and generate these details dynamically
+// someday they may be accepted as configurable parameters.
 var (
-	kavaChain = ChainDetails{
-		RpcPort:  "26657",
-		RestPort: "1317",
```
nddeluca, line 68: REST is essentially a GRPC wrapper with legacy endpoints removed, so makes sense to remove if we don't care about explicitly testing grpc gateway.  I only see a couple reasons why we'd test with it, but it's low priority.

Ethereum websocket (8546) may be added in future.  Not saying we include it here, just mentioning for completeness.

tests/e2e/runner/kvtool.go
```diff
@@ -0,0 +1,104 @@
...
+func NewKvtoolRunner(config KvtoolRunnerConfig) *KvtoolRunner {
+	return &KvtoolRunner{
+		config: config,
+	}
+}
+
+func (k *KvtoolRunner) StartChains() Chains {
+	// install kvtool if not already installed
+	installKvtoolCmd := exec.Command("./scripts/install-kvtool.sh")
+	installKvtoolCmd.Stdout = os.Stdout
```
nddeluca, line 44: I had to look it up -- forgot that stdout is not buffered in golang, so panic on error below won't hide logs 👍

nddeluca (approved):
nice addition & refactor

tests/e2e/.env
pirtleshell: in the branch off this that runs tests on external networks, i just set `E2E_RUN_KVTOOL_NETWORKS` to `false` and then it expects `E2E_KAVA_RPC_URL`, `E2E_KAVA_GRPC_URL`, & `E2E_KAVA_EVM_RPC_URL` env variables to setup the network connection.

i agree it'd be good to extract this into some sort of network configuration file we can add to gitignore to prevent accidental commit of secrets to the repo.

tests/e2e/runner/chain.go
pirtleshell: yup, REST was unused and permanently removed here.

could stand to remove RPC as well, but it's currently setup to ping the `/status` endpoint.
i have a note to self to refactor it into a call to tendermint's node_info grpc.

PR #1655, opened 2023-07-21
fix(evmutil): create module account on InitGenesis

pirtleshell:
## Description

ensures the creation of the x/evmutil module account on init genesis.

this is non-state breaking because mainnet will never call init genesis. additionally, the module account already exists on mainnet.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,11 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 ## [Unreleased]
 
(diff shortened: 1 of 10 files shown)


galxy25 (approved):

rhuairahrighairidh (approved):
nice, I reviewed it against the security questionnaire and merge checklist and everything looks good. Thought there was some state breaking changes, but error text is non breaking, and `unpackERC20ResToBigInt` is only used by queriers.

PR #1667, opened 2023-08-23
Fix kava version for internal testnet

evgeniy-scherbina:


galxy25 (approved):

galxy25:
@evgeniy-scherbina also https://github.com/Kava-Labs/kava/blob/master/ci/env/kava-protonet/KAVA.VERSION

PR #1692, opened 2023-08-31
Rocksdb Metrics

evgeniy-scherbina:


cmd/kava/opendb/run.sh
```diff
@@ -0,0 +1,3 @@
+CGO_CFLAGS="-I/opt/homebrew/opt/rocksdb/include" \
+CGO_LDFLAGS="-L/opt/homebrew/opt/rocksdb/lib -lrocksdb -lstdc++ -lm -lz -L/opt/homebrew/opt/snappy/lib -L/opt/homebrew/opt/lz4/lib -L/opt/homebrew/opt/zstd/lib" \
+  go test -v -tags=rocksdb
```
pirtleshell, line 3: i think this would be better added to the root dir's makefile which handles cross platform CGO flags and is already included in CI pipelines

cmd/kava/opendb/opendb_rocksdb.go
```diff
@@ -42,11 +43,15 @@ const (
 
 	defaultColumnFamilyName = "default"
 
-	maxOpenFilesDBOptName          = "max_open_files"
-	maxFileOpeningThreadsDBOptName = "max_file_opening_threads"
+	enableMetricsOptName             = "rocksdb.enable_metrics"
```
galxy25, line 46: nit: do we need to prefix these with `rocksdb`? Would prefer following existing code convention for the sake of code consistency / coherent read-ablity

go.mod
```diff
@@ -19,6 +19,7 @@ require (
 	github.com/grpc-ecosystem/grpc-gateway v1.16.0
 	github.com/linxGnu/grocksdb v1.8.0
 	github.com/pelletier/go-toml/v2 v2.0.6
+	github.com/pkg/errors v0.9.1
```
pirtleshell, line 22: it seems this package is unmaintained these days: https://github.com/pkg/errors

the cosmos-sdk's errors package provides a similar [`Wrap` method](https://pkg.go.dev/cosmossdk.io/errors#Wrap), albeit with some extra ABCI context.

cmd/kava/opendb/stats_loader.go
```diff
@@ -0,0 +1,264 @@
...
+type float64Histogram struct {
+	Sum   float64
+	Count float64
+	P50   float64
+	P95   float64
+	P99   float64
+	P100  float64
+}
+
+func (l *statLoader) error() error {
```
pirtleshell, line 147: i like this error collection 👍

cmd/kava/opendb/opendb_rocksdb.go
```diff
@@ -168,3 +197,42 @@ func newDefaultOptions() *grocksdb.Options {
...
+// reportMetrics periodically requests stats from rocksdb and reports to prometheus
+// NOTE: should be launched as a goroutine
+func reportMetrics(db *grocksdb.DB, interval time.Duration) {
+	ticker := time.NewTicker(interval)
+	for {
+		select {
+		case <-ticker.C:
+			props, stats, err := getPropsAndStats(db)
+			if err != nil {
+				continue
```
galxy25, line 210: nit: no logging in any of these changes, I know this code is simple enough that it will probably "just work", but I think it's a good habit to add logging by default so there is more information for future developers for when things go wrong / need to be debugged (e.g. having a log statement on whether the metrics are enabled or not to allow the debugger to rule out issues related to setting up the metric routine, or here in this line to capture the error so it can be used to resolve the problem - versus throwing it away and losing information)

cmd/kava/opendb/opendb_rocksdb_test.go
```diff
@@ -337,7 +337,7 @@ func TestNewRocksDBWithOptions(t *testing.T) {
 	cfOpts := newDefaultOptions()
 	cfOpts.SetWriteBufferSize(999_999)
 
-	db, err := newRocksDBWithOptions(name, dir, dbOpts, cfOpts)
+	db, err := newRocksDBWithOptions(name, dir, dbOpts, cfOpts, true, defaultReportMetricsIntervalSecs)
```
galxy25, line 340: Nit: do we need to extend this method signature or can these opts be added onto the dbOpts struct?

cmd/kava/opendb/opendb_rocksdb.go
galxy25: Update: after reviewing more of the files I see this is the current convention, ignore me

cmd/kava/opendb/props_loader.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	return props, nil
+}
+
+func (l *propsLoader) getProperty(propName string) string {
+	value := l.db.GetProperty(propName)
+	if value == "" {
+		l.errorMsgs = append(l.errorMsgs, fmt.Sprintf("property %v is empty", propName))
+		return ""
```
galxy25, line 58: This is c style coding, go idiom is to return `value, error` and have callers check if err is nil before using value

cmd/kava/opendb/opendb_rocksdb.go
pirtleshell: these are app.toml properties accessed through AppOptions, yes? i think they need to be namespaced to something. makes sense to me to put the rocksdb-specific metrics settings with the rocksdb configuration 
could also see them going under the existing `telemetry`, but separating into a `rockdb` object we control seems good to

do you have an example of the convention you're talking about, @galxy25? it's posisble i'm not understanding what these constants are

cmd/kava/opendb/props_loader.go
galxy25: https://go.dev/blog/error-handling-and-go

pirtleshell (approved):

cmd/kava/opendb/stats_loader.go
```diff
@@ -0,0 +1,264 @@
+//go:build rocksdb
+// +build rocksdb
+
+package opendb
+
+import (
+	"strconv"
+
+	"github.com/pkg/errors"
```
galxy25, line 9: Nit: use standard library functionality https://go.dev/blog/go1.13-errors instead

cmd/kava/opendb/run.sh
evgeniy-scherbina: I agree

galxy25 (approved):

cmd/kava/opendb/opendb_rocksdb.go
evgeniy-scherbina: looks like this in `app.toml`:
```toml
[rocksdb]
enable_metrics = true
report_metrics_interval_secs = 150

max_open_files = 142
write_buffer_size = 67108142
```

galxy25: @pirtleshell here https://github.com/Kava-Labs/kava/pull/1692/files#diff-ad965588a6a423832eb132b027de107d69a3a60df814a509acb511a50a6726f1R32 Where `rocksdb` prefix is also used, versus the variables that previously existed in this `opendb_rocksdb.go` file e.g. `write_buffer_size`

evgeniy-scherbina: I think I have to use `dashes` instead of `underscores`:
- `max_open_files = 142` -> `max-open-files = 142`

Initially I used `underscores` because `rocksdb` uses them, but anyway we probably have to follow our convention

go.mod
evgeniy-scherbina: I'll consider using stdlib errors instead

cmd/kava/opendb/props_loader.go
evgeniy-scherbina: generally I agree, but I believe in some cases such approach make sense to allow code like this:
https://github.com/Kava-Labs/kava/blob/yevhenii/rocksdb-metrics/cmd/kava/opendb/stats_loader.go#L157-L193

otherwise instead of `40 lines of code`, I'll have `120 lines of code` (most of them will be error-checking)

also I put some comments regarding this:
`// NOTE: some methods accumulate errors instead of returning them, these methods are private and not intended to use outside`

cmd/kava/opendb/opendb_rocksdb_test.go
evgeniy-scherbina: They can't be added to `dbOpts`, but they are coming from `appOpts`, so I can change signature from:
```go
func newRocksDBWithOptions(..., enableMetrics bool, reportMetricsIntervalSecs int64)
```
to
```go
func newRocksDBWithOptions(..., appOpts types.AppOptions)
```

but I wasn't sure about this because it's a bit harder to `unit-test` such function, because you have to use `MockAppOptions` instead of just passing args so I decided to parse `appOpts` in the highest-level function

but it's debatable decision

cmd/kava/opendb/opendb_rocksdb.go
evgeniy-scherbina: @galxy25 fyi: https://github.com/Kava-Labs/kava/pull/1692/files#diff-ad965588a6a423832eb132b027de107d69a3a60df814a509acb511a50a6726f1R32 - this is not `appOpts`, this is `rocksdb properties`, just want to make sure we're on the same page. Their names are hardcoded on rocksdb-level

cmd/kava/opendb/run.sh
evgeniy-scherbina: fixed

cmd/kava/opendb/opendb_rocksdb.go
evgeniy-scherbina: fixed

evgeniy-scherbina: as of now looks like this:
```toml
[rocksdb]
enable-metrics = true
report-metrics-interval-secs = 15

max-open-files = 142
write-buffer-size = 67108142
```

go.mod
evgeniy-scherbina: fixed

cmd/kava/opendb/stats_loader.go
evgeniy-scherbina: fixed

cmd/kava/opendb/opendb_rocksdb.go
evgeniy-scherbina: @galxy25 
yeah, make sense
could you pls. advise what library to use, and point to an example in kava repo?
I want to make sure logging is consistent across kava repo

drklee3 (approved):

PR #1704, opened 2023-09-08
feat(community): add switchover param

rhuairahrighairidh:
Add a timestamp to the communiity params for when the incentive changes take effect.

## Description

This uses the new style params instead of the params module. It makes it slightly more inconvenient to do a gov proposal, but saves re-writing the params when we eventually upgrade the sdk.

## Checklist
 - [x] PR is against master or a release branch (not feature branch)
 - [x] Security questionnaire reviewed for changes to answers since step 1
 - [x] Code is well tested
   - All public methods have unit tests
   - New features/capabilities have an e2e test
   - Migrations are unit tested
 - [x] [Blockchain and Cosmos SDK Gotchas](https://kava-labs.atlassian.net/wiki/spaces/ENG/pages/1174962181/WIP+Protocol+Code+Contribution+Guidelines#Blockchain-and-Cosmos-SDK-Gotchas) reviewed
 - [ ] ~~Migrations written~~ Leaving migrations out to unblock @drklee3 on #1707 
   - ~~in migration files added directly to master branch, write module migrations generically so both mainnet and testnet can use them. if the change is consensus breaking to a module, the module’s consensus version is incremented and the migration is registered to the module’s services.~~
   - ~~migrations are fully unit tested to ensure state changes are made correctly~~
   - ~~record any release branch migrations needed in the feature tech spec document, to be added when the release branch is created~~
 - [x] Protonet testnet genesis file updated
 - [x] Documentation updated (module spec, proto file comments, and godoc comments)
   - I'll leave spec updates to a later PR. I've added a card to track
 - [x] APIs have been updated and are backwards compatible
   - APIs: evm json-rpc, rest, grpc
 - [x] CLI commands have been updated


proto/kava/community/v1beta1/params.proto
```diff
@@ -0,0 +1,15 @@
+syntax = "proto3";
+package kava.community.v1beta1;
+
+import "gogoproto/gogo.proto";
+import "google/protobuf/timestamp.proto";
+
+option go_package = "github.com/kava-labs/kava/x/community/types";
+
+// Params defines the parameters of the community module.
+message Params {
```
DracoLi, line 10: should a query be added for getting x/community params?

rhuairahrighairidh: yeah, realised there a few missing pieces

DracoLi (approved):

x/community/types/params.go
```diff
@@ -0,0 +1,26 @@
+package types
+
+import "time"
+
+var (
+	DefaultUpgradeTimeDisableInflation = time.Time{}
```
drklee3, line 6: Should we use a non-zero/default value for the default time and enforce it in `Validate()`? 

In case we're doing `Time.After()` or `Time.Before()` checks when using the value on switchover, empty value might cause issues since it's considered earlier

DracoLi: The inflation upgrade logic will handle unset times by upgrading if unset to ensure vanilla kava chain always have new inflation logic enabled. https://github.com/Kava-Labs/kava/pull/1706/files#diff-15d449f4515711b278439f4600052557555689d56ea115a58619aa3061738906R30

DracoLi (commented):
does the internal testnet genesis also need to be updated?

x/community/types/params.go
nddeluca: Agree with @DracoLi here, new chains should default to new logic to match mainnet behavior after the upgrade time

rhuairahrighairidh:
> does the internal testnet genesis also need to be updated?

I updated protonet genesis, going off the assumption that protonet follows master, and internal testnet follows latest release.

ci/env/kava-protonet/KAVA.VERSION
```diff
@@ -1 +1 @@
-47416b3d1af6dfd31bab479d78b24cdbb677be31
+master
```
rhuairahrighairidh, line 1: @evgeniy-scherbina I saw you pinned this commit recently, is it ok to change back to master?

evgeniy-scherbina: yes

x/community/keeper/grpc_query.go
```diff
@@ -18,6 +21,20 @@ func NewQueryServerImpl(k Keeper) types.QueryServer {
 	return &queryServer{keeper: k}
 }
 
+// Balance implements the gRPC service handler for querying x/community params.
```
pirtleshell, line 24:
```suggestion
// Params implements the gRPC service handler for querying x/community params.
```

x/community/types/genesis.go
```diff
@@ -0,0 +1,20 @@
+package types
```
pirtleshell, line 1: though super simple, i think these public methods should have tests.

x/community/types/params.go
```diff
@@ -0,0 +1,26 @@
...
+// DefaultParams returns default params
+func DefaultParams() Params {
+	return NewParams(
+		DefaultUpgradeTimeDisableInflation,
+	)
+}
+
+// Validate checks the params are valid
+func (p Params) Validate() error {
+	return nil
```
pirtleshell, line 25: i'd add a comment here noting why `p.UpgradeTimeDisableInflation.IsZero()` is still a valid state (zero time means disable inflation on the block 1)

pirtleshell (approved):
looks good 👍

x/community/types/params.go
rhuairahrighairidh: added some doc comments to the proto files where the param is defined

drklee3 (approved):
kvtool master template genesis needs to be updated to include the new params to fix the e2e test

PR #1706, opened 2023-09-13
feat: disable inflation upgrade logic

DracoLi:
## Description

Add begin blocker logic to `x/community` to handle the disable inflation upgrade logic that disables inflation from `x/mint` and `x/kavadist` on upgrade time.

## Checklist
 - [x] Code is well tested
 - [x] [Blockchain and Cosmos SDK Gotchas](https://kava-labs.atlassian.net/wiki/spaces/ENG/pages/1174962181/WIP+Protocol+Code+Contribution+Guidelines#Blockchain-and-Cosmos-SDK-Gotchas) reviewed


drklee3 (commented):
looks good after tests are added

karzak (changes requested):

x/community/keeper/incentives.go
```diff
@@ -0,0 +1,58 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// PayCommunityRewards pays rewards to the fee collector account
+func (k Keeper) PayCommunityRewards(ctx sdk.Context) error {
+	return nil
+}
+
+// DisableInflationAfterUpgrade disables inflation on or after the upgrade time
+func (k Keeper) DisableInflationAfterUpgrade(ctx sdk.Context) {
```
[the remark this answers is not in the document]
DracoLi: yep, i can split this out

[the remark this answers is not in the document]
DracoLi: for e2e tests, I think we should wait until both inflation upgrade & pay now rewards are done. then we can add an e2e test for the whole upgrade.

drklee3 (approved):

app/app.go
```diff
@@ -665,6 +655,19 @@ func NewApp(
...
+	app.communityKeeper = communitykeeper.NewKeeper(
+		appCodec,
+		keys[communitytypes.StoreKey],
+		app.accountKeeper,
+		app.bankKeeper,
+		&cdpKeeper,
+		app.distrKeeper,
+		&hardKeeper,
+		&app.mintKeeper,
+		&app.kavadistKeeper,
```
pirtleshell, line 668: i didn't mention it in #1704 , but seeing this Pr makes me reconsider.. i'd understand if we're too far down this path to turn back, but did we consider putting the disable inflation logic in `x/kavadist` instead of `x/community`? it feels like the community module shouldn't be responsible for inflation.

it was my understanding from the picture we discussed in the design meeting that the logic would be getting added to kavadist. doing so would simplify the community module and leave kavadist with the same level of complexity.

i understand if we're already underway and committed to this approach. it seems worth noting that putting the logic in `x/community` means we add a new store, add a new begin blocker, and further tangle the web of where and how inflation is managed in kava.

x/community/keeper/incentives.go
```diff
@@ -0,0 +1,58 @@
...
+	blockTime := ctx.BlockTime()
+	upgradeTime := params.UpgradeTimeDisableInflation
+
+	// a vanilla kava chain should disable inflation on the first block if `upgradeTime` is not set.
+	// thus, we don't upgrade if `upgradeTime` is set and `blockTime` is before `upgradeTime`.
+	if !upgradeTime.IsZero() && blockTime.Before(upgradeTime) {
+		return false
+	}
+
+	return true
```
pirtleshell, line 29: this logic may be more clear if this whole block were changed to
```suggestion
	// a vanilla kava chain should disable inflation on the first block if `upgradeTime` is not set.
	if upgradeTime.IsZero() {
	  return true
	}
	
	return !blockTime.Before(upgradeTime)
```

app/app.go
DracoLi: one note here is we already need to the new blocker and new store for community for the distribution of the rewards that is a part of the post disable inflation logic. Having them together in a single module does make it a bit easier to manage since the disable inflation upgrade and the post rewards distribution are related.

x/community/keeper/incentives.go
DracoLi: yep its much cleaner this way. thx, will update

app/app.go
pirtleshell: gotcha, the initial design diagram showed kavadist pulling the necessary funds from community and distributing them to the fee pool for use in staking rewards. in that setup, there would still be no begin blocker for x/community. the community module would be the community pool of funds (and functionality regarding its use) and kava dist would be for kava distributions.

pirtleshell: all this may be moot anyhow, as we are fairly far down the dev path for this implementation.

x/community/abci_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+		kavadistParams = suite.App.GetKavadistKeeper().GetParams(suite.Ctx)
+		suite.True(kavadistParams.Active)
+	})
+}
+
+func (suite *ABCITestSuite) setUpgradeTimeFromNow(t time.Duration) {
+	suite.Keeper.SetParams(
+		suite.Ctx,
+		types.Params{UpgradeTimeDisableInflation: suite.Ctx.BlockTime().Add(t)},
```
drklee3, line 117: Should be updated to include the new `StakingRewardsPerSecond` param, test is panicking with it being nil

drklee3:
Seems like the changes are also causing several tests in `x/incentive` to fail, with the validator rewards being changed 😰

nddeluca:
@rhuairahrighairidh @drklee3 This is updated and is mergable pending your reviews.

Updates I made:
  - Unit tests are now shared between keeper and abci begin blocker since behavior is 100% the same.
  - ABCI is unaware of keeper initial keeper logic branch (keeper methods required to be called in certain order by begin blocker)
  - Behavior of zero time is changed -- this now doesn't run for vanilla changes.  This is more ideal for new chains as they should set genesis for desired inflation behavior and not rely on this logic.  We will use this logic on mainnet, then delete it next release.  In addition, it simplifies the implementation to not need a previous block time or other flag.
  - Begin blocker now panics if parameters are not in state to catch errors in migrations/upgrades/etc

drklee3 (approved):
looks good - also nice to have incentive tests unaffected now

PR #1707, opened 2023-09-13
Add `RewardsPerSecond` param to `x/community` module

drklee3:
## Description
- Add param to `x/community` for payout amount per second.
- Add `744191` per second reward to protonet genesis (~4.6 KAVA per block, using 23,471,808 KAVA per year)
- Add migration to set parameters in state

## Checklist


proto/kava/community/v1beta1/params.proto
```diff
@@ -12,4 +13,6 @@ message Params {
     (gogoproto.stdtime) = true,
     (gogoproto.nullable) = false
   ];
+
+  cosmos.base.v1beta1.Coin rewards_per_second = 2 [(gogoproto.nullable) = false];
```
DracoLi, line 17: do we even want to support minting non-ukava coins? If not, perhaps its better to just make this an int?

x/community/types/params.go
```diff
@@ -1,26 +1,39 @@
...
+	fmt "fmt"
+	"time"
+
+	sdkmath "cosmossdk.io/math"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
 
 var (
 	DefaultUpgradeTimeDisableInflation = time.Time{}
+	DefaultRewardsPerSecond            = sdk.NewCoin("ukava", sdkmath.ZeroInt())
```
DracoLi, line 13: thoughts on setting this to a non-zero default? I think the default chain should mint here to align with mainnet behaviour.

DracoLi (commented):
i believe we should also update protonet and internal testnet genesis

karzak (commented):
Does this need a migration?

proto/kava/community/v1beta1/params.proto
drklee3: Yup we can use an int instead

drklee3:
> Does this need a migration?

Yup, will check if @rhuairahrighairidh will be adding migrations for the existing params in the target branch that I can add to or if they'll be done in a separate PR

x/community/types/params.go
drklee3: Updated to be target amount

x/community/types/params.go
```diff
@@ -1,26 +1,43 @@
...
 }
 
 // Validate checks the params are valid
 func (p Params) Validate() error {
+	if p.RewardsPerSecond.IsNil() {
+		return errors.New("rewards per second should not be nil")
+	}
+
+	if p.RewardsPerSecond.IsNegative() {
+		return errors.New("rewards per second should not be negative")
```
galxy25, line 39: nit: include the actual value in this path?

galxy25 (approved):

proto/kava/community/v1beta1/params.proto
nddeluca: Agree with using `int` here -- though should we rename this paramter to reflect that? `StakingRewardsPerSecond` or something to make it clear this is what the instended use is and thus using the staking denom?

nddeluca: I also don't like using the same parameter name across modules with different types.  RewardsPerSecond is coins in other places but int here

nddeluca: So imo a re-name also makes sense but open to other thoughts as well

drklee3: `StakingRewardsPerSecond` sounds good to be more clear

x/community/types/params.go
```diff
@@ -1,26 +1,47 @@
...
+	fmt "fmt"
+	"time"
+
+	sdkmath "cosmossdk.io/math"
+)
 
 var (
 	DefaultUpgradeTimeDisableInflation = time.Time{}
+	// DefaultStakingRewardsPerSecond is ~4.6 KAVA per block, 6.3s block time
+	DefaultStakingRewardsPerSecond = sdkmath.NewInt(744191)
```
DracoLi, line 14: might be worth mentioning this new default in the changelog as this will be the new inflation for the default chain going forward.

DracoLi (approved):

PR #1731, opened 2023-09-20
feat: payout staking rewards

DracoLi:
## Description
Payout staking rewards from x/community

Some remaining issues still needs to be resolved:
- How payout should work on upgrade block. Currently it will payout 0 rewards on upgrade block. This behaviour needs to account for both mainnet and a vanilla chain.
- How staking rewards payout work on vanilla chain since x/community will not be funded by default.

## Checklist
 - [x] Tests added for incentives and abci


x/community/keeper/incentives.go
```diff
@@ -1,19 +1,81 @@
...
+		return sdk.Coin{}, errorsmod.Wrapf(types.ErrStakingRewardsPayout, "missing previous block time")
+	}
+
+	params, found := k.GetParams(ctx)
+	if !found {
+		return sdk.Coin{}, errorsmod.Wrapf(types.ErrStakingRewardsPayout, "x/community params not found")
+	}
+
+	blockTime := ctx.BlockTime()
+	secondsElapsed := sdkmath.NewInt(blockTime.Unix() - previousBlockTime.Unix())
```
drklee3, line 61: Should we use a more precise value for the secondsElapsed? `sdk.Dec` with blocktime ms instead of seconds, then truncate reward coins decimals after multiplying

nddeluca:
superseded by #1742

PR #1729, opened 2023-09-18
feat(community): consolidate community funds

drklee3:
## Description
- Transfer `x/distribution` communitypool and `x/kavadist` (non-ukava) account funds to `x/community`

## Checklist
 - [x] Changelog has been updated as necessary.

PR #1743, opened 2023-10-03
Emit events for inflation disable and staking rewards

drklee3:
## Description
- Emits event when inflation is disabled
- Emits event on each staking reward payout

## Checklist

### DIFF
--- a/app/test_common.go
+++ b/app/test_common.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"math/rand"
+	"reflect"
 	"testing"
 	"time"
 
@@ -535,4 +536,32 @@ func NewFundedGenStateWithSameCoinsWithModuleAccount(cdc codec.JSONCodec, coins 
 	builder.WithSimpleModuleAccount(modAcc.Address, nil)
 
 	return builder.BuildMarshalled(cdc)
+}
+
+// EventsContains returns an error if the expected event is not in the provided events
+func EventsContains(events sdk.Events, expectedEvent sdk.Event) error {
+	foundMatch := false
+	for _, event := range events {
+		if event.Type == expectedEvent.Type {
+			if reflect.DeepEqual(attrsToMap(expectedEvent.Attributes), attrsToMap(event.Attributes)) {
+				foundMatch = true
+			}
+		}
+	}
+
+	if !foundMatch {
+		return fmt.Errorf("event of type %s not found or did not match", expectedEvent.Type)
+	}
+
+	return nil
+}
+
+func attrsToMap(attrs []abci.EventAttribute) []sdk.Attribute { // new cosmos changed the event attribute type
+	out := []sdk.Attribute{}
+
+	for _, attr := range attrs {
+		out = append(out, sdk.NewAttribute(string(attr.Key), string(attr.Value)))
+	}
+
+	return out
 }
--- a/x/community/keeper/staking.go
+++ b/x/community/keeper/staking.go
@@ -52,13 +52,14 @@ func (k Keeper) PayoutAccumulatedStakingRewards(ctx sdk.Context) {
 			panic(err)
 		}
 
-		// emit event
+		// emit event with amount transferred
 		ctx.EventManager().EmitEvent(
 			sdk.NewEvent(
 				types.EventTypeStakingRewardsPaid,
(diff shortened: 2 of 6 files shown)


app/test_common.go
```diff
@@ -536,3 +537,31 @@ func NewFundedGenStateWithSameCoinsWithModuleAccount(cdc codec.JSONCodec, coins
...
+
+func attrsToMap(attrs []abci.EventAttribute) []sdk.Attribute { // new cosmos changed the event attribute type
+	out := []sdk.Attribute{}
+
+	for _, attr := range attrs {
+		out = append(out, sdk.NewAttribute(string(attr.Key), string(attr.Value)))
+	}
+
+	return out
+}
```
drklee3, line 567: copied from https://github.com/Kava-Labs/kava/pull/1729

x/community/keeper/staking.go
```diff
@@ -61,7 +61,13 @@ func (k Keeper) PayoutAccumulatedStakingRewards(ctx sdk.Context) {
 	// save state
 	k.SetStakingRewardsState(ctx, state)
 
-	return
+	// emit event
+	ctx.EventManager().EmitEvent(
```
drklee3, line 65: do we want to still emit events even if there are no payouts? or only when truncatedRewards is non-zero

nddeluca: I'm split but leaning towards only emit if non-zero and we can move this into the if statement above.

nddeluca: On deeper thought, ya let's only emit when we actually transfer funds

x/community/keeper/staking.go
```diff
@@ -51,6 +51,14 @@ func (k Keeper) PayoutAccumulatedStakingRewards(ctx sdk.Context) {
 			// occur in cases where the chain is running in an invalid state
 			panic(err)
 		}
+
+		// emit event
+		ctx.EventManager().EmitEvent(
+			sdk.NewEvent(
+				types.EventTypeStakingRewardsPaid,
+				sdk.NewAttribute(types.AttributeKeyStakingRewardAmount, sdk.NewCoin(stakingRewardDenom, truncatedRewards).String()),
```
nddeluca: I think we can use ` transferAmount.String())` here and get the same output?

nddeluca: Pushed this change

x/community/keeper/staking.go
```diff
@@ -60,8 +68,6 @@ func (k Keeper) PayoutAccumulatedStakingRewards(ctx sdk.Context) {
 
 	// save state
 	k.SetStakingRewardsState(ctx, state)
-
-	return
```
nddeluca, line 64: Ah nice right don't need explicit return

x/community/testutil/staking_rewards.go
```diff
@@ -258,10 +258,15 @@ func (suite *stakingRewardsTestSuite) TestStakingRewards() {
...
 			keeper.SetParams(ctx, params)
 
+			stakingRewardEvents := sdk.Events{}
+
 			for {
 				// run community begin blocker logic
 				suite.testFunc(ctx, keeper)
 
+				// accumulate event rewards from events
+				stakingRewardEvents = append(stakingRewardEvents, filterStakingRewardEvents(ctx.EventManager().Events())...)
```
nddeluca, line 268: I made this a little cleaner by moving the event accumulation above the break statement to catch all testFunc runs, in addition collected the events so we can add the assertion for no tracking for zero rewards

x/community/testutil/staking_rewards.go
```diff
@@ -287,6 +292,20 @@ func (suite *stakingRewardsTestSuite) TestStakingRewards() {
...
 			suite.Equal(tc.expectedRewardsTotal.String(), feeCollectorBalanceAdded.String(), "expected fee collector balance to match")
 
+			if tc.expectedRewardsTotal.IsZero() {
+				suite.Equal(0, len(stakingRewardEvents), "expected no events to be emitted")
+			} else {
+				// we add up all reward coin events
+				eventCoins := getRewardCoinsFromEvents(stakingRewardEvents)
+
+				// assert events emitted match expected rewards
+				suite.Equal(
```
nddeluca, line 302: I really like this method of asserting event balances match 👍

nddeluca (approved):

PR #1745, opened 2023-10-05
feat(community): add MsgUpdateParams for governance

drklee3:
## Description
- Add `MsgUpdateParams` and message handler
- Add e2e test for proposal submission

## Checklist


app/app.go
```diff
@@ -667,6 +669,7 @@ func NewApp(
 		&app.mintKeeper,
 		&app.kavadistKeeper,
 		app.stakingKeeper,
+		authtypes.NewModuleAddress(govtypes.ModuleName), // Authority
```
nddeluca, line 672: This is used in quite a few places now -- could we move to a named parameter? `govAuthorityAddr` or similar and reuse?

x/community/types/codec.go
```diff
@@ -12,7 +13,9 @@ import (
 
 // RegisterLegacyAminoCodec registers all the necessary types and interfaces for the module.
 func RegisterLegacyAminoCodec(cdc *codec.LegacyAmino) {
-	cdc.RegisterConcrete(&MsgFundCommunityPool{}, "community/MsgFundCommunityPool", nil)
+	legacy.RegisterAminoMsg(cdc, &MsgFundCommunityPool{}, "community/MsgFundCommunityPool")
+	legacy.RegisterAminoMsg(cdc, &MsgUpdateParams{}, "community/MsgUpdateParams")
```
nddeluca, line 17: Nice thnx for remembering the legacy codec

nddeluca (changes requested):

x/community/keeper/keeper_test.go
```diff
@@ -89,3 +92,56 @@ func (suite *KeeperTestSuite) TestGetAndSetStakingRewardsState() {
...
+		authority sdk.AccAddress
+	}{
+		{
+			name:      "gov",
+			authority: authtypes.NewModuleAddress(govtypes.ModuleName),
+		},
+		{
+			name:      "empty",
+			authority: sdk.AccAddress{},
+		},
```
drklee3, line 116: Do we want to allow empty address for authority?

rhuairahrighairidh: Could validate the address format in `NewKeeper` with `sdk.VerifyAddressFormat`.

proto/kava/community/v1beta1/tx.proto
```diff
@@ -27,3 +31,17 @@ message MsgFundCommunityPool {
...
 message MsgFundCommunityPoolResponse {}
+
+// MsgUpdateParams allows an account to update the community module parameters.
+message MsgUpdateParams {
+  option (gogoproto.goproto_getters) = false;
+
+  // authority is the address that controls the module (defaults to x/gov unless overwritten).
+  string authority = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+
+  // params defines the x/distribution parameters to update.
```
rhuairahrighairidh, line 42: nit: Probably worth making the doc strings clearer (x/community instead of x/distribution?), as we'll likely copy and paste them when we update to the new params in other modules.

tests/e2e/e2e_community_update_params_test.go
```diff
@@ -0,0 +1,168 @@
...
+		Msgs:      []sdk.Msg{&msg},
+		GasLimit:  uint64(gasLimit),
+		FeeAmount: sdk.NewCoins(fee),
+		Memo:      "this is a failure!",
+	}
+	res := kavaAcc.SignAndBroadcastKavaTx(req)
+
+	// ASSERT
+	_, err := util.WaitForSdkTxCommit(suite.Kava.Tx, res.Result.TxHash, 6*time.Second)
+	suite.Error(err)
```
rhuairahrighairidh, line 45: I'd check for a explicit error here. e2e tests could fail for a lot of reasons but we want to check that it's failing because the authority isn't correct.

tests/e2e/e2e_community_update_params_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+	// Unmarshal data to TxMsgData
+	var txMsgData sdk.TxMsgData
+	suite.Kava.EncodingConfig.Marshaler.MustUnmarshal(txResBytes, &txMsgData)
+	suite.T().Logf("txData.MsgResponses: %v", txMsgData.MsgResponses)
+
+	// Parse MsgResponse
+	suite.Kava.EncodingConfig.Marshaler.MustUnmarshal(txMsgData.MsgResponses[0].Value, ptr)
+	suite.Require().NoError(err)
+}
```
rhuairahrighairidh, line 168: nit: might be nice to move this to a stand alone function in e2e/testutil or somewhere so it can be reused in other tests. Auction and committee uses return data.

rhuairahrighairidh (changes requested):

rhuairahrighairidh (approved):

x/community/keeper/msg_server.go
```diff
@@ -47,3 +49,28 @@ func (s msgServer) FundCommunityPool(goCtx context.Context, msg *types.MsgFundCo
...
+// UpdateParams handles UpdateParams msgs.
+func (s msgServer) UpdateParams(
+	goCtx context.Context,
+	msg *types.MsgUpdateParams,
+) (*types.MsgUpdateParamsResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	if s.keeper.GetAuthority().String() != msg.Authority {
+		return nil, errors.Wrapf(
+			govtypes.ErrInvalidSigner,
```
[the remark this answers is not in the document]
drklee3: Hmm I think for now we can be consistent with other modules -- if we switch out gov later then we should also be able to update the error response then, assuming a gov switchout is a significant breaking change

PR #1751, opened 2023-10-20
feat(community): add AnnualizedRewards grpc query

pirtleshell:
## Description
Adds `x/community` query for annualized rewards which gives the annualized staking reward rate

## Checklist


x/community/keeper/grpc_query.go
```diff
@@ -62,3 +63,42 @@ func (s queryServer) TotalBalance(
...
+		StakingRewards: calc.GetAnnualizedRate(),
+	}, nil
+}
+
+// convertDecToLegacyDec is a helper method for converting between new and old Dec types
+// current version of cosmos-sdk in this repo uses sdk.Dec
+// this module uses sdkmath.LegacyDec in its parameters
+// TODO: remove me after upgrade to cosmos-sdk v50 (LegacyDec is everywhere)
+func convertDecToLegacyDec(in sdk.Dec) sdkmath.LegacyDec {
+	return sdkmath.LegacyNewDecFromBigIntWithPrec(in.BigInt(), sdkmath.LegacyPrecision)
```
nddeluca, line 103: This should be the `sdk.Precision` or the precision of the `in` param

x/community/keeper/rewards.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	sdkmath "cosmossdk.io/math"
+)
+
+const SecondsPerYear = 365 * 24 * 3600
+
+// StakingRewardCalculator wraps data and calculation for determining staking rewards
+// It assumes that staking comes from one of two sources depending on if inflation is enabled or not.
+type StakingRewardCalculator struct {
```
nddeluca, line 11: Generally, not a fan of use case style structs like this unless we have to satisfy an interface or the caller doesn't have access to some of the calculation inputs or dependencies. We are also creating a large interface here with public data members and several public methods that aren't used by non-test or test code.

I'd lean towards keeping it simple, pure, and less rigid with something like:
```go
func CalculateStakingAnnualPercentage(totalSupply, totalBonded sdkmath.Int, inflationRate, communityTax, rewardsPerSecond sdkmath.LegacyDec) sdkmath.LegacyDec {
	// no rewards are given if no tokens are bonded, in addition avoid division by zero
	if totalBonded.IsZero() {
		return sdkmath.LegacyZeroDec()
	}

	// the percetage of inflationRate * totalSupply tokens that are distrubed to stakers
	percentInflationDistributedToStakers := sdkmath.LegacyOneDec().Sub(communityTax)

	// the total amount of tokens distributed to stakers in a year
	amountGivenPerYear := inflationRate.
		MulInt(totalSupply).Mul(percentInflationDistributedToStakers).  // portion provided by inflation via mint & distribution modules
		Add(rewardsPerSecond.Mul(sdkmath.LegacyNewDec(SecondsPerYear))) // portion provided by community module

	// divide by bonded tokens or invested tokens to get the percent return
	return amountGivenPerYear.QuoInt(totalBonded)
}

```
Which also allows us to avoid duplicate zero checks and provide slightly higher precision by adding the two portions of rewards before dividing.

x/community/keeper/rewards.go
```diff
@@ -0,0 +1,45 @@
...
+// GetAnnualizedRate returns the annualized staking reward rate
+func (src StakingRewardCalculator) GetAnnualizedRate() sdkmath.LegacyDec {
+	inflationEnabledRate := src.GetInflationRewardRate()
+	inflationDisabledRate := src.GetRPSRewardRate()
+	return inflationEnabledRate.Add(inflationDisabledRate)
+}
+
+// GetRPSRewardRate gets the rewards-per-second contribution of the staking reward rate.
+// Will be zero if rewards per sec is zero.
+func (src StakingRewardCalculator) GetRPSRewardRate() sdkmath.LegacyDec {
```
nddeluca, line 29: Should be private if not used or tested directly.

x/community/keeper/rewards.go
```diff
@@ -0,0 +1,45 @@
...
+	if src.TotalBonded.IsZero() {
+		return sdkmath.LegacyZeroDec()
+	}
+	return src.RewardsPerSecond.Mul(sdkmath.LegacyNewDec(SecondsPerYear)).QuoInt(src.TotalBonded)
+}
+
+// GetInflationRewardRate gets the inflationary contribution of the staking reward rate
+// Will be zero if inflation is zero.
+func (src StakingRewardCalculator) GetInflationRewardRate() sdkmath.LegacyDec {
```
nddeluca, line 38: Should be private if not used or tested directly.

x/community/keeper/rewards.go
```diff
@@ -0,0 +1,45 @@
...
+	sdkmath "cosmossdk.io/math"
+)
+
+const SecondsPerYear = 365 * 24 * 3600
+
+// StakingRewardCalculator wraps data and calculation for determining staking rewards
+// It assumes that staking comes from one of two sources depending on if inflation is enabled or not.
+type StakingRewardCalculator struct {
+	TotalSupply   sdkmath.Int
+	TotalBonded   sdkmath.Int
```
nddeluca, line 13: These data members aren't used outside of internals, so should be private

x/community/keeper/rewards_test.go
```diff
@@ -0,0 +1,117 @@
...
+		},
+		{
+			name:         "inflation only: Oct 2023 case",
+			totalSupply:  sdk.NewInt(857570000),
+			totalBonded:  sdk.NewInt(127680000),
+			inflation:    sdkmath.LegacyMustNewDecFromStr("0.595"),
+			communityTax: sdkmath.LegacyMustNewDecFromStr("0.9495"),
+			perSecReward: sdkmath.LegacyZeroDec(),
+			// expect 20.18% staking reward
+			expectedRate: sdkmath.LegacyMustNewDecFromStr("0.201815746984649124"),
```
nddeluca, line 79: I got `0.201815746984649122` when double checking the math

x/community/keeper/rewards_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+	sdkmath "cosmossdk.io/math"
+	"github.com/stretchr/testify/require"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/kava-labs/kava/x/community/keeper"
+)
+
+func TestStakingRewardsCalculator(t *testing.T) {
```
nddeluca, line 14: Tests are good and covers the edge cases well, but we aren't using any examples that match real world values or larger bounds.

x/community/keeper/rewards_test.go
nddeluca: This is one example where error can be reduced by paying close attention to order we perform calcs in.

nddeluca (changes requested):
Great overall, just some minor comments and improvements that can be made.

nddeluca (approved):

PR #1755, opened 2023-10-24
Update `x/community` lend withdraw proposal to keep funds in `x/community`

drklee3:
## Description
- Remove transfer of funds from `x/community` to `x/distribution` community pool after lend withdraw

## Checklist


nddeluca (approved):

rhuairahrighairidh (approved):
Does `HandleCommunityPoolLendDepositProposal` need to be updated to not use the pool as well? Looks like it will always fail once the pool is empty.

nddeluca:
> Does `HandleCommunityPoolLendDepositProposal` need to be updated to not use the pool as well? Looks like it will always fail once the pool is empty.

We decided to leave it untouched in-case we want to use it before 2024 to deposit funds from distribution and there are not strong reasons to change for the stability committee

PR #1761, opened 2023-10-26
feat: set validator minimum commissions to at least 5% in upgrade handler

drklee3:
## Description
- Iterates over all validators and modifies those with <5% commission, setting them to 5%

## Checklist

### DIFF
--- a/app/upgrades.go
+++ b/app/upgrades.go
@@ -152,6 +152,16 @@ func UpdateValidatorMinimumCommission(
 
 			// Set minimum commission rate to 5%, when commission is < 5%
 			val.Commission.Rate = ValidatorMinimumCommission
+			val.Commission.UpdateTime = ctx.BlockTime()
+
+			// Update MaxRate if necessary
+			if val.Commission.MaxRate.LT(ValidatorMinimumCommission) {
+				val.Commission.MaxRate = ValidatorMinimumCommission
+			}
+
+			if err := app.stakingKeeper.BeforeValidatorModified(ctx, val.GetOperator()); err != nil {
+				panic(fmt.Sprintf("failed to call BeforeValidatorModified: %s", err))
+			}
 			app.stakingKeeper.SetValidator(ctx, val)
 
 			// Keep track of counts just for logging purposes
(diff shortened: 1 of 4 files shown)


app/upgrades.go
```diff
@@ -120,3 +126,61 @@ func upgradeHandler(
...
+				return false
+			}
+
+			val, ok := validator.(stakingtypes.Validator)
+			if !ok {
+				panic("expected stakingtypes.Validator")
+			}
+
+			// Set minimum commission rate to 5%, when commission is < 5%
+			val.Commission.Rate = ValidatorMinimumCommission
```
nddeluca, line 154: We also have to update the Commission.MaxRate if it is less than the new Rate.

In addition, we should set the Commission.UpdateTime to the block time if the rate is changed for a validator

app/upgrades.go
```diff
@@ -120,3 +126,61 @@ func upgradeHandler(
...
+			app.stakingKeeper.SetValidator(ctx, val)
```
nddeluca, line 155: We need to call the staking keeper's `BeforeValidatorModified` hook before we call `SetValidator`

app/upgrades.go
```diff
@@ -120,3 +126,71 @@ func upgradeHandler(
...
+		ctx,
+		func(index int64, validator stakingtypes.ValidatorI) (stop bool) {
+			// Skip if validator commission is already >= 5%
+			if validator.GetCommission().GTE(ValidatorMinimumCommission) {
+				return false
+			}
+
+			val, ok := validator.(stakingtypes.Validator)
+			if !ok {
+				panic("expected stakingtypes.Validator")
```
galxy25, line 150: do we need to panic inline? could be useful to iterate through as many as can be upgraded and collecting a complete list of all invalid ones to make for fewer debug cycles if this unexpected condition is encountered?

galxy25 (approved):

app/upgrades.go
drklee3: Don't think it's really needed, the iterator unmarshals 1 type `types.Validator` before sending it to this fn, so if it panics once in the tests it will be the same for all of them

pirtleshell (approved):
Looks good 👍 

Closes #1573

rhuairahrighairidh (approved):

PR #1762, opened 2023-10-27
Gov & Authz Updates for Kava 15

nddeluca:
- Allows gov to manage incentive module funds through kavadist
- Reduces the gov tally quorum to 20%

### DIFF
--- a/app/upgrades.go
+++ b/app/upgrades.go
@@ -8,13 +8,10 @@ import (
 	storetypes "github.com/cosmos/cosmos-sdk/store/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/types/module"
-<<<<<<< HEAD
-	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
-=======
 	"github.com/cosmos/cosmos-sdk/x/authz"
 	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
 	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
->>>>>>> c3a96997 (add grants that allows gov to manage incentive funds through kavadist;)
+	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
 	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
 
 	communitytypes "github.com/kava-labs/kava/x/community/types"
--- a/app/upgrades.go
+++ b/app/upgrades.go
@@ -8,10 +8,17 @@ import (
 	storetypes "github.com/cosmos/cosmos-sdk/store/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/types/module"
+<<<<<<< HEAD
 	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
+=======
+	"github.com/cosmos/cosmos-sdk/x/authz"
+	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
+	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
+>>>>>>> c3a96997 (add grants that allows gov to manage incentive funds through kavadist;)
 	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
 
 	communitytypes "github.com/kava-labs/kava/x/community/types"
+	kavadisttypes "github.com/kava-labs/kava/x/kavadist/types"
 )
 
 const (
@@ -113,8 +120,14 @@ func upgradeHandler(
 			return toVM, err
 		}
 
+		//
+		// Staking validator minimum commission
+		//
 		UpdateValidatorMinimumCommission(ctx, app)
 
+		//
+		// Community Params
+		//
 		app.communityKeeper.SetParams(ctx, communityParams)
 		app.Logger().Info(
 			"initialized x/community params",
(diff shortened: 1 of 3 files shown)


app/upgrades.go
```diff
@@ -117,6 +124,33 @@ func upgradeHandler(
...
+			return toVM, err
+		}
+		app.Logger().Info("created gov grant for kavadist funds")
+
+		//
+		// Gov Quorum
+		//
+		govTallyParams := app.govKeeper.GetTallyParams(ctx)
+		oldQuorum := govTallyParams.Quorum
+		govTallyParams.Quorum = sdkmath.LegacyMustNewDecFromStr("0.2").String()
```
pirtleshell, line 150: is this value ok for both mainnet & testnet?

tests/e2e/e2e_upgrade_gov_and_authz_test.go
```diff
@@ -0,0 +1,214 @@
...
+	suite.Require().True(ok, "expected generic authorization")
+
+	// kavadist allows gov to MsgSend it's funds
+	suite.Equal(sdk.MsgTypeURL(&banktypes.MsgSend{}), genericAuthorization.Msg)
+	// no expiration
+	var expectedExpiration *time.Time
+	suite.Equal(expectedExpiration, grant.Expiration)
+}
+
+func (suite *IntegrationTestSuite) TestModuleAccountGovTransfers() {
```
pirtleshell, line 81: 🔥 

great test. though it won't pass on mainnet. we might want a way for the test suite to know to skip if running against a place it doesn't have power (or time) to pass a gov proposal.

problem for our future selves, perhaps.

pirtleshell (approved):

tests/e2e/e2e_upgrade_gov_and_authz_test.go
nddeluca: good catch, updated to only run on kvtool

app/upgrades.go
nddeluca: Ya think it's good to keep all env's the same

galxy25 (approved):

tests/e2e/e2e_upgrade_gov_and_authz_test.go
```diff
@@ -0,0 +1,215 @@
...
+			suite.submitAndPassProposal([]sdk.Msg{&execMsg})
+
+			// get ending balance of the receiving account
+			endingBalance := suite.Kava.QuerySdkForBalances(tc.receiver)
+			// subtract the starting balance to get the total transferred amount
+			transferredAmount := endingBalance.AmountOf(tc.amount.Denom).Sub(startingBalance.AmountOf(tc.amount.Denom))
+
+			// assert that the transferred amount is greater than or equal to the sent amount
+			// we use >= here since module accounts can accumulate tokens between blocks
+			suite.Require().True(endingBalance.AmountOf(tc.amount.Denom).GTE(transferredAmount))
```
drklee3, line 135: Should we be comparing the actual change amount to the msg transfer amount instead of endingBalance?

```suggestion
			suite.Require().True(tc.amount.Amount.GTE(transferredAmount))
```

drklee3 (approved):

tests/e2e/e2e_upgrade_gov_and_authz_test.go
pirtleshell: @drklee3 hold off on merge

i ran the tests with
```go
suite.Require().True(transferredAmount.GTE(tc.amount.Amount))
```

ran into this failure:
```
=== RUN   TestIntegrationTestSuite/TestModuleAccountGovTransfers/transfer_from_kavadist_to_community
[whale] 2023/10/31 09:34:36 broadcasting sdk tx. has data = <nil>
[whale] 2023/10/31 09:34:37 response code: 19, hash 9D91BE9F0943C9C78B1028F23E858B3021999A2C83CD9BCC90E006DF2763570E
    e2e_community_update_params_test.go:184: txData.MsgResponses: [&Any{TypeUrl:/cosmos.gov.v1.MsgSubmitProposalResponse,Value:[8 3],XXX_unrecognized:[]}]
[whale] 2023/10/31 09:34:37 broadcasting sdk tx. has data = <nil>
[whale] 2023/10/31 09:34:38 response code: 0, hash DBFBC1DB89C3310EFDA64FB437BF32CC1B703281FFC97BB7BFE3B3E0B262E252
    e2e_upgrade_gov_and_authz_test.go:135: 
                Error Trace:    /kava/kava/tests/e2e/e2e_upgrade_gov_and_authz_test.go:135
                                /go/pkg/mod/github.com/stretchr/testify@v1.8.3/suite/suite.go:112
                Error:          Should be true
```

am looking into it.

pirtleshell: with additional logging:
```
transferred amount: 49963854
amount: 50000000
```

drklee3: Updating the test to account for this, does seem to be around 35 seconds worth of staking rewards which makes sense with the gov voting period being 30s

tests/e2e/e2e_upgrade_gov_and_authz_test.go
```diff
@@ -0,0 +1,276 @@
...
+			// ensure proposal passes
+			passBlock := suite.submitAndPassProposal([]sdk.Msg{&execMsg})
+
+			// get starting balance of the module account to be funded, 1 block
+			// before the proposal passes
+			startingBalance := suite.Kava.QuerySdkForBalancesAtHeight(passBlock-1, tc.receiver)
+			// get ending balance of the receiving account
+			endingBalance := suite.Kava.QuerySdkForBalancesAtHeight(passBlock, tc.receiver)
+
+			paidRewards := suite.getStakingRewardsAtBlock(passBlock)
```
rhuairahrighairidh, line 133: When the receiver is kavadist, should paid rewards be included?

I'm thinking if we can't control the exact amount into or out of these accounts, we should parse the logs to check for a transfer event of the correct amount. Then maybe also check the difference in balance is accounted for by all the transfer events.

ok to merge this if blocking mirrornet, and address in a followup pr

PR #1767, opened 2023-11-07
Correct rocksdb path -- ensure we use KAVA_HOME/data/application.db

nddeluca:


drklee3 (approved):

cmd/kava/opendb/opendb_rocksdb.go
```diff
@@ -90,7 +90,8 @@ func OpenDB(appOpts types.AppOptions, home string, backendType dbm.BackendType)
 // openRocksdb loads existing options, overrides some of them with appOpts and opens database
 // option will be overridden only in case if it explicitly specified in appOpts
 func openRocksdb(dir string, appOpts types.AppOptions) (dbm.DB, error) {
-	dbOpts, cfOpts, err := loadLatestOptions(dir)
+	optionsPath := filepath.Join(dir, "application.db")
```
galxy25, line 93: should we move this to being sourced from config.yml?

nddeluca: ya agree, we can implement a better fix on master w/ an integration test to ensure path doesn't change in the future

PR #1804, opened 2024-01-04
feat(cli): Add `rocksdb compact` command

drklee3:
## Description
- Manual compaction command only enabled with rocksdb

## Checklist

### DIFF
--- a/cmd/kava/cmd/rocksdb/compact.go
+++ b/cmd/kava/cmd/rocksdb/compact.go
@@ -10,6 +10,7 @@ import (
 	"os/signal"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"syscall"
 	"time"
 
@@ -18,12 +19,18 @@ import (
 	"github.com/kava-labs/kava/cmd/kava/opendb"
 	"github.com/linxGnu/grocksdb"
 	"github.com/spf13/cobra"
+	"golang.org/x/exp/slices"
 
 	"github.com/tendermint/tendermint/libs/log"
 )
 
+var allowedDBs = []string{"application", "blockstore", "state"}
+
 var CompactRocksDBCmd = &cobra.Command{
-	Use:   "compact <state|blockstore>",
+	Use: fmt.Sprintf(
+		"compact <%s>",
+		strings.Join(allowedDBs, "|"),
+	),
 	Short: "force compacts RocksDB",
 	Long: `This is a utility command that performs a force compaction on the state or
 blockstore. This should only be run once the node has stopped.`,
@@ -38,8 +45,11 @@ blockstore. This should only be run once the node has stopped.`,
 			return errors.New("compaction is currently only supported with rocksdb")
 		}
 
-		if args[0] != "state" && args[0] != "blockstore" {
-			return errors.New("invalid db name, must be either 'state' or 'blockstore'")
+		if !slices.Contains(allowedDBs, args[0]) {
+			return fmt.Errorf(
+				"invalid db name, must be one of the following: %s",
+				strings.Join(allowedDBs, ", "),
+			)
 		}
 
 		compactRocksDBs(clientCtx.HomeDir, logger, args[0])
--- a/cmd/kava/cmd/rocksdb/compact.go
+++ b/cmd/kava/cmd/rocksdb/compact.go
@@ -84,6 +84,7 @@ func bytesToMB(bytes uint64) float64 {
 	return float64(bytes) / 1024 / 1024
 }
 
+// logColumnFamilyMetadata outputs the column family and level metadata.
 func logColumnFamilyMetadata(
 	store *grocksdb.DB,
 	logger log.Logger,
--- a/cmd/kava/cmd/rocksdb/compact.go
+++ b/cmd/kava/cmd/rocksdb/compact.go
@@ -1,0 +1,176 @@
+//go:build rocksdb
+// +build rocksdb
+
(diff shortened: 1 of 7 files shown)


cmd/kava/cmd/rocksdb/compact.go
```diff
@@ -0,0 +1,170 @@
...
+	rootDir string,
+	logger log.Logger,
+	dbName string,
+) {
+	options := grocksdb.NewDefaultOptions()
+
+	dbPath := filepath.Join(rootDir, "data", dbName+".db")
+	logger.Info("opening db", "path", dbPath)
+
+	store, err := grocksdb.OpenDb(options, dbPath)
```
nddeluca, line 60: We'll have to extend Yevhenii's work at https://github.com/Kava-Labs/kava/blob/master/cmd/kava/opendb/opendb_rocksdb.go#L92 to separate returning of the db instance from the dbm wrapper.

Otherwise, we'll end up opening the db with possibly incompatible options and/or the compacted files will be re-written with different block sizes, etc than app.toml specifies

cmd/kava/cmd/rocksdb/compact.go
```diff
@@ -0,0 +1,170 @@
...
+	logColumnFamilyMetadata(store, logger)
+
+	logger.Info("starting compaction...", "db", dbPath)
+
+	done := make(chan bool, 1)
+	registerSignalHandler(store, logger, done)
+	startCompactionStatsOutput(store, logger, done)
+
+	// Actually run the compaction
+	store.CompactRange(grocksdb.Range{Start: nil, Limit: nil})
```
nddeluca, line 76: Just a nit, but lets name this `db` or something similar.  For me `store` always makes me thinkn of the cosmos-sdk store, not a raw rocksdb struct

cmd/kava/cmd/rocksdb/compact.go
nddeluca: Edit: I see now this is for the state & blockstore -- those use default options from tmdb -- I"m not sure if the default grocksdb options match. The app.toml settings will only apply applicationdb.

On further thought, maybe we should just load the latest options from the DB and ignore defaults or reading app.toml.

Grocksdb can do that like so - https://github.com/Kava-Labs/kava/blob/master/cmd/kava/opendb/opendb_rocksdb.go#L118.  So we always open & compact with the options the db was last opened with

cmd/kava/cmd/rocksdb/compact.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/server"
+	"github.com/linxGnu/grocksdb"
+	"github.com/spf13/cobra"
+
+	"github.com/tendermint/tendermint/libs/log"
+)
+
+var CompactRocksDBCmd = &cobra.Command{
+	Use:   "compact <state|blockstore>",
```
nddeluca, line 25: These are good db's, but the main one we want to target for compaction is the application.db

nddeluca (changes requested):
Did a quick review

cmd/kava/cmd/rocksdb/compact.go
drklee3: oh great, useful function, will update options used

cmd/kava/cmd/rocksdb/compact.go
```diff
@@ -0,0 +1,194 @@
...
+		logger.Error("failed to initialize cometbft db", "path", dbPath, "err", err)
+		return fmt.Errorf("failed to open db %s %w", dbPath, err)
+	}
+	defer db.Close()
+
+	logColumnFamilyMetadata(db, logger)
+
+	logger.Info("starting compaction...", "db", dbPath)
+
+	done := make(chan bool, 1)
```
nddeluca, line 94: Nice signal handling, however there are a couple subtle bugs.

Done may be sent but not selected by the compaction stats routine, or done may be sent after the ticker is selected but before GetProperty is called.  Since t`done` a buffered channel, the `done <- true` line will not block and the deferred close may run and leave enough time for GetProperty to called, still allowing use after close.

In addition, os.Exit() in the signal/interrupt handling will stop the program immediately and the defer in the command will not be run (https://pkg.go.dev/os#Exit).

The simplest fix is chaging `done` to a unbuffered channel and remove the os.Exit call and send on the done channel in the signal handler routine.  If I read the docs right, the call to disable manual compaction will cause CompactRange to return, resulting in done being called, stopping the stats routine and allowing a normal exit.

This separates concerns a bit - The signal handlers only job is to catch and cancel the blocking compaction.  The done channels job is to stop and wait for the stats routine to exit before continuing to the deferred close.

Note: If the stats routine needed cleanup a done channel wouldn't be enough since this doesn't guarantee the routine will return before exit, so if we needed that guarantee we'd have to use a context/cancel function with a wait group or some other pattern.

nddeluca (changes requested):
DB and compaction logic is great 👍  Just some small touches needed to the signal handling and should be good to merge

pirtleshell (approved):
looks great! worth considering the lift to upstream to cometbft's [`compact` command](https://github.com/Kava-Labs/cometbft/blob/main/cmd/cometbft/commands/compact.go)

approved pending resolution of early exit due to buffered channel @nddeluca mentions

galxy25 (approved):

PR #1806, opened 2024-01-05
feat(statedb): Add initial module setup with base StateDB

drklee3:
## Description
Adds:
- New statedb go module, under a new `eth` directory - if we also want to add custom precompiles under there
- License
- Basic README
- Basic Makefile just with test for now
- 1:1 copy of pre-existing StateDB from Ethermint `v0.21.0-kava-v23-1`

## Checklist


eth/statedb/statedb_test.go
```diff
@@ -0,0 +1,579 @@
...
+import (
+	"math/big"
+	"testing"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/ethereum/go-ethereum/common"
+	ethtypes "github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/core/vm"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/evmos/ethermint/x/evm/statedb"
```
nddeluca, line 12: Should be `github.com/kava-labs/kava/eth/statedb`

eth/statedb/go.mod
```diff
@@ -0,0 +1,151 @@
+module github.com/kava-labs/kava/eth/statedb
+
+go 1.20
+
+require (
+	cosmossdk.io/errors v1.0.0-beta.7
+	github.com/cosmos/cosmos-sdk v0.46.11
+	github.com/ethereum/go-ethereum v1.10.26
+	github.com/evmos/ethermint v0.21.0
```
nddeluca, line 9: Shouldn't have a dependency on etheremint

eth/statedb/config.go
```diff
@@ -0,0 +1,62 @@
+// Copyright 2021 Evmos Foundation
```
nddeluca, line 1: Let's create an interface for the TxConfig in the StateDB struct, then remove this implementation and keep it in etheremint

.github/workflows/ci-statedb.yml
drklee3, line 1: Standalone workflow for testing statedb since I don't think it needs to be run when non-statedb code is modified

galxy25: Agreed! Nice optimization

eth/statedb/statedb.go
```diff
@@ -0,0 +1,478 @@
...
+	"github.com/ethereum/go-ethereum/core/vm"
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+// revision is the identifier of a version of state.
+// it consists of an auto-increment id and a journal index.
+// it's safer to use than using journal index alone.
+type revision struct {
+	id           int
+	journalIndex int
```
galxy25, line 36: int64 for more headroom (unless the indices are reset every block then unlikely we run out of them)?

eth/statedb/statedb.go
```diff
@@ -0,0 +1,478 @@
...
+// * Accounts
+type StateDB struct {
+	keeper Keeper
+	ctx    sdk.Context
+
+	// Journal of state modifications. This is the backbone of
+	// Snapshot and RevertToSnapshot.
+	journal        *journal
+	validRevisions []revision
+	nextRevisionID int
```
galxy25, line 54: int64 for more headroom (unless the revisions are reset every block then unlikely we run out of them)?

eth/statedb/statedb.go
```diff
@@ -0,0 +1,478 @@
...
+
+	stateObjects map[common.Address]*stateObject
+
+	txConfig TxConfig
+
+	// The refund counter, also used by state transitioning.
+	refund uint64
+
+	// Per-transaction logs
+	logs []*ethtypes.Log
```
galxy25, line 64: depending how often these are referenced, suggest using a map with the tx hash as the key (and the slice of logs for that tx as the value ofc)

eth/statedb/statedb.go
```diff
@@ -0,0 +1,478 @@
...
+		return stateObject.GetState(hash)
+	}
+	return common.Hash{}
+}
+
+// GetCommittedState retrieves a value from the given account's committed storage trie.
+func (s *StateDB) GetCommittedState(addr common.Address, hash common.Hash) common.Hash {
+	stateObject := s.getStateObject(addr)
+	if stateObject != nil {
+		return stateObject.GetCommittedState(hash)
```
galxy25, line 192: all of these `Getters` would be more idiomatic golang if they returned an error for when the value wasn't in state (potentially even a typed error to allow callers to decide whether they are okat with the default value for a non-existent key

eth/statedb/statedb.go
```diff
@@ -0,0 +1,478 @@
...
+//
+// 1. sends funds to sha(account ++ (nonce + 1))
+// 2. tx_create(sha(account ++ nonce)) (note that this gets the address of 1)
+//
+// Carrying over the balance ensures that Ether doesn't disappear.
+func (s *StateDB) CreateAccount(addr common.Address) {
+	newObj, prev := s.createObject(addr)
+	if prev != nil {
+		newObj.setBalance(prev.account.Balance)
+	}
```
galxy25, line 276: wait where is `newObj` returned or is this a method that is called only for it's side effects?

eth/statedb/statedb.go
drklee3: StateDB needs to implement the StateDB interface from geth so there isn't much flexibility on return values https://github.com/ethereum/go-ethereum/blob/v1.10.26/core/vm/interface.go

galxy25 (approved):

PR #1809, opened 2024-01-11
Initial implementation of Stateful Precompile Contracts

evgeniy-scherbina:
Moved to https://github.com/Kava-Labs/kava/pull/1826


Dockerfile
```diff
@@ -14,6 +14,10 @@ WORKDIR /root/kava
 COPY ./go.mod ./
 COPY ./go.sum ./
 
+RUN echo 3
+RUN mkdir -p /Users/yevheniishcherbina/go/src/github.com/Kava-Labs
```
galxy25, line 18: Recommend making this a variable vs hard-coded, for example

https://stackoverflow.com/questions/40902445/using-variable-interpolation-in-string-in-docker

evgeniy-scherbina: It will be removed, it's temporary

docker-build.sh
```diff
@@ -0,0 +1,4 @@
+rm -rf ./docker-deps && mkdir -p ./docker-deps
+cp -r /Users/yevheniishcherbina/go/src/github.com/Kava-Labs/ethermint docker-deps/
```
galxy25, line 2: Recommend making this a variable vs hard-coded, for example

https://stackoverflow.com/questions/40902445/using-variable-interpolation-in-string-in-docker

galxy25 (approved):

docker-build.sh
evgeniy-scherbina: It will be removed, it's temporary

PR #1811, opened 2024-01-17
Update cosmos-sdk to v0.47.7

DracoLi:
## Description
See `Kava Cosmos v47.7 Upgrade Changes` doc.

## Checklist


galxy25 (approved):
all tests pass and app builds with rocksdb, approved for protonet deploy with async code and security audit ongoing as part of validation process before public testnet launch

app/app.go
```diff
@@ -193,6 +196,7 @@ var (
 		crisis.AppModuleBasic{},
 		slashing.AppModuleBasic{},
 		ibc.AppModuleBasic{},
+		ibctm.AppModuleBasic{},
```
nddeluca, line 199: https://github.com/cosmos/ibc-go/pull/2825/files, this just registers the codec's as it's not in ibc core anymore

app/app_test.go
```diff
@@ -143,3 +146,13 @@ func unmarshalJSONKeys(jsonBytes []byte) ([]string, error) {
 
 	return keys, nil
 }
+
+func removeIbcTmModule(modules []string) []string {
+	var result []string
+	for _, str := range modules {
+		if str != "07-tendermint" {
```
nddeluca, line 153: We can use `"github.com/cosmos/ibc-go/v7/modules/light-clients/07-tendermint".ModuleName` here

cmd/kava/cmd/app.go
```diff
@@ -62,7 +64,7 @@ func (ac appCreator) newApp(
 
 	homeDir := cast.ToString(appOpts.Get(flags.FlagHome))
 	snapshotDir := filepath.Join(homeDir, "data", "snapshots") // TODO can these directory names be imported from somewhere?
-	snapshotDB, err := sdk.NewLevelDB("metadata", snapshotDir)
+	snapshotDB, err := cometbftdb.NewGoLevelDB("metadata", snapshotDir)
```
nddeluca, line 67: This is a regression here -- should be `cometbftdb.NewDB("metadata", server.GetAppDBBackedn(appOpts), snapshotDir)` to open using configured db type

cmd/kava/cmd/app.go
```diff
@@ -89,6 +91,24 @@ func (ac appCreator) newApp(
 		cast.ToUint32(appOpts.Get(server.FlagStateSyncSnapshotKeepRecent)),
 	)
 
+	// Setup chainId
+	home := cast.ToString(appOpts.Get(flags.FlagHome))
+	chainID := cast.ToString(appOpts.Get(flags.FlagChainID))
+	if len(chainID) == 0 {
```
nddeluca, line 97: I believe this should fallback to genesis.json and not the client.toml

go.mod
```diff
@@ -1,36 +1,38 @@
...
+	github.com/cosmos/cosmos-sdk v0.47.7
 	github.com/cosmos/go-bip39 v1.0.0
-	github.com/cosmos/ibc-go/v6 v6.1.1
+	github.com/cosmos/gogoproto v1.4.10
+	github.com/cosmos/ibc-go/v7 v7.3.1
 	github.com/ethereum/go-ethereum v1.10.26
 	github.com/evmos/ethermint v0.21.0
 	github.com/go-kit/kit v0.12.0
-	github.com/gogo/protobuf v1.3.3
+	github.com/gogo/protobuf v1.3.2
```
nddeluca, line 20: Downgrade?

x/validator-vesting/querier.go
```diff
@@ -1,358 +0,0 @@
-package validator_vesting
```
nddeluca, line 1: Just noting that I believe coinmarketcap does use these.  I'm not sure a grpc querier would let us return an int value as the response

nddeluca (approved):
Looks good.  I think in a follow on PR we should switch all of our Dec usage to sdkmath instead of the legacy types, but not required for this PR

Zgeordo (approved):

PR #1812, opened 2024-01-18
build: inject brew deps for MacOS rocksdb build

pirtleshell:
## Description
Handles linking necessary dependencies installed via brew when building with rocksdb on MacOS.

Not sure if having installed all these through brew is what we want to enforce on mac. This will cause build failures if you've installed any of the dependencies from source (the linker will complain that the corresponding `/opt/homebrew/opt/*` directory does not exist).

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -162,6 +162,13 @@ ifeq (rocksdb,$(findstring rocksdb,$(COSMOS_BUILD_OPTIONS)))
   CGO_ENABLED=1
   BUILD_TAGS += rocksdb
   ldflags += -X github.com/cosmos/cosmos-sdk/types.DBBackend=rocksdb
+	# inject brew dependencies on mac
+	# this assumes that if you are building on mac, you've installed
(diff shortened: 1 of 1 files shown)


galxy25 (approved):

Makefile
```diff
@@ -162,6 +162,13 @@ ifeq (rocksdb,$(findstring rocksdb,$(COSMOS_BUILD_OPTIONS)))
   CGO_ENABLED=1
   BUILD_TAGS += rocksdb
   ldflags += -X github.com/cosmos/cosmos-sdk/types.DBBackend=rocksdb
+	# inject brew dependencies on mac
+	# this assumes that if you are building on mac, you've installed
+	# rocksdb, snappy, lz4, and zstd via brew
+	ifeq ($(NATIVE_GO_OS),darwin)
```
nddeluca, line 168: Maybe we should use a flag? This may break compiling against rocksdb build from source outside of brew?

PR #1813, opened 2024-01-19
cli: add shard command

pirtleshell:
## Description
Adds the shard command to iavl v1 branch of v0.25
Updates command as needed to comply with upgraded dependency function signatures.

## Checklist

### DIFF
--- a/cmd/kava/cmd/shard.go
+++ b/cmd/kava/cmd/shard.go
@@ -1,0 +1,232 @@
+package cmd
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	dbm "github.com/cometbft/cometbft-db"
+
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/flags"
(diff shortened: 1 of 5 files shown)


cmd/kava/cmd/shard.go
```diff
@@ -0,0 +1,225 @@
...
+			printRollbackProgress := func(h int64) {
+				fmt.Print(clearLine)
+				fmt.Printf(progress, h)
+			}
+
+			// rollback tendermint db
+			height := latest
+			for height > endBlock {
+				printRollbackProgress(height - 1)
+				height, _, err = tmstate.Rollback(blockStore, stateStore, true)
```
drklee3, line 147: Can infinite loop here if Rollback() returns the same height - https://github.com/Kava-Labs/cometbft/blob/904878c5022c4f79f70506fc0fed7de2a322b8e5/state/rollback.go#L29-L36

Could use an additional check if the returned height is unchanged

pirtleshell: 🙌 great catch!
i've updated the original command in #1837 

am in the process of re-backporting the command to the iavl v1 branches

PR #1818, opened 2024-01-26
feat(x/cdp): update begin bloocker to run every X blocks based off params

galxy25:
## Description

* add new cdp module param to protonet genesis with unit test for validation logic
* update cdp / cdp related tests for new module param
* update telemetry docs and setup for collecting against local node
* update kvool commit for new cdp param
* add migration for new param with migration test

## Checklist
 - [x] `x/cdp` `abci_test.go` test that begin blocker runs at configured interval


x/cdp/module.go
```diff
@@ -26,6 +27,9 @@ var (
 	// _ module.AppModuleSimulation = AppModule{}
 )
 
+// ConsensusVersion defines the current module consensus version.
+const ConsensusVersion = 2
```
galxy25, line 31: @nddeluca @pirtleshell should this be 1 since it wasn't defined before or does the first version of the cdp module store point to the current version (pre kava-16)

galxy25: updated

x/cdp/abci.go
```diff
@@ -20,6 +20,16 @@ func BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock, k keeper.Keeper)
 
 	params := k.GetParams(ctx)
 
+	// only run CDP Begin Blocker every `BeginBlockerExecutionBlockInterval` blocks
+	blockHeight := ctx.BlockHeight()
+
+	if blockHeight%params.BeginBlockerExecutionBlockInterval != 0 {
+		ctx.Logger().Info("skipping x/cdp begin blocker")
```
nddeluca, line 27: We don't need info logs for this -- let's switch to Debug or remove.  Validators & exchanges already think kava/tendermint/cosmos has too many logs

x/cdp/migrations/v2/store.go
```diff
@@ -0,0 +1,22 @@
...
+	migrateParamsStore(ctx, paramstore)
+	return nil
+}
+
+// migrateParamsStore ensures the param key table exists and has the begin_blocker_execution_block_interval property
+func migrateParamsStore(ctx sdk.Context, paramstore paramtypes.Subspace) {
+	if !paramstore.HasKeyTable() {
+		paramstore.WithKeyTable(types.ParamKeyTable())
+	}
+	paramstore.Set(ctx, types.KeyBeginBlockerExecutionBlockInterval, types.DefaultBeginBlockerExecutionBlockInterval)
```
nddeluca, line 21: Nice solution instead of loading and settings all params again like we have done in others 🚀 .  I wrote a test locally to confirm writing the entire legacy paramset with a legacy key table, then reading with the new paramset and new key table works as expected.

x/cdp/migrations/v2/store_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+	// Check param doesn't exist before
+	require.False(t, paramstore.Has(ctx, types.KeyBeginBlockerExecutionBlockInterval))
+
+	// Run migrations.
+	err := v2cdp.MigrateStore(ctx, paramstore)
+	require.NoError(t, err)
+
+	// Make sure the new params are set.
+	require.True(t, paramstore.Has(ctx, types.KeyBeginBlockerExecutionBlockInterval))
```
nddeluca, line 32: Let's also add an assertion that the value is what we expect.  The migrator is also ensuring the default value is set but we don't have a test asserting it does -- only that it sets the value to something.

x/cdp/types/params_test.go
```diff
@@ -783,24 +805,61 @@ func (suite *ParamsTestSuite) TestParamValidation() {
...
+				debtLot:                            sdk.ZeroInt(),
+				breaker:                            types.DefaultCircuitBreaker,
+				beginBlockerExecutionBlockInterval: types.DefaultBeginBlockerExecutionBlockInterval,
 			},
 			errArgs: errArgs{
 				expectPass: false,
 				contains:   "debt auction lot should be positive",
 			},
 		},
+		{
```
nddeluca, line 823: Nice, very helpful to have these when refactoring params.  Easy to break validations

x/cdp/abci.go
```diff
@@ -20,6 +20,16 @@ func BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock, k keeper.Keeper)
...
+		return
+	}
+
+	ctx.Logger().Debug("running x/cdp begin blocker")
```
nddeluca, line 31: I'm not sure how useful this is, but fine to leave.  But could be biased as I tend to lean towards no logs 😓

x/cdp/abci_test.go
```diff
@@ -174,6 +221,74 @@ func (suite *ModuleTestSuite) TestSeizeSingleCdpWithFees() {
...
+	acc = ak.GetModuleAccount(suite.ctx, types.ModuleName)
+	// get the current amount of xrp held by the cdp module
+	finalXrpCollateral := bk.GetBalance(suite.ctx, acc.GetAddress(), "xrp").Amount
+	seizedXrpCollateral := originalXrpCollateral.Sub(finalXrpCollateral)
+	// calculate the number of cdps that were liquidated based on the total
+	// seized collateral divided by the size of each cdp when it was created
+	xrpLiquidations := int(seizedXrpCollateral.Quo(i(10000000000)).Int64())
+	// should be 0 because the cdp begin blocker is configured to
+	// skip execution every odd numbered block
+	suite.Equal(0, xrpLiquidations)
```
nddeluca, line 271: Let's add a message here for the test failure, it's not clear from the test output otherwise.  Something like
`suite.Equal(0, xrpLiquidations, "expected cdp begin blocker not to run liqudations")`

nddeluca: We can also add a message with a length check to make it clear we are looking for migrations to not run

x/cdp/abci_test.go
```diff
@@ -174,6 +221,74 @@ func (suite *ModuleTestSuite) TestSeizeSingleCdpWithFees() {
...
+	// test case 2 assert
+	acc = ak.GetModuleAccount(suite.ctx, types.ModuleName)
+	// get the current amount of xrp held by the cdp module
+	finalXrpCollateral = bk.GetBalance(suite.ctx, acc.GetAddress(), "xrp").Amount
+	seizedXrpCollateral = originalXrpCollateral.Sub(finalXrpCollateral)
+	// calculate the number of cdps that were liquidated based on the total
+	// seized collateral divided by the size of each cdp when it was created
+	xrpLiquidations = int(seizedXrpCollateral.Quo(i(10000000000)).Int64())
+	// should be 10 because of magic numbers elsewhere in this file
+	suite.Equal(10, xrpLiquidations)
```
nddeluca, line 289: ditto on message here since we are focused on running vs not running and not the number in the test.  Could assert length greater than zero here as well to remove dependence on numbers since that's not the goal of the test

nddeluca (changes requested):
Looks great, just a couple of minor improvements needed imo.

x/cdp/abci.go
```diff
@@ -20,6 +20,16 @@ func BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock, k keeper.Keeper)
 
 	params := k.GetParams(ctx)
 
+	// only run CDP Begin Blocker every `BeginBlockerExecutionBlockInterval` blocks
+	blockHeight := ctx.BlockHeight()
+
+	if blockHeight%params.BeginBlockerExecutionBlockInterval != 0 {
```
nddeluca, line 26: On further thought here -- I believe we should only skip Synchronize and Liquidate @galxy25 @rhuairahrighairidh.  It's important to keep the pricefeed market status and global interest factor up to date for transactions, keeper liquidations, etc

galxy25: agreed, updated

x/cdp/abci.go
```diff
@@ -36,6 +39,14 @@ func BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock, k keeper.Keeper)
 			panic(err)
 		}
 
+		if blockHeight%params.BeginBlockerExecutionBlockInterval != 0 {
+			ctx.Logger().Debug("skipping x/cdp SynchronizeInterestForRiskyCDPs and LiquidateCdps")
```
nddeluca, line 43: Could throw `cp.Type` in these messages since they will be repeated for each type

x/cdp/abci.go
```diff
@@ -36,6 +39,14 @@ func BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock, k keeper.Keeper)
...
+			return
+		}
+
+		err = k.LiquidateCdps(ctx, cp.LiquidationMarketID, cp.Type, cp.LiquidationRatio, cp.CheckCollateralizationIndexCount)
```
nddeluca, line 47: Looks like a copypasta mistake -- this LiquidateCdps shouldn't be added as it is run below

x/cdp/abci.go
```diff
@@ -36,6 +39,14 @@ func BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock, k keeper.Keeper)
 			panic(err)
 		}
 
+		if blockHeight%params.BeginBlockerExecutionBlockInterval != 0 {
```
nddeluca, line 42: This is probably a good case for a named parameter now since it's duplicated in a loop and we are already storing the block height in a variabled

```go
skipSyncronizeAndLiquidations := ctx.BlockHeight()%params.BeginBlockerExecutionBlockInterval != 0

...

  if skipskipSyncronizeAndLiquidations {
  ...
  }
```

x/cdp/types/params.go
```diff
@@ -39,22 +40,25 @@ var (
...
-		CircuitBreaker:          breaker,
+		GlobalDebtLimit:                    debtLimit,
+		CollateralParams:                   collateralParams,
+		DebtParam:                          debtParam,
+		SurplusAuctionThreshold:            surplusThreshold,
+		SurplusAuctionLot:                  surplusLot,
+		DebtAuctionThreshold:               debtThreshold,
+		DebtAuctionLot:                     debtLot,
+		CircuitBreaker:                     breaker,
+		BeginBlockerExecutionBlockInterval: beginBlockerExecutionBlockInterval,
```
nddeluca, line 61: Maybe we should call this `LiquidationBlockInterval` now?  We aren't really skipping the begin blocker, just liquidations & those updates of risky cdps for the purpose of liquidation

x/cdp/abci.go
galxy25: good idea, updated

x/cdp/abci.go
```diff
@@ -20,6 +21,9 @@ func BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock, k keeper.Keeper)
 
 	params := k.GetParams(ctx)
 
+	// only run CDP Begin Blocker every `BeginBlockerExecutionBlockInterval` blocks
```
nddeluca, line 24: nit: comment doesn't match after rename

x/cdp/abci.go
```diff
@@ -36,6 +40,13 @@ func BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock, k keeper.Keeper)
 			panic(err)
 		}
 
+		if skipSyncronizeAndLiquidations {
+			ctx.Logger().Debug(fmt.Sprintf("skipping x/cdp SynchronizeInterestForRiskyCDPs and LiquidateCdps for %s", cp.Type))
+			return
+		}
+
+		ctx.Logger().Debug(fmt.Sprintf("running x/cdp SynchronizeInterestForRiskyCDPs and LiquidateCdpsfor %s", cp.Type))
```
nddeluca, line 48: nit: space at LiquidateCdpsfor

nddeluca (approved):

x/cdp/abci.go
```diff
@@ -36,6 +40,13 @@ func BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock, k keeper.Keeper)
 			panic(err)
 		}
 
+		if skipSyncronizeAndLiquidations {
+			ctx.Logger().Debug(fmt.Sprintf("skipping x/cdp SynchronizeInterestForRiskyCDPs and LiquidateCdps for %s", cp.Type))
+			return
```
rhuairahrighairidh, line 45: should this be `continue` instead of `return`? Currently it'll update the first `CollateralParam` every block but return before it updates the others.

rhuairahrighairidh: If changed to `continue` it'll run `RunSurplusAndDebtAuctions` every block. If you want to optimise further I believe this function doesn't need to be run every block.

galxy25: agreed, updated to `continue` as `RunSurplusAndDebtAuctions` doesn't incur a significant performance penalty but we can always adjust in the future if that changes

PR #1822, opened 2024-02-08
Optimize CDP Begin Blocker

nddeluca:
Shows a ~10x improvement on mainnet and tested on blocks 1 to ~4.8 million

### DIFF
--- a/x/cdp/keeper/interest.go
+++ b/x/cdp/keeper/interest.go
@@ -181,9 +181,8 @@ func (k Keeper) SynchronizeInterestForRiskyCDPs(ctx sdk.Context, targetRatio sdk
 	iterator.Close()
 
 	globalInterestFactor, found := k.GetInterestFactor(ctx, cp.Type)
-	if !found {
-		globalInterestFactor = sdk.OneDec()
-		k.SetInterestFactor(ctx, cp.Type, globalInterestFactor)
+	if !found && len(cdpIDs) > 0 {
+		panic(fmt.Sprintf("global interest factor not found for type %s", cp.Type))
 	}
 	prevAccrualTime, found := k.GetPreviousAccrualTime(ctx, cp.Type)
 	if !found {
--- a/x/cdp/keeper/interest.go
+++ b/x/cdp/keeper/interest.go
@@ -182,7 +182,8 @@ func (k Keeper) SynchronizeInterestForRiskyCDPs(ctx sdk.Context, targetRatio sdk
 
 	globalInterestFactor, found := k.GetInterestFactor(ctx, cp.Type)
 	if !found {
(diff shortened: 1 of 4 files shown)


x/cdp/keeper/interest.go
```diff
@@ -161,11 +162,110 @@ func (k Keeper) CalculateNewInterest(ctx sdk.Context, cdp types.CDP) sdk.Coin {
...
+	for ; iterator.Valid(); iterator.Next() {
+		_, id, _ := types.SplitCollateralRatioKey(iterator.Key())
+		cdpIDs = append(cdpIDs, id)
+		if int64(len(cdpIDs)) >= cp.CheckCollateralizationIndexCount.Int64() {
+			break
+		}
+	}
+	iterator.Close()
+
+	globalInterestFactor, found := k.GetInterestFactor(ctx, cp.Type)
```
nddeluca, line 183: The previous version created this if it didn't exist, but the accumulate interest functions called before this method already performed that logic

nddeluca: Looking into why this affected the eip712 test

nddeluca: Resolved -- forgot to address the case when both previous accural & global interest factor are not set, but this is added as an abci test now

galxy25 (approved):
nice optimizations, I knew there was some low hanging fruit here to pluck with a little bit of profiling and tweaking - https://kava-labs.atlassian.net/wiki/spaces/ENG/pages/1362001933/First+Round+Performance+Optimizations+Design+Doc#Option-1 #TODONE

x/cdp/keeper/interest.go
```diff
@@ -161,11 +162,111 @@ func (k Keeper) CalculateNewInterest(ctx sdk.Context, cdp types.CDP) sdk.Coin {
...
+	if !found {
+		globalInterestFactor = sdk.OneDec()
+		k.SetInterestFactor(ctx, cp.Type, globalInterestFactor)
```
nddeluca, line 186: @rhuairahrighairidh returning here might be better? Though, I don't believe there is a case where a cdp can exist without a global interest factor set (create cdp sets it), so regardless believe it to be compatible with old behavior

rhuairahrighairidh: yeah slight lean to just returning. I don't think it matters but not having interest sync (from a early return) feels not as bad as potentially incorrectly setting a global factor

rhuairahrighairidh (approved):

nddeluca:
Moving to draft for some touch ups and to align better with current functionality

pirtleshell (approved):

galxy25 (approved):

nddeluca:
Merging to master, but not back-porting to v0.26.x until after release

PR #1826, opened 2024-02-09
Feat/precompile

evgeniy-scherbina:


precompile/contracts/ibc/contract.go
```diff
@@ -0,0 +1,145 @@
...
+		return nil, remainingGas, vmerrs.ErrWriteProtection
+	}
+
+	// Set the nonce of the precompile's address (as is done when a contract is created) to ensure
+	// that it is marked as non-empty and will not be cleaned up when the statedb is finalized.
+	accessibleState.GetStateDB().SetNonce(ContractAddress, 1)
+	// Set the code of the precompile's address to a non-zero length byte slice to ensure that the precompile
+	// can be called from within Solidity contracts. Solidity adds a check before invoking a contract to ensure
+	// that it does not attempt to invoke a non-existent contract.
+	accessibleState.GetStateDB().SetCode(ContractAddress, []byte{0x1})
```
drklee3, line 59: Should we move this to the geth [`contract.Run`](https://github.com/Kava-Labs/go-ethereum/blob/935f656757169f24b08b9e79a6371c1202a1fa8e/precompile/contract/contract.go#L70-L94) parent method?

Since it seems it would be shared logic for all stateful precompiles and can be covered by the parent test instead of for each individual precompile.

PR #1830, opened 2024-02-15
Register solomachine module for ibc-go v6 to v7 migration

galxy25:
## Checklist


DracoLi (approved):
looks good! we should merge this to master as well

PR #1834, opened 2024-02-23
Spike: Support decimal conversion in internal bridge

rhuairahrighairidh:
## Description

This allows assets connected in the internal bridge to support different decimal precision on each side.
eg 8 decimal places on the sdk side and 18 on the evm

When converting down the way (eg an 18 decimal number to an 8) the transfer amount must be divisible to leave no dust. Not the best user experience, so we could alternatively return the dust to their account.

Only supports the evm native asset bridge.

## Checklist


x/evmutil/keeper/conversion_evm_native.go
```diff
@@ -222,3 +239,19 @@ func (k Keeper) LockERC20Tokens(
 
 	return err
 }
+
+// MulByPowersTen computes i*10^power and returns integers for the part before and after the decimal point.
+func MulByPowersTen(i *big.Int, power int64) (*big.Int, *big.Int) {
+	positivePower := big.NewInt(power)
```
galxy25, line 245: do we need to do some validation of inputs here, return an error if `i` or `power` is outside of the expected range?

x/evmutil/types/errors.go
```diff
@@ -11,4 +11,5 @@ var (
 	ErrUnexpectedContractEvent = errorsmod.Register(ModuleName, 6, "unexpected contract event")
 	ErrInvalidCosmosDenom      = errorsmod.Register(ModuleName, 7, "invalid cosmos denom")
 	ErrSDKConversionNotEnabled = errorsmod.Register(ModuleName, 8, "sdk.Coin not enabled to convert to ERC20 token")
+	ErrInvalidConversionAmount = errorsmod.Register(ModuleName, 8, "amount cannot be converted")
```
galxy25, line 14: could three splitting this generic error out into 2 more specific ones (for the conversion amount being two small or too large) if we keep the semantics of "you can't convert an amount that includes dust"

x/evmutil/keeper/conversion_evm_native.go
rhuairahrighairidh: yeah validation would be good, not sure what the limits of `big.Int`s are, but there should be at least some validation in the `decimal_conversion` param because we'll never be converting from like 1 decimal place to 1000 places.

x/evmutil/keeper/conversion_evm_native.go
```diff
@@ -102,13 +109,22 @@ func (k Keeper) ConvertERC20ToCoin(
 		return err
 	}
 
+	// compute amount * 10^pair.DecimalConversion
+	// If DecimalConversion is negative the output will be smaller than the input, with a potential decimal part.
+	// This just errors if there is a decimal part, so users must pad their conversion amounts with zeros.
+	// Alternatively we could only transfer the non decimal part, leaving the rest on the evm side.
```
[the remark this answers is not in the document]
rhuairahrighairidh: Yeah this is probably the better approach. I prefer txs that move the exact amount specified in the tx, but the errors are probably going to cause confusion

PR #1838, opened 2024-02-29
fix: update vesting tests to use fixed block time

pirtleshell:
## Description
fixes broken unit test by updating the constant


x/validator-vesting/keeper/grpc_query_test.go
```diff
@@ -91,7 +91,7 @@ func (suite *grpcQueryTestSuite) TestCirculatingSupplyUSDX() {
 func (suite *grpcQueryTestSuite) TestCirculatingSupplySWP() {
 	res, err := suite.queryClient.CirculatingSupplySWP(context.Background(), &types.QueryCirculatingSupplySWPRequest{})
 	suite.Require().NoError(err)
-	suite.Require().Equal(sdkmath.NewInt(198437490), res.Amount)
+	suite.Require().Equal(sdkmath.NewInt(201302073), res.Amount)
```
pirtleshell, line 94: @DracoLi can you verify that this new value is correct? or was a regression introduced somehow?

pirtleshell: jk, your value was valid two days ago, but now it's this value 😆 

updated these tests to use a constant bocktime

DracoLi (approved):

PR #1839, opened 2024-03-04
Add ibc packet forward middleware for ibc transfer

DracoLi:
## Description
- Uses `packet-forward-middleware v7.1.2` and not v8 since v8 does not support cosmos-sdk v47
- Follows integration guide here https://github.com/cosmos/ibc-apps/blob/middleware/packet-forward-middleware/v7.1.2/middleware/packet-forward-middleware/docs/integration.md
- e2e tests will be added in a later PR
- **Migration Requirements**: Will have a follow up PR in v26 release branch to add `packetforward` store key.

## Checklist


galxy25 (changes requested):
approved pending completion of checklist items for

- adding migrations 
- filling out the security questionare for the feature 
- New features/capabilities have an e2e test
- (cc at a higher level, for scoping the release @karzak @rhuairahrighairidh do we need to do an additional security review for packet fordwarding or feel that we've done enough internal due diligence / the feature is widely deployed and battle tested?)

PR #1842, opened 2024-03-06
only deploy internal testnet if deploy version or genesis state changes

galxy25:
## Description
only deploy internal testnet if deploy version or genesis state changes

## Checklist

### DIFF
--- a/.github/workflows/cd-internal-testnet.yml
+++ b/.github/workflows/cd-internal-testnet.yml
@@ -1,5 +1,6 @@
 name: Continuous Deployment (Internal Testnet)
 # run after every successful CI job of new commits to the master branch
+# if deploy version or config has changed
 on:
   workflow_run:
     workflows: [Continuous Integration (Kava Master)]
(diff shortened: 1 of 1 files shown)


.github/workflows/cd-internal-testnet.yml
```diff
@@ -1,21 +1,40 @@
...
+        - uses: actions/checkout@v4
+          with:
+            fetch-depth: 0 # OR "2" -> To retrieve the preceding commit.
+        - name: Get all changed internal testnet files
+          id: changed-internal-testnet-config
+          uses: tj-actions/changed-files@v42
+          with:
+            # Avoid using single or double quotes for multiline patterns
+            files: |
+              ci/env/kava-internal-testnet/**
```
DracoLi, line 26: might be an issue if internal testnet version is set to `MASTER`. But might be okay? Since protonet can be used if latest testnet is needed.

DracoLi (approved):

.github/workflows/cd-internal-testnet.yml
galxy25: not following, if they set the version to master, that will count as a changed file causing internal testnet to deploy. If they want to deploy a different version, they will have to change the version again causing a change to be detected and internal testnet deployed.

DracoLi: what I mean is change to MASTER -> then internal testnet will be deployed as expected. Then on future master merges, the change file here will indicate no version or genesis changes, and thus not redeploy on master changes.

PR #1848, opened 2024-03-15
Implement bep3 evm native conversion logic

DracoLi:
## Checklist

### DIFF
--- a/x/evmutil/keeper/conversion_evm_native.go
+++ b/x/evmutil/keeper/conversion_evm_native.go
@@ -71,7 +71,15 @@ func (k Keeper) ConvertCoinToERC20(
 		return err
 	}
 
-	if err := k.UnlockERC20Tokens(ctx, pair, coin.Amount.BigInt(), receiverAccount); err != nil {
+	// handle bep3 conversion pair (8 decimals sdk.Coin -> 18 decimals erc20)
+	// when converting bep3 bank coins to erc20, we will need to unlock the 18
+	// decimals erc20 equivalent of the 8 decimals bep3 sdk.Coins
+	amountToUnlock := coin.Amount.BigInt()
+	if IsEvmNativeBep3Conversion(pair) {
+		amountToUnlock = ConvertBep3CoinAmountToERC20Amount(coin.Amount.BigInt())
+	}
+
+	if err := k.UnlockERC20Tokens(ctx, pair, amountToUnlock, receiverAccount); err != nil {
 		return err
 	}
 
@@ -102,13 +110,32 @@ func (k Keeper) ConvertERC20ToCoin(
 		return err
 	}
 
+	amountToLock := amount.BigInt()
+	amountToMint := amount.BigInt()
+
+	// handle bep3 conversion pair (18 decimals erc20 to 8 decimals sdk.Coin)
+	// When converting bep3 erc20 assets to sdk.Coin, the following will happen
+	// 1. Convert provided 18 decimals erc20 amount to 8 decimals sdk amount, then mint those.
+	//    Dust from converting 18 decimals to 8 decimals are ignored and will
+	// 		remain in the original wallet.
+	// 2. Convert to-be-minted 8 decimals sdk.Coin back to 18 decimals, then lock those.
+	if IsEvmNativeBep3Conversion(pair) {
+		amountToMint, _ = ConvertBep3ERC20AmountToCoinAmount(amount.BigInt())
+
+		// make sure we have at least 1 sdk.Coin to mint
+		if amountToMint.Cmp(big.NewInt(0)) == 0 {
+			return errorsmod.Wrapf(types.ErrInsufficientConversionAmount, "unable to convert bep3 coin due converting less than 1%s", pair.Denom)
+		}
+		amountToLock = ConvertBep3CoinAmountToERC20Amount(amountToMint)
+	}
+
 	// lock erc20 tokens
-	if err := k.LockERC20Tokens(ctx, pair, amount.BigInt(), initiator); err != nil {
+	if err := k.LockERC20Tokens(ctx, pair, amountToLock, initiator); err != nil {
 		return err
 	}
 
 	// mint conversion pair coin
-	coin, err := k.MintConversionPairCoin(ctx, pair, amount.BigInt(), receiver)
+	coin, err := k.MintConversionPairCoin(ctx, pair, amountToMint, receiver)
 	if err != nil {
 		return err
 	}
--- a/x/evmutil/keeper/conversion_evm_native_bep3.go
+++ b/x/evmutil/keeper/conversion_evm_native_bep3.go
@@ -1,0 +1,38 @@
+package keeper
+
+import (
+	"math/big"
+
+	"github.com/kava-labs/kava/x/evmutil/types"
+)
+
+var (
+	DefaultBEP3ConversionDenoms = []string{"bnb", "busd", "btcb", "xrpb"}
+)
+
+// IsEvmNativeBep3Conversion returns true if the given pair is a BEP3 conversion pair.
+func IsEvmNativeBep3Conversion(pair types.ConversionPair) bool {
+	for _, denom := range DefaultBEP3ConversionDenoms {
+		if pair.Denom == denom {
+			return true
+		}
+	}
+	return false
+}
+
+// ConvertBep3CoinAmountToERC20Amount converts a bep3 coin amount with 8 decimals
+// to the equivalent ERC20 token with 18 decimals.
+func ConvertBep3CoinAmountToERC20Amount(amount *big.Int) *big.Int {
+	multiplier := new(big.Int).Exp(big.NewInt(10), big.NewInt(10), nil)
+	result := new(big.Int).Mul(amount, multiplier) // amount * 10^10
+	return result
+}
+
+// ConvertBep3ERC20AmountToCoinAmount converts a bep3 ERC20 token with 18 decimals
+// to the equivalent coin amount with 8 decimals, and returning the remainder.
+func ConvertBep3ERC20AmountToCoinAmount(amount *big.Int) (*big.Int, *big.Int) {
+	divisor := new(big.Int).Exp(big.NewInt(10), big.NewInt(10), nil)
+	quotient := new(big.Int).Div(amount, divisor)
+	remainder := new(big.Int).Mod(amount, divisor)
+	return quotient, remainder
+}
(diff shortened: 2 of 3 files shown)


x/evmutil/keeper/conversion_evm_native_bep3.go
```diff
@@ -0,0 +1,38 @@
+package keeper
+
+import (
+	"math/big"
+
+	"github.com/kava-labs/kava/x/evmutil/types"
+)
+
+var (
+	DefaultBEP3ConversionDenoms = []string{"bnb", "busd", "btcb", "xrpb"}
```
rhuairahrighairidh, line 10: nit, I'd make this variable private (or use constants) to avoid other packages being able to modify it. Low chance of it being a problem but adding unauthorised assets here would allow minting of coins

x/evmutil/keeper/conversion_evm_native.go
```diff
@@ -102,13 +110,32 @@ func (k Keeper) ConvertERC20ToCoin(
...
+	// When converting bep3 erc20 assets to sdk.Coin, the following will happen
+	// 1. Convert provided 18 decimals erc20 amount to 8 decimals sdk amount, then mint those.
+	//    Dust from converting 18 decimals to 8 decimals are ignored and will
+	//    remain in the original wallet.
+	// 2. Convert to-be-minted 8 decimals sdk.Coin back to 18 decimals, then lock those.
+	if IsEvmNativeBep3Conversion(pair) {
+		amountToMint, _ = ConvertBep3ERC20AmountToCoinAmount(amount.BigInt())
+
+		// make sure we have at least 1 sdk.Coin to mint
+		if amountToMint.Cmp(big.NewInt(0)) == 0 {
```
rhuairahrighairidh, line 126: good catch on this edge case, would be good to cover in a test (in this or levi's branch)

rhuairahrighairidh (commented):
looks good

x/evmutil/keeper/conversion_evm_native.go
```diff
@@ -71,7 +71,15 @@ func (k Keeper) ConvertCoinToERC20(
 		return err
 	}
 
-	if err := k.UnlockERC20Tokens(ctx, pair, coin.Amount.BigInt(), receiverAccount); err != nil {
+	// handle bep3 conversion pair (8 decimals sdk.Coin -> 18 decimals erc20)
+	// when converting bep3 bank coins to erc20, we will need to unlock the 18
+	// decimals erc20 equivalent of the 8 decimals bep3 sdk.Coins
+	amountToUnlock := coin.Amount.BigInt()
+	if IsEvmNativeBep3Conversion(pair) {
```
nddeluca, line 78: After a map is used & conversion factor is pulled to a top level variable, I wonder if 3 private methods are necessary.

option 1
```golang
// coin -> erc20
amountToUnlock := coin.Amount
if _, isBep3Asset := bep3Denoms[coin.Denom]; isBep3Asset {
  amountToUnlock = amountToUnlock.Mul(bep3ConversionFactor)
}

// erc20 -> coin
amountToLock := amount
amountToMint := amount
if _, isBep3Asset := bep3Denoms[pair.Denom]; isBep3Asset {
  amountToMint = amountToMint.Quo(bep3ConversionFactor)

  if amountToMint.IsZero() {
    return errorsmod.Wrapf(types.ErrInsufficientConversionAmount, "unable to convert bep3 coin due converting less than 1%s", pair.Denom)
  }

  amountToLock := amountToMint.Mul(bep3ConversionFactor)
}
```

vs option 2

```golang
// coin -> erc20
amountToUnlock := coin.Amount
if isBep3Asset(coin.Denom) {
  amountToUnlock = covertBep3CoinToErc20(amountToUnlock)
}

// erc20 -> coin
amountToLock := amount
amountToMint := amount
if isBep3Asset(pair.Denom) {
  amountToMint := convertBep3Erc20ToCoin(amountToMint)

  if amountToMint.IsZero() {
    return errorsmod.Wrapf(types.ErrInsufficientConversionAmount, "unable to convert bep3 coin due converting less than 1%s", pair.Denom)
  }

  amountToLock := convertBep3CoinToErc20(amountToLock)
```

vs option 3

```golang
// coin -> erc20
amountToUnlock := convertToErc20Decimals(coin)

// erc20 -> coin
amountToLock, amountToMint, err := calculateAmountToLockAndMint(pair.Denom, amount)
if err != nil {
  return err
}
```

- Option 1 is the default implementation, but doesn't match the complexity level of the calling function well.
- Option 2 it may use some named functions, but is the complexity of the top level function actually reduced?  Also, is `amountToLock := convertBep3CoinToErc20(amountToLock)` actually more comprehensible than `amountToLock := amountToMint.Mul(bep3ConversionFactor)`?  Enough so that the extra function is worth it?
- Option 3 naming could be better, but this pushes the complexity down a level -- improving readability at the top level, but also keeping the implementation easy to find & understand.  IMO, this seems like the best balance.

nddeluca (changes requested):
Really good test cases covering dust, too small of balance, and insuffient balances on both the sdk and erc20 sides.

Let's just add some defensive testing to ensure all four bep3 assets are covered. In addition to using a map for denom lookup, removing unused code, and making extracted helpers private (or adding unit tests for them).

Looks great otherwise.  Refactoring implementation is optional -- just added my thoughts there on readability & complexity.

x/evmutil/keeper/conversion_evm_native_bep3.go
nddeluca: Good catch @rhuairahrighairidh

x/evmutil/keeper/conversion_evm_native.go
DracoLi: I do prefer option 2's separation of the bep3 logic with the rest of the conversion logic. It's very clear when reading the conversion functions that bep3 assets are handled differently, and we can easily remove that logic in the future without touching any existing conversion logic. 
I do also think abstracting some of the bep3 logic will help with readability, especially for the erc20 -> coin conversion logic. I'll test out a hybrid of option 2 and 3's approach.

nddeluca: Cool 👍. Ya do agree option 3 doesn't clearly show it's for bep3 only, hybrid approach would be good

DracoLi: https://github.com/Kava-Labs/kava/pull/1848/commits/98fc5962aaee656d87004abbf367211b1db70eca

x/evmutil/keeper/conversion_evm_native_bep3.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	"math/big"
+
+	errorsmod "cosmossdk.io/errors"
+	"github.com/kava-labs/kava/x/evmutil/types"
+)
+
+var (
+	bep3Denoms = map[string]bool{
```
nddeluca, line 11: Nice, normally I'd say use a `map[string]struct{}`, but with 4 items the benefited clarify and simpler logic in `isBep3Asset` is nice 👍

x/evmutil/keeper/conversion_evm_native_bep3.go
```diff
@@ -0,0 +1,55 @@
...
+	bep3ConversionFactor = new(big.Int).Exp(big.NewInt(10), big.NewInt(10), nil)
+)
+
+func isBep3Asset(denom string) bool {
+	return bep3Denoms[denom]
+}
+
+// convertBep3CoinAmountToERC20Amount converts a bep3 coin amount with 8 decimals
+// to the equivalent ERC20 token with 18 decimals.
+func convertBep3CoinAmountToERC20Amount(amount *big.Int) *big.Int {
```
nddeluca, line 27: Just a personal preference and no need to change, but I like named return arguments for functions like this.

Ex
```golang
func convertBep3CoinAmountToERC20Amount(amount *big.Int) (result *big.Int) {
	return new(big.Int).Mul(amount, bep3ConversionFactor)
}
func convertBep3ERC20AmountToCoinAmount(amount *big.Int) (quotient *big.Int) {
	return new(big.Int).Div(amount, bep3ConversionFactor)
}

File: x/evmutil/keeper/conversion_evm_native_bep3.go [hunk 6]
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// convertBep3ERC20AmountToCoinAmount converts a bep3 ERC20 token with 18 decimals
+// to the equivalent coin amount with 8 decimals, and dropping the remainder.
+func convertBep3ERC20AmountToCoinAmount(amount *big.Int) *big.Int {
+	quotient := new(big.Int).Div(amount, bep3ConversionFactor)
+	return quotient
+}
+
+// bep3ERC20AmountToCoinMintAndERC20LockAmount converts 18 decimals erc20 bep3
```
nddeluca, line 39: appreciate the doc comments 👍

x/evmutil/keeper/conversion_evm_native_bep3_test.go
```diff
@@ -0,0 +1,415 @@
...
+var (
+	bep3Denoms = []string{"bnb", "busd", "btcb", "xrpb"}
+)
+
+func TestBep3ConversionTestSuite(t *testing.T) {
+	suite.Run(t, new(Bep3ConversionTestSuite))
+}
+
+func (suite *Bep3ConversionTestSuite) TestConvertCoinToERC20_Bep3() {
+	for _, denom := range bep3Denoms {
```
nddeluca, line 28: This actually makes the tests quite a bit easier to read -- easy to see that all denoms are tested and much more resistant to refactorings that could drop a denom, accidentally rename one, break the check some other unique way, etc

x/evmutil/keeper/conversion_evm_native.go
```diff
@@ -102,13 +110,23 @@ func (k Keeper) ConvertERC20ToCoin(
 		return err
 	}
 
+	amountToLock := amount.BigInt()
+	amountToMint := amount.BigInt()
+
+	if isBep3Asset(pair.Denom) {
```
nddeluca, line 116: I see what you mean know, and this is a nice approach showing there is are two paths in the logic: One for bep3 assets, and one for not.  Also, makes it clear that the amounts are only modified in the bep3 case, but also keeps the dirty details down a level in a function call.

nddeluca (approved):

x/evmutil/keeper/conversion_evm_native_bep3.go
DracoLi: good point, i think this will benefit here

PR #1851, opened 2024-03-19
Optimize Pricefeed EndBlocker

nddeluca:
Added tests that cover 100% of legacy behavior (see TestKeeper_SetCurrentPrices_MatchesAllMarketsBehavior), and added an optimized version of set current prices for all markets.

### DIFF
--- a/x/pricefeed/keeper/keeper.go
+++ b/x/pricefeed/keeper/keeper.go
@@ -131,6 +131,74 @@ func (k Keeper) SetCurrentPrices(ctx sdk.Context, marketID string) error {
 	return nil
 }
 
+// SetCurrentPricesForAllMarkets updates the price of an asset to the median of all valid oracle inputs
+func (k Keeper) SetCurrentPricesForAllMarkets(ctx sdk.Context) {
+	orderedMarkets := []string{}
+	marketPricesByID := make(map[string]types.CurrentPrices)
+
+	for _, market := range k.GetMarkets(ctx) {
+		if market.Active {
+			orderedMarkets = append(orderedMarkets, market.MarketID)
+			marketPricesByID[market.MarketID] = types.CurrentPrices{}
+		}
+	}
+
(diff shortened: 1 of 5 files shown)


x/pricefeed/keeper/keeper.go
```diff
@@ -131,6 +131,74 @@ func (k Keeper) SetCurrentPrices(ctx sdk.Context, marketID string) error {
...
+		}
+
+		notExpiredPrices, _ := marketPricesByID[marketID]
+
+		if len(notExpiredPrices) == 0 {
+			// NOTE: The current price stored will continue storing the most recent (expired)
+			// price if this is not set.
+			// This zero's out the current price stored value for that market and ensures
+			// that CDP methods that GetCurrentPrice will return error.
+			k.setCurrentPrice(ctx, marketID, types.CurrentPrice{})
```
nddeluca, line 178: Verified that removing this line fails tests.

x/pricefeed/keeper/keeper.go
```diff
@@ -131,6 +131,74 @@ func (k Keeper) SetCurrentPrices(ctx sdk.Context, marketID string) error {
...
+			continue
+		}
+
+		medianPrice := k.CalculateMedianPrice(notExpiredPrices)
```
nddeluca, line 182: Tests fail if mean is used instead of median

x/pricefeed/keeper/keeper.go
```diff
@@ -131,6 +131,74 @@ func (k Keeper) SetCurrentPrices(ctx sdk.Context, marketID string) error {
...
+
+		// check case that market price was not set in genesis
+		//if validPrevPrice && !medianPrice.Equal(prevPrice.Price) {
+		if validPrevPrice && !medianPrice.Equal(prevPrice.Price) {
```
nddeluca, line 186: Tested event emission only occurs when this logic statement is true

galxy25 (approved):

nddeluca:
Merging to master, but not back-porting to v0.26.x until after release

PR #1865, opened 2024-03-28
Set the CDP Block Interval to 100 during v0.26.x upgrade

nddeluca:
This updates the CDP parameters to have a liquidation block interval of 100, instead of the default value of 1.

Since the interest synchronization and the liquidation process makes up a substantial portion of block time, this has a significant impact on block execution performance.

The included E2E ensures that no other parameters are changed from before upgrade along with verifying that the liquidation interval is correctly set.

### DIFF
--- a/tests/e2e/e2e_upgrade_handler_test.go
+++ b/tests/e2e/e2e_upgrade_handler_test.go
@@ -8,6 +8,7 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	consensustypes "github.com/cosmos/cosmos-sdk/x/consensus/types"
 	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types/v1"
+	cdptypes "github.com/kava-labs/kava/x/cdp/types"
 )
 
 func (suite *IntegrationTestSuite) TestUpgradeParams_SDK() {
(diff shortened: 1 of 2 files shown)


tests/e2e/e2e_upgrade_handler_test.go
```diff
@@ -98,6 +99,26 @@ func (suite *IntegrationTestSuite) TestUpgradeParams_Consensus() {
...
+
+	paramsBefore, err := grpcClient.Query.Cdp.Params(beforeUpgradeCtx, &cdptypes.QueryParamsRequest{})
+	suite.Require().NoError(err)
+	paramsAfter, err := grpcClient.Query.Cdp.Params(afterUpgradeCtx, &cdptypes.QueryParamsRequest{})
+	suite.Require().NoError(err)
+
+	expectedParams := paramsBefore.Params
+	expectedParams.LiquidationBlockInterval = int64(100)
+
+	suite.Require().Equal(expectedParams, paramsAfter.Params,
```
nddeluca, line 118: Tested that this failed when modifying another parameter beyond the block interval

karzak:
LGTM - Double checked in `x/cdp/abci_test.go` that we have coverage which shows that begin blocker runs only (and definitely runs) at `LiquidationBlockInterval - looked good.

karzak (approved):

karzak:
After a bit more thought, I would argue setting `LiquidationBlockInterval` to `50` instead `100` - `50` gives worst case of running roughly once every 6 minutes - this seems preferable to every 12 minutes in cases where there is a backlog of risky cdps. 

Thoughts @rhuairahrighairidh?

karzak (changes requested):
See comment on updating the value - otherwise LGTM

rhuairahrighairidh (commented):
looks good, agree with kevin the more frequent liquidations is a bit better

PR #1898, opened 2024-05-01
ci: extract separate rocksdb base image

pirtleshell:
## Description
I pulled out the golang & rocksdb installation into a base container that is published to `kava/rocksdb` docker repo. This should shorten the rocksdb container build time on merge to master by ~1.5hr.

Additionally, I updated the docker build & push CI commands to inject the go build cache into the docker building which should further shorten the Github Action CI time for merges to master.

### DIFF
--- a/.github/workflows/ci-docker.yml
+++ b/.github/workflows/ci-docker.yml
@@ -50,6 +50,17 @@ jobs:
           username: ${{ inputs.dockerhub-username }}
           password: ${{ secrets.CI_DOCKERHUB_TOKEN }}
 
+      - name: Go Build Cache for Docker
+        uses: actions/cache@v3
+        with:
+          path: go-build-cache
+          key: ${{ runner.os }}-go-build-cache-${{ hashFiles('**/go.sum') }}
+
+      - name: inject go-build-cache into docker
+        uses: reproducible-containers/buildkit-cache-dance@v2.1.2
+        with:
+          cache-source: go-build-cache
+
       # publish to docker hub, tag with short git hash
       - name: Build and push (goleveldb)
         uses: docker/build-push-action@v5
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-docker.yml
```diff
@@ -50,6 +50,17 @@ jobs:
...
+      - name: Go Build Cache for Docker
+        uses: actions/cache@v3
+        with:
+          path: go-build-cache
+          key: ${{ runner.os }}-go-build-cache-${{ hashFiles('**/go.sum') }}
+
+      - name: inject go-build-cache into docker
+        uses: reproducible-containers/buildkit-cache-dance@v2.1.2
+        with:
+          cache-source: go-build-cache
```
drklee3, line 62: Do the Dockerfiles need to be updated for this to work?

With those `--mount=type=cache,target=...` [flags for RUN commands](https://docs.docker.com/build/ci/github-actions/cache/#cache-mounts)

pirtleshell: they have the appropriate cache mounts, which are currently working and used when building locally. this addition (hopefully) gives the github action caching powers as well

drklee3: ahh yes I missed it, mistakenly only checked the rocksdb-base 😄 seems good then!

drklee3 (approved):

.github/workflows/ci-docker.yml
pirtleshell: yee! i think some sort of cache mount for the building-rocksdb-command is what would allow caching of that container layer, but i didn't take the time to figure out what 😂 just opted to extract as a separate container for the same result..

PR #1901, opened 2024-05-06
test(e2e-ibc): downgrade to ibc v7 for ibc tests

pirtleshell:
## Description

Uses IBC v7 & kava's overrides for cosmos-sdk, cometbft, & ethermint in the IBC e2e tests set up with interchain.

Conformance tests are included. They pass, but still intermittently fail with errors that look like:
```
tendermint rpc client status: post failed: Post "http://127.0.0.1:<port>": dial tcp 127.0.0.1:<port>: connect: can't assign requested address
```

my current theory for these failures is a lack of rate limiting from the conformance tests (too many requests made to a particular port). though the pass rate was greatly improved by reducing the number of nodes (using only 1 validator for both chains in the conformance tests), which might be indicative of my system running out of available ports to bind to.

we shall see how the tests fair in the CI.

Using the same dependency versions in the interchain tests as we do in the chain will make it easier to inject complex genesis state during chain startup.


tests/e2e-ibc/conformance_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/strangelove-ventures/interchaintest/v7/conformance"
+	"github.com/strangelove-ventures/interchaintest/v7/ibc"
+	"github.com/strangelove-ventures/interchaintest/v7/testreporter"
+	"github.com/stretchr/testify/require"
+
+	kavainterchain "github.com/kava-labs/kava/tests/interchain"
+)
+
+var (
+	zero = 0
```
nddeluca, line 21: These could be named based on what they are used for in addition to being constants?
```
const (
  numFullNodes = 0
  numValidators = 1
)
```

pirtleshell: yup yup haha they could and should 👍

PR #1907, opened 2024-05-09
feat(x/precisebank): Add FractionalBalance types

drklee3:
## Description
- Add necessary types to track account fractional balances.
- Add FractionalBalance type to genesis
- Does __not__ include internal module remainder balance, next PR.

## Checklist


x/precisebank/types/fractional_balance.go
```diff
@@ -0,0 +1,49 @@
...
+	sdkmath "cosmossdk.io/math"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// FractionalBalance contains __only__ the fractional balance of an address.
+// We want to extend the current KAVA decimal digits from 6 to 18, thus 12 more
+// digits are added to the fractional balance. With 12 digits, the maximum
+// value of the fractional balance is 1_000_000_000_000 - 1.
+// We subtract 1, as 1 more will roll over to the integer balance.
+var MAX_FRACTIONAL_AMOUNT = sdkmath.NewInt(1_000_000_000_000).SubRaw(1)
```
nddeluca, line 15: Since this can't be a `const`, I'd lean towards making this private within the module and offer a public function outside consumers (if any).

x/precisebank/types/fractional_balances.go
```diff
@@ -0,0 +1,42 @@
...
+
+// Validate returns an error if any FractionalBalance in the slice is invalid.
+func (fbs FractionalBalances) Validate() error {
+	seenAddresses := make(map[string]struct{})
+
+	for _, fb := range fbs {
+		// Ensure there is no white-space before/after address as that can
+		// bypass duplicate address check.
+		// Upper/lowercase is not an issue as the address is a bech32 encoded
+		// string which only allows lowercase.
```
nddeluca, line 24: This comment conflicts with the test case the error message `decoding bech32 failed: string not all lowercase or all uppercase` in FractionalBalance validation test cases.

x/precisebank/types/fractional_balance.go
```diff
@@ -0,0 +1,49 @@
...
+	return FractionalBalance{
+		Address: address,
+		Amount:  amount,
+	}
+}
+
+// GetAddress returns the sdk.AccAddress of the FractionalBalance. Assumes that
+// the FractionalBalance is valid.
+func (fb FractionalBalance) GetAddress() sdk.AccAddress {
+	addr, _ := sdk.AccAddressFromBech32(fb.Address)
```
nddeluca, line 29: We should panic if the assumption is broken instead of silently failing & returning a nil address.

In addition -- I don't see this used anywhere so it needs to be verified in a new test, existing test, or deleted.

x/precisebank/types/fractional_balance.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+// Validate returns an error if the FractionalBalance has an invalid address or
+// negative amount.
+func (fb FractionalBalance) Validate() error {
+	if _, err := sdk.AccAddressFromBech32(fb.Address); err != nil {
+		return err
+	}
+
+	if fb.Amount.IsNegative() {
```
nddeluca, line 40: Should we allow a zero balance?  It seems like no record in genesis should be implied as zero, and we should fail fast by ensuring a zero fractional balance is invalid here.

x/precisebank/types/fractional_balance.go
nddeluca: Though I also assume at some point, this max will be based on the default conversion factor, but the above should still apply.

x/precisebank/types/fractional_balances_test.go
```diff
@@ -0,0 +1,79 @@
...
+				types.NewFractionalBalance(sdk.AccAddress{1}.String(), sdkmath.NewInt(100)),
+				types.NewFractionalBalance(sdk.AccAddress{1}.String(), sdkmath.NewInt(100)),
+			},
+			"duplicate address: kava1qy0xn7za",
+		},
+		{
+			"invalid - duplicate address with whitespace",
+			types.FractionalBalances{
+				types.NewFractionalBalance(sdk.AccAddress{1}.String(), sdkmath.NewInt(100)),
+				types.NewFractionalBalance(" "+sdk.AccAddress{1}.String(), sdkmath.NewInt(100)),
```
nddeluca, line 61: Is whitespace in an address valid? Should we re-order checking if a fractional balance is valid so we can ensure that our uniqueness check only happens on a valid collection of FractionalBalance's?

x/precisebank/types/fractional_balance_test.go
```diff
@@ -0,0 +1,140 @@
...
+			"invalid - negative amount",
+			"kava1gpxd677pp8zr97xvy3pmgk70a9vcpagsakv0tx",
+			sdkmath.NewInt(-100),
+			"negative amount: -100",
+		},
+		{
+			"invalid - max amount + 1",
+			"kava1gpxd677pp8zr97xvy3pmgk70a9vcpagsakv0tx",
+			types.MAX_FRACTIONAL_AMOUNT.AddRaw(1),
+			"amount exceeds max of 999999999999: 1000000000000",
```
nddeluca, line 91: It seems like `amount 1000000000000 exceeds max of of 999999999999` would read better and I believe it's best to reserve `:` for cases of wrapped errors

x/precisebank/types/fractional_balances.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	fmt "fmt"
+	"strings"
+)
+
+// FractionalBalances is a slice of FractionalBalance
+type FractionalBalances []FractionalBalance
+
+// Len returns the length of the slice.
+func (fbs FractionalBalances) Len() int {
```
nddeluca, line 12: This isn't used anywhere and not tested.  Remove or add an test? Tho I'm not sure I see a lot of genesis data being length checked, so do we need this?

nddeluca (changes requested):
Minor stuff and some slight cleanup needed. I like the overall direction and the strong validations and testing 👍

x/precisebank/types/fractional_balances_test.go
drklee3: ah right, I misunderstood the TrimSpace in `AccAddressFromBech32()`, whitespace in addresses is not valid

x/precisebank/types/fractional_balance.go
drklee3: yeah that's true, zero balances should instead be deleted from store if its set, and not allow 0 balance in genesis

x/precisebank/types/fractional_balances.go
drklee3: yeah don't need this, removed

x/precisebank/types/genesis.go
```diff
@@ -1,17 +1,30 @@
...
+// any failed validation criteria.
 func (gs *GenesisState) Validate() error {
+	if err := gs.Balances.Validate(); err != nil {
+		return fmt.Errorf("invalid balances: %w", err)
+	}
+
+	// TODO:
+	// - Validate remainder amount
+	// - Validate sum(fractionalBalances) + remainder = whole integer value
+	// - Cannot validate here: reserve account exists & balance match
```
pirtleshell, line 15: this only gets validated as a chain invariant, yes?

pirtleshell (approved):
a few nits and notes
looks great though. nice test coverage for the validations

x/precisebank/types/fractional_balance.go
pirtleshell: i didn't see this comment. i'd prefer it to be an exported value rather than a getter method for an underlying value. IMO it's more idiomatic

though it should be named `MaxFractionalAmount`, and not all caps.

my two cents anyway

x/precisebank/types/fractional_balance.go
```diff
@@ -0,0 +1,51 @@
...
+// We want to extend the current KAVA decimal digits from 6 to 18, thus 12 more
+// digits are added to the fractional balance. With 12 digits, the maximum
+// value of the fractional balance is 1_000_000_000_000 - 1.
+// We subtract 1, as 1 more will roll over to the integer balance.
+var maxFractionalAmount = sdkmath.NewInt(1_000_000_000_000).SubRaw(1)
+
+// GetMaxFractionalAmount returns the maximum value of a FractionalBalance.
+func GetMaxFractionalAmount() sdkmath.Int {
+	return maxFractionalAmount
+}
```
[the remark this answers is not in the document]
nddeluca: Do you mean something like
```golang
const MaxFractionalValue = int64(1_000_000_000_000-1)

var maxFractionalAmount := sdk.NewInt(MaxFractionalValue)
```
(disregarding the poor naming)?  We can't have a const sdk.Int value.

proto/kava/precisebank/v1/genesis.proto
```diff
@@ -1,7 +1,32 @@
...
+// FractionalBalance defines the fractional portion of an account balance
+message FractionalBalance {
+  option (gogoproto.equal) = false;
+  option (gogoproto.goproto_getters) = false;
+
+  // address is the address of the balance holder.
+  string address = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+
+  // amount indicates amount of only the fractional balance owned by the address.
+  string amount = 2 [
```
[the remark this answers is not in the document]
nddeluca: Good call out and agree, could be an extension later but not worth adding the complexity now

x/precisebank/types/genesis.go
nddeluca: The invariant can validate that, and we can validate other forms of the same validation during execution.

However, it seems like a good thing to test in InitGenesis? We can query the reserve balance on init to ensure the starting state of the module is valid.

drklee3: yup the last point we can validate in InitGenesis, just can't do it in this Validate() method since it needs some other keeper methods

proto/kava/precisebank/v1/genesis.proto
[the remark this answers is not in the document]
drklee3: added a note for this

x/precisebank/types/fractional_balance.go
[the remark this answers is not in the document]
pirtleshell: sorry "constant" was the wrong choice of word. i was imagining
```go
var MaxFractionalAmount := sdkmath.NewInt(1_000_000_000_000).SubRaw(1)
```
but it was a brain fart. i don't actually want that exposed in a way that somebody somewhere could change it.

I want `MaxFractionalAmount()` without `Get` 😄

x/precisebank/types/fractional_balance.go
drklee3: hmm yeah that does seem cleaner, or a 3rd option being like the same as `sdkmath.ZeroInt()`?

```go
func MaxFractionalInt() sdkmath.Int { return sdkmath.NewInt(1_000_000_000_000).SubRaw(1) }
```

Some overhead of making a new one each time its called though, not entirely sure why they do that since Ints should be immutable with via public methods

drklee3: ah wait I guess not using a global var prevents it being re-assigned a new value.

var MaxFractionalInt = ...

makes it possible to do `MaxFractionalInt = sdk.NewInt(xx)` when the func should be easier to enforce it isn't mutable

x/precisebank/types/fractional_balance.go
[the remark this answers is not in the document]
pirtleshell: though, worth noting that the sdk does sometimes expose public values. but i think only when they think its ok to change them.
like [`DefaultPowerReduction`](https://github.com/Kava-Labs/cosmos-sdk/blob/28801fc666f1c33f315e815bd57f2952fd6e0b00/types/staking.go#L28)
it's only a getter in the keeper [here](https://github.com/Kava-Labs/cosmos-sdk/blob/28801fc666f1c33f315e815bd57f2952fd6e0b00/x/staking/keeper/params.go#L48)

but i don't think it applies here. a func wrapping the private var is good 👍 i'm just recommending (bikeshedding) the name `MaxFractionalAmount()` over `GetMaxFractionalAmount()`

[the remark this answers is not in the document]
drklee3: yeah I think keeping func & name without Get is good - similar to those sdkmath.MaxInt() / OneInt() / etc methods 

https://github.com/Kava-Labs/kava/pull/1907#discussion_r1597190673

pirtleshell (approved):

x/precisebank/types/fractional_balance_test.go
```diff
@@ -0,0 +1,164 @@
...
+	"math/big"
+	"testing"
+
+	sdkmath "cosmossdk.io/math"
+	"github.com/kava-labs/kava/app"
+	"github.com/kava-labs/kava/x/precisebank/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestMaxFractionalAmount_Immutable(t *testing.T) {
```
nddeluca, line 13: Great tests here

x/precisebank/types/fractional_balance_test.go
```diff
@@ -0,0 +1,164 @@
...
+		"max amount should be immutable",
+	)
+}
+
+func TestMaxFractionalAmount_Copied(t *testing.T) {
+	max1 := types.MaxFractionalAmount().BigIntMut()
+	max2 := types.MaxFractionalAmount().BigIntMut()
+
+	// Checks that the returned two pointers do not reference the same object
+	require.NotSame(t, max1, max2, "max fractional amount should be copied")
```
nddeluca, line 41: Failure messages are very helpful 👍

x/precisebank/types/fractional_balance_test.go
```diff
@@ -0,0 +1,164 @@
...
+	}{
+		{
+			"valid",
+			"kava1gpxd677pp8zr97xvy3pmgk70a9vcpagsakv0tx",
+			sdkmath.NewInt(100),
+			"",
+		},
+		{
+			"valid - uppercase address",
+			"KAVA1GPXD677PP8ZR97XVY3PMGK70A9VCPAGSAKV0TX",
```
nddeluca, line 84: Nice, thnx for included this valid case -- if it ever changes upstream, will be a great unit test to see fail

x/precisebank/types/fractional_balances_test.go
```diff
@@ -0,0 +1,86 @@
...
+		{
+			"invalid - duplicate address",
+			types.FractionalBalances{
+				types.NewFractionalBalance(sdk.AccAddress{1}.String(), sdkmath.NewInt(100)),
+				types.NewFractionalBalance(sdk.AccAddress{1}.String(), sdkmath.NewInt(100)),
+			},
+			"duplicate address kava1qy0xn7za",
+		},
+		{
+			"invalid - duplicate address upper/lower case",
```
nddeluca, line 59: 🚀

nddeluca (approved):

PR #1911, opened 2024-05-13
feat(x/precisebank): Add remainder amount to genesis

drklee3:
## Description
- Validate total fractional amounts in genesis type
- Validate against fractional balances such that `(sum(balances) + remainder) % conversionFactor == 0`
- Add new utility type `SplitBalance` for splitting up full balances into each

## Checklist


x/precisebank/types/fractional_balances_test.go
```diff
@@ -84,3 +85,53 @@ func TestFractionalBalances_Validate(t *testing.T) {
...
+			amount := sdkmath.NewInt(rand.Int63())
+			balances[i] = types.NewFractionalBalance(addr, amount)
+
+			sum = sum.Add(amount)
+		}
+
+		return balances, sum
+	}
+
+	multiBalances, sum := generateRandomFractionalBalances(10)
```
pirtleshell, line 105: can you add a test that checks the behavior for a bonkers number of accounts?
like what happens in the unlikely scenario in which there are enough partial balances to overflow sdkmath.Int?

maybe not worth adding the test case? lol. but maybe we do just to show we thought about it.

as an aside... the max fractional balance for a single account is (1e12 - 1), so it would take more than $10^{65}$ accounts to actually make an overflow happen. 😂 not to mention the fact that a total supply of ~1e9 KAVA means we can't overflow Int on mainnet if we tried. only 1.08e27 akava even exists. we'd need a total supply of 1e47 KAVA to overflow on conversion to akava

ci/env/kava-protonet/genesis.json
```diff
@@ -3006,7 +3006,9 @@
       },
       "in_flight_packets": {}
     },
-    "precisebank": {},
+    "precisebank": {
+      "remainder": "0"
```
pirtleshell, line 3010: i talked with you about this outside of github, but worth noting here:

the remainder should always be zero. if it is not that means we either didn't properly increment/decrement accounts with their owned balance or we increased the total supply. in either case, something is wrong.

we should definitely verify that in == out in memory, but it doesn't need to be chain state.

pirtleshell (changes requested):

x/precisebank/types/fractional_balance.go
```diff
@@ -7,9 +7,20 @@ import (
...
-// maxFractionalAmount is the largest valid value in a FractionalBalance amount.
-// This is for direct internal use so that there are no extra allocations.
-var maxFractionalAmount = sdkmath.NewInt(1_000_000_000_000).SubRaw(1)
+var (
+	// conversionFactor is used to convert the fractional balance to integer
+	// balances.
+	conversionFactor = sdkmath.NewInt(1_000_000_000_000)
+	// maxFractionalAmount is the largest valid value in a FractionalBalance amount.
+	// This is for direct internal use so that there are no extra allocations.
+	maxFractionalAmount = sdkmath.NewInt(1_000_000_000_000).SubRaw(1)
```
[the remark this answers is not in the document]
drklee3: yes much cleaner 😄

x/precisebank/types/fractional_balances_test.go
drklee3: just added a basic test (with an invalid FractionalBalance max int) at least to verify it will panic instead of some overflow

x/precisebank/genesis.go
```diff
@@ -14,24 +15,38 @@ func InitGenesis(
...
 		panic(fmt.Sprintf("failed to validate %s genesis state: %s", types.ModuleName, err))
 	}
 
-	// initialize module account
+	// Initialize module account
 	if moduleAcc := ak.GetModuleAccount(ctx, types.ModuleName); moduleAcc == nil {
 		panic(fmt.Sprintf("%s module account has not been set", types.ModuleName))
 	}
 
-	// TODO:
```
nddeluca, line 28: The validation is good, but I think it can be more simply described like:

```golang
reserveBalance := bk.GetBalance(ctx, moduleAcc.GetAddress(), "ukava").Mul(types.ConversionFactor())

if !reserveBalance.Equal(gs.TotalAmountWIthRemainder() {
  ...
}
```

Tho, I'd also see an argument that this is misleading, and really the comparison should be:

```golang
reserveBalance := bk.GetBalance(ctx, moduleAcc.GetAddress(), "ukava").Mul(types.ConversionFactor()).Sub(gs.Remainder())

if !reserveBalance.Equal(gs.FractionalAmount.SumAmount()) {
  panic("
}
```

as this provides the best representation of what the reserve akava balance is and that is should represent all the fractional balances.  I'm not sure additional abstractions or variables are really needed here.  We could push this into helper methods on the genesis state, but I'd question if that indirection improves readability and will be used more than once when thinking if it is worth the added lines of code.  This is subjective though, I just prefer the simplicity and less code to maintain.

Also, more generally we should strive to work in the most precise units as possible (akava in this case).  In addition, I generally like to avoid Quo unless it's needed (not important here, but multiplication is more performant).

In the case above, using Quo also limits our defensiveness.  If our other validations didn't exist or have a bug and TotalAmountWithRemainder() was not a multiple of C, we'd fail to catch it with the current check since it truncates the precision and doesn't validate the fractional amount is zero.  I think there is a good case to report all panics, logs, etc in this module as akava units for this reason as well.

x/precisebank/types/fractional_balance.go
```diff
@@ -7,9 +7,20 @@ import (
...
-// maxFractionalAmount is the largest valid value in a FractionalBalance amount.
-// This is for direct internal use so that there are no extra allocations.
-var maxFractionalAmount = sdkmath.NewInt(1_000_000_000_000).SubRaw(1)
+var (
+	// conversionFactor is used to convert the fractional balance to integer
+	// balances.
+	conversionFactor = sdkmath.NewInt(1_000_000_000_000)
+	// maxFractionalAmount is the largest valid value in a FractionalBalance amount.
+	// This is for direct internal use so that there are no extra allocations.
+	maxFractionalAmount = conversionFactor.SubRaw(1)
```
nddeluca, line 16: I wonder if we need maxFractionalAmount at all?

It seems simpler and represents how we'd write it on paper if we validate `0 <= amount < C` rather than `0 <= amount <= C - 1`.

x/precisebank/types/split_balance.go
```diff
@@ -0,0 +1,32 @@
+package types
+
+import sdkmath "cosmossdk.io/math"
+
+// SplitBalance represents a full extended balance split into the corresponding
+// integer and fractional parts. IntegerAmount is managed by x/bank while the
+// fractional part is managed by x/precisebank.
+type SplitBalance struct {
```
nddeluca, line 8: My other comment refactors this out, but regardless, I wonder if adding this is premature?  I'm worried it doesn't represent the logic well and it may cause us to see have tunnel vision looking for use-cases and we will end up defining code in terms of this abstraction, and not the simplest or most appropriate abstraction.

Maybe it's best to delay adding this until we have a full working implementation?  The first use case in the genesis validation only uses half the interface, so I probably don't have the full context on what other use cases you foresee.

nddeluca (approved):
Validation & tests are great, I don't see any issues with the behavior.

My comments are not blockers, just points for discussion and possible simplifications.

x/precisebank/testutil/fractional_balances.go
```diff
@@ -0,0 +1,155 @@
...
+
+	return addr
+}
+
+// GenerateEqualFractionalBalances generates count number of FractionalBalances
+// with randomly generated amounts such that the sum of all amounts is a
+// multiple of types.CONVERSION_FACTOR. If a remainder is desired, any single
+// FractionalBalance can be removed from the returned slice and used as the
+// remainder.
+func GenerateEqualFractionalBalances(
```
nddeluca, line 41: Test helpers are great 👍

x/precisebank/genesis.go
drklee3: Used the first one since it seems the most clear. Also used Quo for the error message so its clear that module account should have ukava even though we're comparing akava together

x/precisebank/types/constants.go
```diff
@@ -0,0 +1,11 @@
+package types
+
+// IntegerCoinDenom is the denomination for integer coins that are managed by
+// x/bank. This is the "true" denomination of the coin, and is also used for
+// the reserve to back all fractional coins.
+const IntegerCoinDenom = "ukava"
```
nddeluca, line 6: Filed https://github.com/Kava-Labs/kava/issues/1914, I'm still split if adding the staking and evm keeper params as depedencies is worth it.

ci/env/kava-protonet/genesis.json
pirtleshell: documenting this discussion:
our expectation is for the remainder to be 0 and never change for our use case. however, that _doesn't_ mean that precisebank should not be prepared to handle a world in which there is a non-zero remainder (or one that changes).
the concern of it being 0 in our use of it in x/evm is a higher-level concern than x/precisebank should enforce.
we'll keep remainder in genesis state because it makes for a more pure & correct bank implementation 👍

PR #1912, opened 2024-05-14
feat(x/precisebank): Add keeper methods for store

drklee3:
## Description
- Add store methods to get/set/delete/etc account fractional balances & remainder amount
- Add invariants to ensure stored state is correct

## Checklist

### DIFF
--- a/x/precisebank/keeper/fractional_balance.go
+++ b/x/precisebank/keeper/fractional_balance.go
@@ -79,7 +79,8 @@ func (k *Keeper) IterateFractionalBalances(
 
 // GetAggregateSumFractionalBalances returns the sum of all fractional balances.
 func (k *Keeper) GetAggregateSumFractionalBalances(ctx sdk.Context) sdkmath.Int {
-	var sum sdkmath.Int
+	sum := sdkmath.ZeroInt()
+
 	k.IterateFractionalBalances(ctx, func(_ sdk.AccAddress, amount sdkmath.Int) bool {
 		sum = sum.Add(amount)
 		return false
--- a/x/precisebank/keeper/fractional_balance.go
+++ b/x/precisebank/keeper/fractional_balance.go
@@ -91,8 +91,8 @@ func (k *Keeper) IterateFractionalBalances(
 	}
 }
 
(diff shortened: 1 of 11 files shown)


x/precisebank/keeper/fractional_balance.go
```diff
@@ -0,0 +1,104 @@
...
+	amount sdkmath.Int,
+) {
+	if amount.IsZero() {
+		k.DeleteFractionalBalance(ctx, address)
+		return
+	}
+
+	// Ensure the fractional balance is valid before setting it. Use the
+	// NewFractionalAmountFromInt wrapper to use its Validate() method.
+	if err := types.NewFractionalAmountFromInt(amount).Validate(); err != nil {
```
pirtleshell, line 46: it might make more sense to enforce the caller to pass in the `FractionalAmount` and just call `Validate()` on that?

pirtleshell (approved):
great test coverage & invariant checks 👍 

added some non-blocking comments / discussion

x/precisebank/keeper/fractional_balance.go
```diff
@@ -0,0 +1,104 @@
...
+		panic(fmt.Errorf("amount is invalid: %w", err))
```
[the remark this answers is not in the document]
drklee3: yeah I was thinking about this too, my thought was these store methods are used internally and receiving an invalid value is unrecoverable and incorrect behavior, like if something went wrong with Marshalling, something went real wrong.

But I'm also not entirely sure if that's consistent or ideal since it receiving an invalid value would be a concern of something external from the method, so yup it would be good to define something more clear for us to be consistent on when to panic and when to error 😃

x/precisebank/keeper/fractional_balance.go
drklee3: hmm yeah I first tested with using `FractionalAmount` as the arg, but it seemed a bit strange since we'll only have `sdkmath.Int` after main logic and we'd have to wrap it every time its set.

I think an alternative is refactoring `NewFractionalAmountFromInt.Validate()` to just a regular function instead of creating a new wrapper type?

Created #1917 for discussion on refactor

PR #1916, opened 2024-05-16
feat(x/precisebank): Implement GetBalance

drklee3:
## Description
- Implement GetBalance for extended balances
  - Passes through to `x/bank` for non-extended denoms - diverges from `x/evmutil` behavior which will panic on "akava" calls
- Include empty methods for interface compliance
- Add bank / account keeper mocks for testing
  - Adds mockery config for [mockery package setup](https://vektra.github.io/mockery/latest/migrating_to_packages/)


x/precisebank/keeper/invariants.go
```diff
@@ -121,3 +127,28 @@ func BalancedFractionalTotalInvariant(k Keeper) sdk.Invariant {
 		), broken
 	}
 }
+
+// FractionalDenomNotInBankInvariant checks that the bank does not hold any
+// fractional balances. These assets, e.g. akava, should only exist in the
+// x/precisebank module and shouldn't be "real" assets managed by x/bank.
```
pirtleshell, line 133: can we add a note that the reason it's not allowed isn't because the assets aren't "real", but because the akava are backed by ukava already tracked by x/bank and including them would result in double counting (i guess except in the case where there is a non-zero remainder?)

x/precisebank/keeper/send.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// IsSendEnabledCoins uses the parent x/bank keeper to check the coins provided
+// and returns an ErrSendDisabled if any of the coins are not configured for
+// sending. Returns nil if sending is enabled for all provided coin
+func (k Keeper) IsSendEnabledCoins(ctx sdk.Context, coins ...sdk.Coin) error {
+	return nil
```
pirtleshell, line 11: it's unclear to me if the `nil` returns in this file are placeholders to be implemented or are true no-ops

x/precisebank/keeper/view.go
```diff
@@ -0,0 +1,37 @@
...
+	addr sdk.AccAddress,
+	denom string,
+) sdk.Coin {
+	// Pass through to x/bank for denoms except ExtendedCoinDenom
+	if denom != types.ExtendedCoinDenom {
+		return k.bk.GetBalance(ctx, addr, denom)
+	}
+
+	// x/bank for integer balance - spendable balance only
+	spendableCoins := k.bk.SpendableCoins(ctx, addr)
```
pirtleshell, line 22: i see that x/bank's keeper's GetBalance returns total balance including locked coins.

why not the true balance here? is it to avoid complications with vesting accounts? or do we only return spendable balance b/c there's no concept of non-spendable balance in evmland?

x/precisebank/keeper/view_test.go
```diff
@@ -0,0 +1,109 @@
...
+	"testing"
+
+	sdkmath "cosmossdk.io/math"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/kava-labs/kava/x/precisebank/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestKeeper_GetBalance(t *testing.T) {
+	tests := []struct {
```
pirtleshell, line 13: can you add a test case for zero ukava balance (zero &/or nil) with nonzero akava balance?
(ie, i send a never-used-before account <1e12 akava then query that account's balance)

pirtleshell (approved):
added some notes, questions, & a missing test case

awesome job with the mocks & continuing your usual excellence in test coverage 🧑‍🍳🤌

x/precisebank/keeper/keeper.go
```diff
@@ -3,21 +3,43 @@ package keeper
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	evmtypes "github.com/evmos/ethermint/x/evm/types"
+
+	"github.com/kava-labs/kava/x/precisebank/types"
 )
 
-// TODO: Enforce that Keeper implements the expected keeper interfaces
-// var _ types.BankKeeper = Keeper{}
+// Enforce that Keeper implements the expected keeper interfaces
+var _ evmtypes.BankKeeper = Keeper{}
```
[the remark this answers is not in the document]
drklee3: Yeah I originally was using precisebank expected keeper interface, but realized they're different purposes even though they have mostly the same methods:
- evmtypes.BankKeeper - x/evm requirements so we need to define those methods in x/precisebank, some of these could be removed though since they're not used in x/evm
- types.BankKeeper - x/precisebank requirements that are slightly different like needing x/bank `BlockedAddr` later on, this method doesn't need to be implemented on x/precisebank keeper, only consumed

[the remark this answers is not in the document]
pirtleshell: makes sense! seems fine to keep the dependency as it expresses exactly what is needed.
could also see adding a `EvmBankKeeper` interface but i dont think the duplication is worth it just to remove the dependency.

if we run into DevX problems with precisebank's reliance on x/evm as a dependecy we can reevaluate.

pirtleshell (approved):

x/precisebank/keeper/view.go
drklee3: right hmm this does diverge from `x/bank` behavior which is not great. and yeah I think it was originally used in evmutil to avoid having vested coins show up in balances since that would be confusing to get insufficient balance errors when the queried balance is enough.

PR #1920, opened 2024-05-21
feat(x/precisebank): Implement MintCoins

drklee3:
## Description
- Implement MintCoins method that matches x/bank MintCoins validation behavior

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -327,6 +327,16 @@ test-cli: build
 test-migrate:
 	@$(GO_BIN) test -v -count=1 ./migrate/...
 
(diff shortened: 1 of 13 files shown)


pirtleshell (commented):
fantastic test cases & comments 👍

i added some notes; nothing blocking except to handle the MintCoins error that is dropped when a mint carryover is needed.

aside: it may be overkill, but one category of mints you haven't tested is when there is a non-zero starting remainder. ie. initializing with a module balance & remainder and confirming subsequent mints change/don't change that value accordingly. i feel pretty convinced of the correctness given all the cases you have covered, though.

pirtleshell (approved):

PR #1923, opened 2024-05-24
feat(x/precisebank): Implement SendCoins

drklee3:
## Description
- SendCoins
- SendCoinsFromModuleToAccount
- SendCoinsFromAccountToModule


x/precisebank/keeper/send.go
```diff
@@ -19,23 +29,188 @@ func (k Keeper) SendCoins(
...
+	recipientFractionalBal := k.GetFractionalBalance(ctx, to)
+
+	// Account new fractional balances (NOT YET carried)
+	senderNewFractionalBal := senderFractionalBal.Sub(fractionalAmt)
+	recipientNewFractionalBal := recipientFractionalBal.Add(fractionalAmt)
+
+	// Check if sender needs to borrow and recipient needs to carry
+	senderNeedsBorrow := senderNewFractionalBal.IsNegative()
+	recipientNeedsCarry := recipientNewFractionalBal.GTE(types.ConversionFactor())
+
```
pirtleshell, line 104: i'm struck by how complex the control flow is in this method after this point. i'm trying to wrap my head around the tradeoff being made.

the added checks allow us to prevent two unnecessary transfers (1 ukava from sender to module, then 1 ukava from module to recipient) by allowing the sender to cover the recipient's borrow if it can be covered by their carry: they just send one more integer unit instead of passing the ukava through this module's account.

i wonder about if we could reduce complexity by just always handling the carryover & borrow in the same way, regardless of the fact that 1 out of the 4 possible cases could avoid the two transfers to/from precisebank.

possible cases:
1. sender's balance >= sent, receiver's new balance < conversion factor
2. sender's balance < sent, receiver's new balance < conversion factor
3. sender's balance >= sent, receiver's new balance > conversion factor
4. sender's balance < sent, receiver's new balance > conversion factor

only in case 4 is it possible to cover the receiver's necessary borrow with the sender's carryover; otherwise, some part of the sender -> precisebank -> receiver ukava transfers must occur.

i'm not sure what to make of it. i think my point is that this method just requires a lot of mental overhead to think through, and it feels like there is a refactor or tradeoff asking to be made for the sake of comprehension and readability.

i think the code is correct (and properly tested in all cases!), just wondering if we can spend some time trying to formulate it in an easier to grok way. maybe separating each of the above cases and handling them, with some non-DRY code? or foregoing the optimization in case 4?

i am still thinking it through.... i think the mental overhead is coming from the separated `canDirectTransferCarry` / `!canDirectTransferCarry` conditions.

pirtleshell: this is what i arrived at. i think that making explicit the 4 possible cases in the code allows for a clearer understanding of what's happening here, all while still keeping onlly the minimum number of necessary transfers.

```go
senderNeedsBorrow := senderNewFractionalBal.IsNegative()
recipientNeedsCarry := recipientNewFractionalBal.GTE(types.ConversionFactor())

// handle adjusting balances necessary for fractional balance carries/borrows
switch {
// 1. sender's balance >= sent, receiver's new balance < conversion factor
case !senderNeedsBorrow && !recipientNeedsCarry:
	{
		// happy path! no updates necessary
	}
// 2. sender's balance < sent, receiver's new balance < conversion factor
case senderNeedsBorrow && !recipientNeedsCarry:
	{
		borrowCoin := sdk.NewCoin(types.IntegerCoinDenom, sdk.NewInt(1))
		if err := k.bk.SendCoinsFromAccountToModule(ctx, from, types.ModuleName, sdk.NewCoins(borrowCoin)); err != nil {
			return err
		}
		senderNewFractionalBal = senderNewFractionalBal.Add(types.ConversionFactor())
	}
// 3. sender's balance >= sent, receiver's new balance > conversion factor
case !senderNeedsBorrow && recipientNeedsCarry:
	{
		carryCoin := sdk.NewCoin(types.IntegerCoinDenom, sdk.NewInt(1))
		if err := k.bk.SendCoinsFromModuleToAccount(ctx, types.ModuleName, to, sdk.NewCoins(carryCoin)); err != nil {
			return err
		}
		recipientNewFractionalBal = recipientNewFractionalBal.Sub(types.ConversionFactor())
	}
	// 4. sender's balance < sent, receiver's new balance > conversion factor
case senderNeedsBorrow && recipientNeedsCarry:
	{
		// direct transfer possible
		// add 1 to integer amount
		integerAmt = integerAmt.AddRaw(1)
		senderNewFractionalBal = senderNewFractionalBal.Add(types.ConversionFactor())
		recipientNewFractionalBal = recipientNewFractionalBal.Sub(types.ConversionFactor())
	}
}

if integerAmt.IsPositive() {
	transferCoin := sdk.NewCoin(types.IntegerCoinDenom, integerAmt)
	if err := k.bk.SendCoins(ctx, from, to, sdk.NewCoins(transferCoin)); err != nil {
		return err
	}
}
```

i'm not tied to this, but i think it's worth trying to make this (very important!) logic easier to follow. but also be sure to add back in all your great comments 😆

nddeluca: Valid points @pirtleshell, but I'm not sure the case statement helps here.  I think @drklee3 portrays the logic well, but we can make a couple simplifications by using mod instead of Add/Sub ConversionFactor, and remove the named direct transfer variable to flatten the logic.  I also removed the up front balance check to reduce gas costs for successful transfers.

Thoughts @drklee3 @pirtleshell ?

```golang
// sendExtendedCoins transfers amt extended coins from a sending account to a
// receiving account. An error is returned upon failure. This function is
// called by SendCoins() and should not be called directly.
func (k Keeper) sendExtendedCoins(
	ctx sdk.Context,
	from, to sdk.AccAddress,
	amt sdkmath.Int,
) error {
	integerSendAmt := amt.Quo(types.ConversionFactor())

	senderFractionalBal := k.GetFractionalBalance(ctx, from)
	senderNewFractionalBal := senderFractionalBal.Sub(amt).Mod(types.ConversionFactor())

	recipientFractionalBal := k.GetFractionalBalance(ctx, to)
	recipientNewFractionalBal := recipientFractionalBal.Add(amt).Mod(types.ConversionFactor())

	// we went past zero and wrapped around, must decrease sender balance by 1
	borrowRequired := senderNewFractionalBal.GT(senderFractionalBal)
	// we went past ConversionFactor-1 and wrapped around, must increase receipient balance by 1
	carryRequired := recipientNewFractionalBal.LT(recipientFractionalBal)

	// we can skip the reserves and increase transfer amount by 1
	if carryRequired && borrowRequired {
		integerSendAmt = integerSendAmt.Add(sdkmath.OneInt())
	}

	// transfer integer amount from sender to receipient
	// we do the the larger transfer first to fail fast
	if integerSendAmt.IsPositive() {
		transferCoin := sdk.NewCoin(types.IntegerCoinDenom, integerSendAmt)
		if err := k.bk.SendCoins(ctx, from, to, sdk.NewCoins(transferCoin)); err != nil {
			// We assume most transactions will succeed, so skip checking the balance ahead of time
			// to save gas cost.  The bank module will check spendable balance for us.
			return k.wrapError(ctx, from, amt, err)
		}
	}

	// we didn't borrow from sender during transfer, so take 1 from sender and increase reserves
	if borrowRequired && !carryRequired {
		borrowCoin := sdk.NewCoin(types.IntegerCoinDenom, sdkmath.OneInt())
		if err := k.bk.SendCoinsFromAccountToModule(ctx, from, types.ModuleName, sdk.NewCoins(borrowCoin)); err != nil {
			// We assume most transactions will succeed, so skip checking the balance ahead of time
			// to save gas cost.  The bank module will check spendable balance for us.
			return k.wrapError(ctx, from, amt, err)
		}
	}

	// we didn't carry to recipient during transfer, so take 1 from reserve and decrease reserve
	if carryRequired && !borrowRequired {
		carryCoin := sdk.NewCoin(types.IntegerCoinDenom, sdkmath.OneInt())
		if err := k.bk.SendCoinsFromModuleToAccount(ctx, types.ModuleName, to, sdk.NewCoins(carryCoin)); err != nil {
			// This will only error if state is invalid, we should panic as caller can not fix
			panic(err)
		}
	}

	// update fractional balances
	k.SetFractionalBalance(ctx, from, senderNewFractionalBal)
	k.SetFractionalBalance(ctx, to, recipientNewFractionalBal)

	return nil
}

func (k Keeper) wrapError(ctx sdk.Context, addr sdk.AccAddress, amt sdk.Int, err error) error {
	if !errors.Is(err, sdkerrors.ErrInsufficientFunds) {
		return err
	}

	// Check balance is sufficient
	bal := k.GetBalance(ctx, addr, types.ExtendedCoinDenom)
	coin := sdk.NewCoin(types.ExtendedCoinDenom, amt)

	// TODO: This checks spendable coins and returns error with spendable
	// coins, not full balance. If GetBalance() is modified to return the
	// full, including locked, balance then this should be updated to deduct
	// locked coins.

	// Use sdk.NewCoins() so that it removes empty balances - ie. prints
	// empty string if balance is 0. This is to match x/bank behavior.
	spendable := sdk.NewCoins(bal)

	return errorsmod.Wrapf(
		sdkerrors.ErrInsufficientFunds,
		"spendable balance %s is smaller than %s",
		spendable, coin,
	)
}
```

x/precisebank/keeper/send.go
```diff
@@ -1,16 +1,26 @@
...
 )
 
 // IsSendEnabledCoins uses the parent x/bank keeper to check the coins provided
 // and returns an ErrSendDisabled if any of the coins are not configured for
 // sending. Returns nil if sending is enabled for all provided coin
 func (k Keeper) IsSendEnabledCoins(ctx sdk.Context, coins ...sdk.Coin) error {
-	panic("unimplemented")
+	// TODO: This does not actually seem to be used by x/evm, so it should be
+	// removed from the expected_interface in x/evm.
+	return k.bk.IsSendEnabledCoins(ctx, coins...)
```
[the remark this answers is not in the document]
drklee3: uhh it is now! 😄 

https://app.shortcut.com/kava-labs/story/13580/remove-unused-methods-in-x-evm-expected-interface-bankkeeper

x/precisebank/keeper/send.go
pirtleshell: nice, @nddeluca 👍 

that is closer to what i had before i pulled the carry & borrow transfers up into the switch-case. i prefer your separation of pure and statefulness. particularly i think this is great for comprehension:
```go
transferBorrowToReserve := borrowRequired && !carryRequired
transferCarryToRecipient := carryRequired && !borrowRequired
if borrowRequired && carryRequired {
	integerSendAmt = integerSendAmt.Add(sdkmath.OneInt())
}
```
also some great pure function extractions ✨ 

---

when i was thinking through this i was under the impression we needed to check for balance at the start but i'm trying to remember why.. but, if they don't have sufficient balance one of these transfers will fail:
1. integer transfer of ukava to the receiver
2. borrow transfer of 1 ukava to the module account

seems correct to optimize gas costs for successful txs by relying on those to error if there's insufficient balance 👍

drklee3: all great points 😃 just confirmed that even if its only a fractional send and fractional balance is not sufficient (e.g 0.02, send 0.03) that still incurs a borrow and thus the error - so a preemptive balance check is not necessary. Will also include those 4 cases in a doc comment since that's also useful to understand

x/precisebank/keeper/send.go
```diff
@@ -19,23 +32,291 @@ func (k Keeper) SendCoins(
...
+		transferCoin := sdk.NewCoin(types.IntegerCoinDenom, integerAmt)
+		if err := k.bk.SendCoins(ctx, from, to, sdk.NewCoins(transferCoin)); err != nil {
+			return k.wrapError(ctx, from, amt, err)
+		}
+	}
+
+	// Case #2: Sender borrow, NO recipient carry
+	// Sender borrows by transferring 1 integer amount to reserve to account for
+	// lack of fractional balance.
+	if senderNeedsBorrow && !recipientNeedsCarry {
```
drklee3, line 138: Using the bools directly instead of assigning a new `transferBorrowToReserve` - I think more clear this way

x/precisebank/keeper/send.go
```diff
@@ -19,23 +32,291 @@ func (k Keeper) SendCoins(
...
+	borrowRequired := newFractionalBalance.IsNegative()
+
+	if borrowRequired {
+		// Borrowing 1 integer equivalent amount of fractional coins. We need to
+		// add 1 integer equivalent amount to the fractional balance otherwise
+		// the new fractional balance will be negative.
+		newFractionalBalance = newFractionalBalance.Add(types.ConversionFactor())
+	}
+
+	return newFractionalBalance, borrowRequired
```
drklee3, line 204: Bit different from suggested @nddeluca though I think its clearer this way, by comparing with negative / conversionFactor amounts instead of previous

x/precisebank/keeper/send.go
```diff
@@ -19,23 +32,293 @@ func (k Keeper) SendCoins(
...
+			to, // recipient carrying
+			sdk.NewCoins(carryCoin),
+		); err != nil {
+			// Panic instead of returning error, as this will only error
+			// with invalid state or logic. Reserve should always have
+			// sufficient balance to carry fractional coins.
+			panic(fmt.Errorf("failed to carry fractional coins to %s: %w", to, err))
+		}
+	}
+
```
pirtleshell, line 168: @drklee3 can you add a comment here about how Case #4 is a no-op because the fractional balance sent if fully covered by the user's current fractional balance?

pirtleshell (approved):

Makefile
```diff
@@ -334,6 +334,7 @@ endif
 
 test-fuzz:
 	@$(GO_BIN) test $(FUZZLDFLAGS) -run NOTAREALTEST -v -fuzztime 10s -fuzz=FuzzMintCoins ./x/precisebank/keeper
+	@$(GO_BIN) test $(FUZZLDFLAGS) -run NOTAREALTEST -v -fuzztime 10s -fuzz=FuzzSendCoins ./x/precisebank/keeper
```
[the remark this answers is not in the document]
drklee3: Without the dummy test name, it'll run all the tests in the package each time 🫠 With it, it'll run 0 tests and only the specified fuzz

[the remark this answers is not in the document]
pirtleshell: hah interesting design choice, golang. 😂 
thanks for the explanation

x/precisebank/keeper/send.go
```diff
@@ -19,23 +32,298 @@ func (k Keeper) SendCoins(
...
+	return k.SendCoins(ctx, senderAddr, recipientAddr, amt)
+}
+
+// updateInsufficientFundsError returns a modified ErrInsufficientFunds with
+// extended coin amounts if the error is due to insufficient funds. Otherwise,
+// it returns the original error. This is used since x/bank transfers will
+// return errors with integer coins, but we want the more accurate error that
+// contains the full extended coin balance and send amounts.
+func (k Keeper) updateInsufficientFundsError(
```
pirtleshell, line 301: nit: can we name this `wrapInsufficientFundsError` for more consistent naming with other error wrappers/updates?

drklee3: hmm I named this to not be wrap since it's not actually wrapping but returning a new error with the correct full balance 🤔 thought it would be a bit misleading to be called wrap but can rename if its still clear enough

pirtleshell: hm yeah.. haha no strong preference. keep `update` 😆

nddeluca (approved):
great work 🚀

PR #1932, opened 2024-06-04
fix(x/precisebank): Ensure exact reserve balance on integer carry when minting

drklee3:
## Description
- Fix reserve minting an extra coin when the recipient module both carries fractional over to integer balance AND remainder is insufficient.
- Adjusts fractional carry to simply send from reserve, instead of doing an additional mint.
- Add invariant to ensure reserve matches exactly with fractional balances + remainder, failing on both insufficient and excess funds.

### DIFF
--- a/x/precisebank/keeper/invariants_integration_test.go
+++ b/x/precisebank/keeper/invariants_integration_test.go
@@ -107,7 +107,7 @@ func (suite *invariantsIntegrationTestSuite) TestReserveBackingFractionalInvaria
 
 			tt.setupFn(suite.Ctx, suite.Keeper)
 
-			invariantFn := keeper.ReserveBackingFractionalInvariant(suite.Keeper)
+			invariantFn := keeper.ReserveBacksFractionsInvariant(suite.Keeper)
 			msg, broken := invariantFn(suite.Ctx)
 
 			if tt.wantBroken {
--- a/x/precisebank/keeper/invariants_integration_test.go
+++ b/x/precisebank/keeper/invariants_integration_test.go
@@ -1,0 +1,121 @@
+package keeper_test
+
+import (
+	"testing"
+
+	sdkmath "cosmossdk.io/math"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/kava-labs/kava/x/precisebank/keeper"
+	"github.com/kava-labs/kava/x/precisebank/testutil"
+	"github.com/kava-labs/kava/x/precisebank/types"
+	"github.com/stretchr/testify/suite"
+)
+
+type invariantsIntegrationTestSuite struct {
+	testutil.Suite
+}
+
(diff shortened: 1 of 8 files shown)


x/precisebank/keeper/mint.go
```diff
@@ -89,8 +93,24 @@ func (k Keeper) mintExtendedCoin(
 	// x + y < (2 * ConversionFactor) - 2
 	// x + y < 1 integer amount + fractional amount
 	if newFractionalBalance.GTE(types.ConversionFactor()) {
```
nddeluca, line 95: The logic looks great, however we can optimize the carry and negative remainder case to instead add to the integer amount being minted.  This avoids the extra reserve -> account transfer and the reserve mint to make up the balance change and saves quite a bit of gas.

1) No carry, zero/positive remainder -> do nothing extra
2) No carry, negative remainder -> mint to reserve
3) Carry, zero/positive remainder -> transfer 1 from reserve to account
4) Carry, negative remainder -> add 1 to integer acct mint amount

x/precisebank/keeper/invariants_integration_test.go
```diff
@@ -0,0 +1,121 @@
...
+		wantMsg    string
+	}{
+		{
+			"valid - empty state",
+			func(_ sdk.Context, _ keeper.Keeper) {},
+			false,
+			"",
+		},
+		{
+			"valid - fractional balances, no remainder",
```
pirtleshell, line 46: i know it won't happen on mainnet because all akava will be completely backed by ukava, but is no fractional balances & a nonzero remainder ever a valid state we need to test?

pirtleshell (approved):
great bug catch by @nddeluca & fix by you @drklee3 ! 👍 

added some nits and a question but LGTM

x/precisebank/keeper/invariants_integration_test.go
drklee3: Added a test case for it - should be broken state since no fractional balances should be 0 remainder

x/precisebank/keeper/mint.go
```diff
@@ -115,20 +168,28 @@ func (k Keeper) mintExtendedCoin(
...
-	if prevRemainder.LT(fractionalMintAmount) {
-		// Need additional 1 integer coin in reserve to back minted fractional
+	// Mint an additional reserve integer coin if remainder is insufficient.
+	// The remainder is the amount of fractional coins that can be minted and
+	// still be fully backed by reserve. If the remainder is less than the
+	// minted fractional amount, then the reserve needs to be increased to
+	// back the additional fractional amount.
+	// Optimization: This is only done when the integer amount does NOT carry,
+	// as a direct account mint is done instead of integer carry transfer +
+	// insufficient remainder reserve mint.
```
pirtleshell, line 179: appreciate this detailed explanation 👍

pirtleshell (approved):

pirtleshell (approved):

PR #1935, opened 2024-06-09
chore: make function comments match function names

todaymoon:
## Description

make function comments match function names

## Checklist


pirtleshell (approved):
thanks for the contribution!

PR #1937, opened 2024-06-11
bump internal-testnet VERSION and genesis file for wbtc config

downing034:
## Description
- update internal-testnet commit hash so we can deploy newer code there. We're about 10 months behind
- add missing keys to internal-testnet genesis file, which were added in the last 10months to production.

## Checklist

# Discussion Item
Can this be included in the release process. The need to keep internal-testnet on a pinned version to avoid breaking changes, but it would be great to include updating lower environments when we're doing a release (since this assumes the code is stable).


ci/env/kava-internal-testnet/KAVA.VERSION
```diff
@@ -1 +1 @@
-a967d2fdda299ec8e1e3b99fb55bd06ecfdb0469
+7df1eef8cc977816a2fdff2fd208aec87fbdb699
```
downing034, line 1: so internal testnet will update to more recent code

ci/env/kava-internal-testnet/genesis.json
```diff
@@ -22,6 +22,8 @@
   },
   "app_hash": "",
   "app_state": {
+    "06-solomachine": null,
```
downing034, line 25: all of the keys added are keys that were missing in the internal-testnet config from a few launches back, but required in order for the genesis file to pass the validation builds

PR #1943, opened 2024-06-13
Update genesis.json

downing034:
## Checklist

### DIFF
--- a/ci/env/kava-internal-testnet/genesis.json
+++ b/ci/env/kava-internal-testnet/genesis.json
@@ -530,6 +530,10 @@
               "amount": "1000000000"
             },
             {
+              "denom": "erc20/bitgo/wbtc",
+              "amount": "200000000"
+            },
+            {
               "denom": "erc20/multichain/usdc",
               "amount": "1000000000000000000"
             },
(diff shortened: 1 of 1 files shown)


pirtleshell:
@downing034 you're probably aware, but you'll need to update the `KAVA_VERSION` to get this deployed. options are a follow-up PR targeting the merged commit from this PR, or updating in this PR to `master` with the understanding all future merges will trigger re-deploys (and will wipe state)

ci/env/kava-internal-testnet/genesis.json
```diff
@@ -618,6 +622,10 @@
             {
               "denom": "xrpb",
               "amount": "103000000000"
+            },
+            {
+              "denom": "erc20/bitgo/wbtc",
```
pirtleshell, line 627: coins must be sorted alphabetically by denom

you can validate the genesis locally with
```
make install
kava validate-genesis ci/env/kava-internal-testnet/genesis.json
```

downing034: ^Thanks for the feedback. @pirtleshell Do we no longer have the "evmutils" module in the genesis.json? We used to need the contract addresses add there as well (our frontend confluence documentation around this still references it).

PR #1957, opened 2024-06-28
feat(x/precisebank): Add SpendableCoin method

drklee3:
## Description
- Change `GetBalance()` to return full balances instead of spendable
- Add `SpendableCoin()` method with support of akava


x/precisebank/keeper/view_integration_test.go
```diff
@@ -0,0 +1,122 @@
...
+			),
+		},
+		{
+			"non-extended denom - ukava returns ukava",
+			types.IntegerCoinDenom,
+			sdk.NewCoins(sdk.NewCoin(types.IntegerCoinDenom, sdk.NewInt(1000))),
+			sdk.ZeroInt(),
+			sdk.NewCoins(sdk.NewCoin(types.IntegerCoinDenom, sdk.NewInt(10))),
+			sdk.NewCoin(types.IntegerCoinDenom, sdk.NewInt(990)),
+		},
```
pirtleshell, line 72: can we add a similar test case, but with non-zero fractional balance?
i could see it preventing a future regression

pirtleshell (approved):
LGTM 👍

PR #1958, opened 2024-07-02
feat(x/precisebank): Display 0 reserve balance to module consumers

drklee3:
Module reserve represents fractional balances, so it should be hidden to consumers to not have a misleading total balance that doubles the fractional balances.

This modifies `GetBalance()` and `SpendableCoin()` to always return zero coins when fetching the reserve address balance.

### DIFF
--- a/x/precisebank/keeper/view_integration_test.go
+++ b/x/precisebank/keeper/view_integration_test.go
@@ -140,7 +140,14 @@ func (suite *viewIntegrationTestSuite) TestKeeper_HiddenReserve() {
 	// Make the reserve hold a non-zero balance
 	// Mint fractional coins to an account, which should cause a mint of 1
 	// integer coin to the reserve to back it.
-	suite.MintToAccount(addr1, sdk.NewCoins(sdk.NewCoin(types.ExtendedCoinDenom, sdk.NewInt(1000))))
+	unrelatedCoin := sdk.NewCoin("unrelated", sdk.NewInt(1000))
+	suite.MintToAccount(
+		addr1,
+		sdk.NewCoins(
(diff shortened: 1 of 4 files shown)


x/precisebank/keeper/view_integration_test.go
```diff
@@ -129,3 +129,30 @@ func (suite *viewIntegrationTestSuite) TestKeeper_SpendableCoin() {
...
+	// Reserve balances should not be shown to consumers of x/precisebank, as it
+	// represents the fractional balances of accounts.
+
+	moduleAddr := authtypes.NewModuleAddress(types.ModuleName)
+	addr1 := sdk.AccAddress{1}
+
+	// Make the reserve hold a non-zero balance
+	// Mint fractional coins to an account, which should cause a mint of 1
+	// integer coin to the reserve to back it.
+	suite.MintToAccount(addr1, sdk.NewCoins(sdk.NewCoin(types.ExtendedCoinDenom, sdk.NewInt(1000))))
```
pirtleshell, line 143: can you mint some of another denom to this address & write a test confirming that that denom is unaffected?

on mainnet, we never expect any other denom, but feels worthwhile to verify that the overridden zero balance is only for akava

pirtleshell: correction / update: we _do_ expect another denom. it will have ukava if the sum of all the different fractional balances is greater than the conversion factor.

so the ask: can you also mint ukava to the module in this test & verify that it correctly responds with a non-zero balance when queried?

pirtleshell (approved):

nddeluca (approved):
Nice work, great use of comments 👍

PR #1960, opened 2024-07-03
feat: Use x/precisebank for x/evm keeper

drklee3:
## Description
- Use x/precisebank for x/evm keeper, replacing x/evmutil
- Remove legacy x/evmutil `EvmBankKeeper`


karzak (approved):
Had a look through the resulting `evmutil` and `precisebank` modules - code and separation of invariants looks good. Created cards to remind us to update the `evmutil` spec and create one in `precisebank` before we release.

PR #1961, opened 2024-07-03
refactor(x/precisebank): Replace FractionalAmount wrapper with func

### LINKED ISSUE #1917: Identify alternative to x/precisebank NewFractionalAmountFromInt wrapper type for validation
Reused logic for validating fractional amounts (non-empty, range, etc)

Currently implemented as a wrapper type with a Validate() method but we could also just use a function eg `ValidateFractionalAmount`

drklee3:
## Description
- Remove FractionalAmount wrapper type
- Replace FractionalAmount.Validate() with a regular validate func

Resolves #1917


x/precisebank/types/fractional_balance.go
```diff
@@ -39,5 +41,24 @@ func (fb FractionalBalance) Validate() error {
...
+func ValidateFractionalAmount(amt sdkmath.Int) error {
+	if amt.IsNil() {
+		return fmt.Errorf("nil amount")
+	}
+
+	if !amt.IsPositive() {
+		return fmt.Errorf("non-positive amount %v", amt)
+	}
+
+	if amt.GT(maxFractionalAmount) {
```
karzak, line 59: Should this be `GTE` - ie the fraction should always be less than 1ukava, or is there a case where it's exactly 1ukava

karzak (approved):
One question, refactor looks fine

x/precisebank/types/fractional_balance.go
drklee3: maxFractionalAmount would be (1_000_000_000_000 - 1)
though removed the use of `maxFractionalAmount` now and replacing it with `conversionFactor` directly now, which also matches this change

PR #1963, opened 2024-07-09
feat: Migrate fractional balance state from evmutil to precisebank

drklee3:
## Description
- Set fractional balance in x/precisebank from all balances in x/evmutil
- Delete corresponding state in x/evmutil

## Notes
This is treated as a migration for x/evmutil, not x/precisebank, as it is the store of evmutil that will be removed (breaking change)

- Next PR will be to remove the accounts from x/evmutil, genesis, keeper methods, types etc.
- This PR includes a copy of the necessary types/keys to function after the accounts are removed.
- Moving reserve balances to precisebank module is not a store migration, so it will be added to the upgrade handler.


x/evmutil/migrations/v3/store.go
```diff
@@ -0,0 +1,38 @@
...
+	for ; iterator.Valid(); iterator.Next() {
+		// TODO: Change types package to v3types once moved to v3/types and
+		// removed from evmutil/types
+		var acc types.Account
+		if err := cdc.Unmarshal(iterator.Value(), &acc); err != nil {
+			panic(err)
+		}
+
+		// Panics if the balance is invalid.
+		precisebankkeeper.SetFractionalBalance(ctx, acc.Address, acc.Balance)
```
drklee3, line 31: Alternatively this could be done in the upgrade handler, and leave this migration to only deletion:

Currently in #1966 : module migrations (including this fractional balances transfer), then reserve transfers.

Alternative: fractional balance and reserve transfers in upgrade handler, then run module migrations (delete from evmutil) after

pirtleshell: thinking through this more (& w/ some discussion with @nddeluca), there shouldn't be a migrator at all.
there shouldn't be a consensus version bump in evmutil. we can think of the balance migration as adding a new table (precisebank balances), setting values in that table, and then deleting rows on the evmutil balances table. there is no db migration necessary.

we should restrict migrators to only that particular module's store migrations (like renaming a key) with no dependency on another module's state.

with that said, we probably want to close this pr. the migration should be made entirely in the upgrade handler as follows (credit to @nddeluca):
```
1) Run migrations in upgrade handler
  1a) This does not require any evmutil migrations (module state format is not changing)
  1b) This runs InitGenesis for precisebank
2) Explicitly iterate evmutil state and set precisebank state
3) Explicitly delete all evmutil state for balances
4) Remove evmutil keeper methods/code later in patch, etc
```

it will require exposing operations for deleting account balance state on the evmutil keeper.

pirtleshell (changes requested):

PR #1964, opened 2024-07-09
refactor: Remove legacy fractional balance types

drklee3:
## Description
- Remove proto type for x/evmutil fractional balances
- Remove relevant methods for x/evmutil fractional balances

Includes some partially redundant changes from #1960 to as the bank keeper was also removed due to it depending on the removed methods


drklee3:
Required still for upgrade handler, to be revisited after upgrade

PR #1966, opened 2024-07-16
feat: Add upgrade handler, fractional balances & reserve transfer

drklee3:
## Description

- Add upgrade handler
- Migrates from x/evmutil to x/precisebank:
  - Fractional balances
  - Reserve funds
    - Mints or burns coins to ensure fractional balances are fully backed.
- Initialize remainder if necessary to ensure valid state.
- Add gRPC query service to x/precisebank for validation of total fractional balances - Same change in https://github.com/Kava-Labs/kava/pull/1970
- E2E test with fixed kvtool -> https://github.com/Kava-Labs/kvtool/pull/203

### DIFF
--- a/app/upgrades.go
+++ b/app/upgrades.go
@@ -3,13 +3,14 @@ package app
 import (
 	"fmt"
 
-	"github.com/cometbft/cometbft/libs/log"
+	sdkmath "cosmossdk.io/math"
 
 	storetypes "github.com/cosmos/cosmos-sdk/store/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/types/module"
 	bankkeeper "github.com/cosmos/cosmos-sdk/x/bank/keeper"
 	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
+	evmutilkeeper "github.com/kava-labs/kava/x/evmutil/keeper"
 	evmutiltypes "github.com/kava-labs/kava/x/evmutil/types"
 	precisebankkeeper "github.com/kava-labs/kava/x/precisebank/keeper"
 	precisebanktypes "github.com/kava-labs/kava/x/precisebank/types"
@@ -61,7 +62,8 @@ func upgradeHandler(
 		plan upgradetypes.Plan,
 		fromVM module.VersionMap,
 	) (module.VersionMap, error) {
-		app.Logger().Info(fmt.Sprintf("running %s upgrade handler", name))
+		logger := app.Logger()
+		logger.Info(fmt.Sprintf("running %s upgrade handler", name))
 
 		// Run migrations for all modules and return new consensus version map.
 		versionMap, err := app.mm.RunMigrations(ctx, app.configurator, fromVM)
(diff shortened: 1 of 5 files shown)


app/upgrades.go
```diff
@@ -1,3 +1,274 @@
...
+	remainder := precisebanktypes.ConversionFactor().
+		Sub(extraFractionalAmount).
+		// Mod conversion factor to ensure remainder is valid.
+		// If extraFractionalAmount is a multiple of conversion factor, the
+		// remainder is 0.
+		Mod(precisebanktypes.ConversionFactor())
+
+	// Panics if the remainder is invalid. In a correct chain state and only
+	// mint/burns due to transfers, this would be 0.
+	precisebankKeeper.SetRemainderAmount(ctx, remainder)
```
pirtleshell, line 200: because we know this should be 0 on mainnet & testnet, can we add an assertion here to ensure that?

app/upgrades.go
```diff
@@ -1,3 +1,274 @@
...
+	}
+
+	logger.Info(fmt.Sprintf("transferred reserve balance: %s", reserveBalance))
+
+	// Ensure x/precisebank reserve fully backs all fractional balances.
+	totalFractionalBalances := precisebankKeeper.GetTotalSumFractionalBalances(ctx)
+
+	// Does NOT ensure state is correct, total fractional balances should be a
+	// multiple of conversion factor but is not guaranteed due to the remainder.
+	// Remainder initialization is handled by InitializeRemainder.
```
pirtleshell, line 237: similar to the above, because our upgrade handlers are specific to mainnet and testnet and we know there should be no remainder, maybe we should be making the stricter assertion here?

curious what @nddeluca thinks

tests/e2e/e2e_upgrade_handler_test.go
```diff
@@ -1,18 +1,84 @@
...
+		afterEvmutilBalRes.Balance,
+	)
+
+	afterPrecisebankBalRes, err := grpcClient.Query.Bank.Balance(afterUpgradeCtx, &banktypes.QueryBalanceRequest{
+		Address: precisebankAddr,
+		Denom:   precisebanktypes.IntegerCoinDenom,
+	})
+	suite.Require().NoError(err)
+	suite.Require().NotNil(afterPrecisebankBalRes.Balance)
+	suite.Require().True(afterPrecisebankBalRes.Balance.Amount.IsPositive())
```
pirtleshell, line 68: for completeness can we check the balance amount exactly in this test?

tests/e2e/e2e_upgrade_handler_test.go
```diff
@@ -1,18 +1,84 @@
...
+		"should be empty before upgrade",
+	)
+
+	suite.T().Logf("x/evmutil balances before upgrade: %s", previousEvmutilBalRes.Balance)
+	suite.T().Logf("x/precisebank balances before upgrade: %s", previousPrecisebankBalRes.Balance)
+
+	// -----------------------------
+	// After upgrade
+	// - Check reserve balance transfer
+	// - Check reserve fully backs fractional amounts
```
pirtleshell, line 49: is there a way to confirm the remainder was set properly without a new query? maybe checking total fractional balances == reserve (auto assumes there is a 0 remainder)

could be worth considering adding a query for remainder

app/upgrades.go
drklee3: I'm not 100% certain yet if it would be 0 if the evmutil bugs affected this - may need to double check how the state is in mirrornet or at least with a mainnet genesis export possibly

drklee3: [Added ticket to track](https://app.shortcut.com/kava-labs/story/13831/verify-mainnet-state-for-fractional-balance-migration)

app/upgrades.go
pirtleshell: thanks for creating the tracking ticket. we will re-assess after the mainner mirror test 👍

pirtleshell (approved):
looks great! 🙌 thanks for the updates.
@drklee3 can you rebase onto current release branch before merge? i did locally and confirmed that these are the only files changed 👍
```
❯ git diff --name-only release/v0.27.x
Makefile
app/upgrades.go
app/upgrades_test.go
tests/e2e/.env
tests/e2e/e2e_upgrade_handler_test.go
tests/e2e/kvtool
```

PR #1970, opened 2024-07-17
feat(x/precisebank): Add query service

drklee3:
## Description
- Add query service to precisebank, mostly for e2e test purposes in https://github.com/Kava-Labs/kava/pull/1966
- Fix client grpc codec


proto/kava/precisebank/v1/query.proto
```diff
@@ -0,0 +1,32 @@
...
+
+import "cosmos_proto/cosmos.proto";
+import "gogoproto/gogo.proto";
+import "google/api/annotations.proto";
+
+option go_package = "github.com/kava-labs/kava/x/precisebank/types";
+option (gogoproto.goproto_getters_all) = false;
+
+// Query defines the gRPC querier service for precisebank module
+service Query {
```
pirtleshell, line 12: may be worth adding an endpoint for fractional balance by address as well? i could see it being useful in the event of needing to debug a tx or audit an address.

not a blocking addition

x/precisebank/keeper/grpc_query_test.go
```diff
@@ -0,0 +1,85 @@
...
+	types.RegisterQueryServer(queryHelper, keeper.NewQueryServerImpl(suite.Keeper))
+
+	suite.queryClient = types.NewQueryClient(queryHelper)
+}
+
+func TestGrpcQueryTestSuite(t *testing.T) {
+	suite.Run(t, new(grpcQueryTestSuite))
+}
+
+func (suite *grpcQueryTestSuite) TestQueryDelegatedBalance() {
```
pirtleshell, line 36:
```suggestion
func (suite *grpcQueryTestSuite) TestQueryTotalFractionalBalance() {
```

x/precisebank/keeper/grpc_query.go
```diff
@@ -0,0 +1,34 @@
...
+// This is mostly for external verification of the total fractional balances,
+// being a multiple of the conversion factor and backed by the reserve.
+func (s queryServer) TotalFractionalBalances(
+	goCtx context.Context,
+	req *types.QueryTotalFractionalBalancesRequest,
+) (*types.QueryTotalFractionalBalancesResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	return &types.QueryTotalFractionalBalancesResponse{
+		Total: s.keeper.GetTotalSumFractionalBalances(ctx),
```
pirtleshell, line 32: my guess is this is a pretty expensive operation on mainnet. we may consider blocking the query at the routing level in our production apis.

proto/kava/precisebank/v1/query.proto
```diff
@@ -0,0 +1,32 @@
...
+
+// QueryTotalFractionalBalancesResponse defines the response type for Query/TotalFractionalBalances method.
+message QueryTotalFractionalBalancesResponse {
+  // total is the total sum of all fractional balances managed by the precisebank
+  // module.
+  string total = 1 [
+    (cosmos_proto.scalar) = "cosmos.Int",
+    (gogoproto.customtype) = "cosmossdk.io/math.Int",
+    (gogoproto.nullable) = false
+  ];
```
pirtleshell, line 31: thoughts on returning a coin here instead of just an integer?

pirtleshell (approved):

proto/kava/precisebank/v1/query.proto
drklee3: yup seems useful and can reduce the headache if we do try debugging it without the query 😄 will add fractional balance & remainder query as well in a followup pr

PR #1971, opened 2024-07-19
feat(x/precisebank): Add gRPC query for remainder and account fractional balance

drklee3:
## Description
- Add Remainder and FractionalBalance gRPC query methods


pirtleshell (approved):
excellent additions 👍

PR #1974, opened 2024-07-23
docs(x/precisebank): Add explanations for logic

drklee3:
## Description
- Add explanation details for math logic

### DIFF
--- a/x/precisebank/README.md
+++ b/x/precisebank/README.md
@@ -131,7 +131,6 @@ supply of fractional units.
 
 $$b(R) \cdot C = \sum_{n \in \mathcal{A}}{f(n)} + r$$
 
-
 $$b'(R) \cdot C = \sum_{n \in \mathcal{A}}{f'(n)} + r'$$
 
 With these two formulas, we can determine the new remainder and reserve by using
@@ -172,7 +171,7 @@ $$q * C = r' - r$$
 
 for some integer $q$.
 
-With our known range of $r$:
+With our known range of $r$ and $r'$:
 
 $$0 \leq r' < C, 0 \leq r < C$$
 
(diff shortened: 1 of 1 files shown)


x/precisebank/README.md
```diff
@@ -127,50 +126,92 @@ $$a'(1) = a(1) - a$$
...
+
 $$f'(1) = f(1) - a \mod{C}$$
 
 $$f'(1)\bmod{C}= f(1)\bmod{C} - a \bmod{C} \mod{C}$$
 
 $$f'(1) = f(1) - a \bmod{C} \mod{C}$$
 
+This results in two cases for $f'(1)$:
+
 $$f'(1) = \begin{cases} f(1) - a\bmod{C} & f'(1) \leq f(1) \\
```
nddeluca, line 210: It's probably more clear and formal to write

$$f'(1) = \begin{cases} f(1) - a\bmod{C} & 0 \leq f(1) -a\bmod{C} \\
f(1) - a\bmod{C} + C & 0 > f(1) - a\bmod{C} \end{cases}$$

at this step as we shouldn't include the result as the definition. This was lazy typing on my part.

Checking $f'(1)$ and $f(1)$ in later steps is an optimization since the assumption is $f'(1)$ and $f(1)$ are already known and noticing

$$f'(1) \leq f(1) \Longleftrightarrow  f'(1) = f(1) - a\bmod{C} $$

$$f'(1) > f(1) \Longleftrightarrow  f'(1) = f(1) - a\bmod{C} + C$$

nddeluca (approved):
Very well explained, thanks for taking the time to clarify steps.

I made one comment on a place we could improve, but is good to merge regardless.

nddeluca (approved):

PR #1975, opened 2024-07-24
docs(x/precisebank): Add state, keepers, events, client to spec 

drklee3:
## Description
- Include the typical parts of spec, on top of the main math logic

### DIFF
--- a/x/precisebank/README.md
+++ b/x/precisebank/README.md
@@ -381,7 +381,7 @@ $$b'(R) - b(R) = \begin{cases} 0 & r' > r \land f'(1) < f(1) \\
 
 ## State
 
-The `x/precisebank` keeps state of the following:
+The `x/precisebank` module keeps state of the following:
 1. Account fractional balances.
 2. Remainder amount. This amount represents the fractional amount that is backed
    by the reserve account but not yet in circulation. This can be non-zero if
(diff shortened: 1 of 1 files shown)


x/precisebank/README.md
```diff
@@ -354,3 +378,195 @@ $$b'(R) - b(R) = \begin{cases} 0 & r' > r \land f'(1) < f(1) \\
...
 0 & r' \leq r \land f'(1) \geq f(1) \\
 \end{cases}$$
+
+## State
+
+The `x/precisebank` module keeps state of the following:
+1. Account fractional balances.
+2. Remainder amount. This amount represents the fractional amount that is backed
+   by the reserve account but not yet in circulation. This can be non-zero if
+   a fractional amount less than `1ukava` is minted.
```
nddeluca, line 388: Should we add a note about current mainnet state and remainder is always zero (we think) at the end of each block due to the EVM always using matching mint/burns?

x/precisebank/README.md
```diff
@@ -354,3 +378,195 @@ $$b'(R) - b(R) = \begin{cases} 0 & r' > r \land f'(1) < f(1) \\
...
+    }
+  ]
+}
+```
+
+#### MintCoins
+
+```json
+{
+  "type": "mint",
```
nddeluca, line 456: This is actually `coinbase`?

x/precisebank/README.md
```diff
@@ -354,3 +378,195 @@ $$b'(R) - b(R) = \begin{cases} 0 & r' > r \land f'(1) < f(1) \\
...
+```
+
+## Messages
+
+The `x/precisebank` module does not have any messages and is intended to be used
+by other modules as a replacement of the bank module.
+
+## Events
+
+### Keeper Events
```
nddeluca, line 416: We missed the "coin_spent" and "coin_received" events, oops

nddeluca (approved):
The spec is approved, feel free to merge into the other PR and we can get more eyes on it.

However, a couple things to update -- the `mint` event is called `coinbase`, and we forgot about the `coin_spent` and `coin_received` events in the x/precisebank module, but that's work for another PR and we can add those to the spec then.

x/precisebank/README.md
drklee3: yup, updated

Issue #1636, opened 2023-06-22
Linters haven't been run recently


faddat:
While looking into #1633 I discovered that the linters hadn't recently run, I am guessing because golangci.yml wasn't .golangci.yml.

Before adding PFM or changing things I'd like to get the linters running correctly again, to reduce the likelihood of error.

PR #1977, opened 2024-07-25
chore(linter): update Golang linter configuration

sesheffield:
## Description
* adding a .golangci.yml file to dictate the golang linters.
   * disabling any that are currently failing to prevent master from failing
* added CODEOWNERS 

To test locally: 
* `act -j golangci-lint`

### Strategy moving forward:
* Team has decided to use [`new-from-rev` ](https://golangci-lint.run/usage/configuration/) that runs the linter against the files changed in a PR. 
  * --new-from-rev REV               Show only new issues created after git revision REV
* We should move forward enabling linters by uncommenting in: `.golangci.yml`
* Fixing linters should be continually addressed as normal tech debt and apart of the typical development cycle. 

## Checklist


.github/workflows/ci-lint.yml
```diff
@@ -7,11 +7,28 @@ jobs:
...
+      checks: write # allow write access to checks to allow the action to annotate code in the PR.
     steps:
-      - uses: actions/checkout@v4
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with: { fetch-depth: 0 }
+      - name: Load Version
+        id: load-version
+        run: |
+          echo "GOLANGCI_VERSION=v$(cat .golangci-version)"
```
pirtleshell, line 19: i think you're going to end up with double `v`s in the env var here

.golangci-version
```diff
@@ -0,0 +1 @@
+v1.59
```
pirtleshell, line 1: nit: can you setup your editor to add newlines at the end of files?

.github/workflows/ci-lint.yml
```diff
@@ -7,11 +7,28 @@ jobs:
...
+  golang-sec:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with: { fetch-depth: 0 }
+      - name: Run Gosec Security Scanner
+        uses: cosmos/gosec@v0.10.4
+        with:
+          args: ./...
```
pirtleshell, line 34: this should be added in a separate pr, and in a way that will not result in the master branch CI failing (make it only operate on a diff, disable rules that we break to be re-enabled one at a time in subsequent prs, or update code if it is very simple and obviously incorrect). 

if we break a rule that requires a meaningful & consequential change to the code, it should be made in a pr that highlights what is happening and why the change is safe.

on principle & to the best of our ability, we should never merge a commit that fails the PR CI or will break master's CI

.github/CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+# Global rule:
+* @rhuairahrighairidh @karzak @pirtleshell @drklee3 @nddeluca @DracoLi @evgeniy-scherbina @sesheffield
```
pirtleshell, line 2: 🏆 excellent add.
can you include a comment referencing [this](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) for easy access/explanation of this for those not in the know?

pirtleshell (changes requested):
i appreciate your changes here. however, i think this PR is doing too much at once and should be broken into more singularly-focused pieces.

additionally your commit messages and PR descriptions should include details on why changes are being made and be more specific. for example "added a new make command", does not tell a reader of the git log what was actually changed. PR titles & commit messages should follow the guidelines outlined in https://www.conventionalcommits.org/en/v1.0.0/. Example: "feat(ci): enabled golang linters" (assuming this is the only thing the PR does, which is not the case for this PR).

regarding the linters: i think we need to first decide as a team what lints to use. i also think the final form of linting should prefer starting with the defaults and only adding or disabling functionality as the need arises, rather than starting with everything disabled and enabling individual linters/rules. this will also make the config file much simpler and easier to read. exceptions to the defaults, including additional new linters, should have explanatory comments in the config YAML.

some other thoughts/questions:
* please also delete the existing, incorrectly named config file
* please add @faddat as a co-author to the commit that changes the config file. he previously did a good amount of prep for the linters, including identifying the misnamed file in #1636 (thank you, @faddat!!)
* does the current codebase pass all these lints? are we sure the next commit to master won't fail CI because of things enabled here?

sesheffield:
> i appreciate your changes here. however, i think this PR is doing too much at once and should be broken into more singularly-focused pieces.
> 
> additionally your commit messages and PR descriptions should include details on why changes are being made and be more specific. for example "added a new make command", does not tell a reader of the git log what was actually changed. PR titles & commit messages should follow the guidelines outlined in https://www.conventionalcommits.org/en/v1.0.0/. Example: "feat(ci): enabled golang linters" (assuming this is the only thing the PR does, which is not the case for this PR).
> 
> regarding the linters: i think we need to first decide as a team what lints to use. i also think the final form of linting should prefer starting with the defaults and only adding or disabling functionality as the need arises, rather than starting with everything disabled and enabling individual linters/rules. this will also make the config file much simpler and easier to read. exceptions to the defaults, including additional new linters, should have explanatory comments in the config YAML.
> 
> some other thoughts/questions:
> 
> * please also delete the existing, incorrectly named config file
> * please add @faddat as a co-author to the commit that changes the config file. he previously did a good amount of prep for the linters, including identifying the misnamed file in [Linters haven't been run recently #1636](https://github.com/Kava-Labs/kava/issues/1636) (thank you, @faddat!!)
> * does the current codebase pass all these lints? are we sure the next commit to master won't fail CI because of things enabled here?


1. please also delete the existing, incorrectly named config file
    * not sure what file you're referring to. 
2. please add @faddat as a co-author to the commit that changes the config file. he previously did a good amount of prep for the linters, including identifying the misnamed file in [Linters haven't been run recently #1636](https://github.com/Kava-Labs/kava/issues/1636) (thank you, @faddat!!)
    * not a valid codeowner [aa558db](https://github.com/Kava-Labs/kava/pull/1977/commits/aa558dbeddda48c6ac00ff20281bdfdc83bde59b)
3. does the current codebase pass all these lints? are we sure the next commit to master won't fail CI because of things enabled here?
    * [8944ae6](https://github.com/Kava-Labs/kava/pull/1977/commits/8944ae607b9c2489503bdfca2f1f330015a3da11)

sesheffield:
@pirtleshell 

> regarding the linters: i think we need to first decide as a team what lints to use. i also think the final form of linting should prefer starting with the defaults and only adding or disabling functionality as the need arises, rather than starting with everything disabled and enabling individual linters/rules. this will also make the config file much simpler and easier to read. exceptions to the defaults, including additional new linters, should have explanatory comments in the config YAML.

I agree that we should discuss and finalize the linters as a team. I can set up a time with the team to discuss this further. My approach was to add all linters initially, as I found that even the basic linters were failing, which made starting with a comprehensive setup more practical.

The linters currently enabled and disabled in the configuration align with the existing "style" the team has developed. Starting with all linters that we can have enabled allows us to identify and address any issues early on. Once we have a baseline where everything is passing, we can then adjust the configuration by enabling or disabling specific rules as needed. This approach will help us ensure that we maintain a high standard of code quality from the start.

I agree that the final configuration should be simplified and commented for clarity. Once we finalize the linters and rules, I'll make sure to document any exceptions and customizations in the config YAML.

Let me know if you have any specific linters or rules you'd like to prioritize, and we can adjust accordingly.

.golangci.yml
```diff
@@ -0,0 +1,145 @@
...
+  misspell:
+    locale: US
+    ignore-words: expect
+  prealloc:
+    simple: true # enables simple preallocation checks
+    range-loops: true # enabled preallocation checks in range loops
+    for-loops: false # disables preallocation checks in for loops
+
+issues:
+  exclude-rules:
```
pirtleshell, line 123: can we remove the exclusions until they rear their ugly heads in future commits?

pirtleshell (approved):
looks great, thanks for the much needed addition!

PR #1981, opened 2024-07-29
test(insolvency): add tests for Kava lend insolvency check

sesheffield:
## Description
[closes](https://app.shortcut.com/kava-labs/story/13867/write-failing-test-for-global-insolvency-check)
[security questionnaire
](https://app.shortcut.com/kava-labs/story/13868/fill-out-security-questionnaire-for-hard-insolvency-check-fix)

When a user borrows funds from Kava Lend, the system checks if there are sufficient funds to process the request. A lack of sufficient funds indicates insolvency, meaning that bad debt has accumulated in a particular money market, resulting in liabilities exceeding deposits.

The current check is global. For example, if a user wants to borrow ATOM, the `ValidateBorrow` function will throw an error if the BNB market is insolvent. We want to change this check to be local, so that an insolvency error is only thrown if the specific money market being borrowed from is insolvent, not another unrelated market.

This PR adds tests to highlight the current issue with the borrowing functionality in Kava Lend. The current global insolvency check causes borrow requests to fail even if the specific market being borrowed from is solvent. These tests demonstrate the problem and will serve as the basis for a follow-up PR to fix the issue.

* Added new tests to:
  * Attempt to borrow funds from an insolvent market (should fail).
  * Attempt to borrow funds from a solvent market when there is also an insolvent market (currently does not fail but should).
* Example:
    * Money market x is solvent
    * Money market y is insolvent
    * User goes to borrow from money market x - it fails.
    * User goes to borrow from money market y - it fails.
    * Once the logic is fixed, the test for money market x should pass, and the test for money market y should continue to fail. 

* Additional Notes
  * These tests are designed to pass once the issue is fixed in a follow-up PR.

## Checklist

### DIFF
--- a/x/hard/keeper/borrow_test.go
+++ b/x/hard/keeper/borrow_test.go
@@ -1,7 +1,6 @@
 package keeper_test
 
 import (
-	"fmt"
 	"strings"
 	"time"
 
@@ -837,7 +836,6 @@ func (suite *KeeperTestSuite) TestValidateBorrow() {
 		sdk.NewCoins(sdk.NewCoin("ukava", availableToBorrow.AmountOf("ukava"))),
 	)
 	suite.Require().Error(err)
-	fmt.Println("err: ", err)
 	suite.Require().True(strings.Contains(err.Error(), "protocol reserves exceed available cash"))
 
 	suite.ctx = suite.ctx.WithBlockTime(suite.ctx.BlockTime().Add(blockDuration))
--- a/x/hard/keeper/borrow_test.go
+++ b/x/hard/keeper/borrow_test.go
@@ -25,18 +25,15 @@ const (
 
 func (suite *KeeperTestSuite) TestBorrow() {
 	type setupArgs struct {
-		usdxBorrowLimit sdk.Dec
-		priceKAVA       sdk.Dec
-		loanToValueKAVA sdk.Dec
-		priceBTCB       sdk.Dec
-		loanToValueBTCB sdk.Dec
-		priceBNB        sdk.Dec
-		loanToValueBNB  sdk.Dec
-		borrower        sdk.AccAddress
-		depositCoins    []sdk.Coin
-		wantDepositErr  string
-
-		// renamed this to be more clear
+		usdxBorrowLimit    sdk.Dec
+		priceKAVA          sdk.Dec
+		loanToValueKAVA    sdk.Dec
+		priceBTCB          sdk.Dec
+		loanToValueBTCB    sdk.Dec
+		priceBNB           sdk.Dec
+		loanToValueBNB     sdk.Dec
+		borrower           sdk.AccAddress
+		depositCoins       []sdk.Coin
 		initialBorrowCoins sdk.Coins
 	}
 
@@ -640,9 +637,7 @@ func (suite *KeeperTestSuite) TestBorrow() {
 			suite.keeper = keeper
 			// Run BeginBlocker once to transition MoneyMarkets
 			hard.BeginBlocker(suite.ctx, suite.keeper)
-			if err = suite.keeper.Deposit(suite.ctx, tc.setup.borrower, tc.setup.depositCoins); err != nil {
-				suite.Require().Equal(tc.setup.wantDepositErr, err.Error())
-			}
+			suite.Require().NoError(suite.keeper.Deposit(suite.ctx, tc.setup.borrower, tc.setup.depositCoins)
 			// Execute user's previous borrows
 			if err = suite.keeper.Borrow(suite.ctx, tc.setup.borrower, tc.setup.initialBorrowCoins); tc.setup.initialBorrowCoins.IsZero() {
 				suite.Require().True(strings.Contains(err.Error(), "cannot borrow zero coins"))
--- a/x/hard/keeper/borrow_test.go
+++ b/x/hard/keeper/borrow_test.go
@@ -1,6 +1,7 @@
 package keeper_test
 
 import (
+	"fmt"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


x/hard/keeper/borrow_test.go
```diff
@@ -39,389 +38,580 @@ func (suite *KeeperTestSuite) TestBorrow() {
...
 	}
 	type errArgs struct {
 		expectPass bool
 		contains   string
 	}
 	type borrowTest struct {
 		name    string
-		args    args
-		errArgs errArgs
+		args    []args
```
drklee3, line 49: I think we can refactor this to make it more clear and maintainable with the added multiple borrows

args currently contain a lot of the initial setup so instead of reusing it for following borrows we could split this into something like this:

```go
type setupArgs struct {
	usdxBorrowLimit           sdk.Dec
	priceKAVA                 sdk.Dec
	loanToValueKAVA           sdk.Dec
	priceBTCB                 sdk.Dec
	loanToValueBTCB           sdk.Dec
	priceBNB                  sdk.Dec
	loanToValueBNB            sdk.Dec
	borrower                  sdk.AccAddress
	depositCoins              []sdk.Coin
	// renamed this to be more clear
	initialBorrowCoins        sdk.Coins
	expectedAccountBalance    sdk.Coins
	expectedModAccountBalance sdk.Coins
	wantDepositErr            string
}

type borrowArgs struct {
	borrowCoins sdk.Coins
	// Empty if the borrow should succeed
	wantErr string
}

type borrowTest struct {
	name    string
	setup   setupArgs
	borrows []borrowArgs
}
```

x/hard/keeper/borrow_test.go
```diff
@@ -39,389 +38,580 @@ func (suite *KeeperTestSuite) TestBorrow() {
...
-				_, f := suite.keeper.GetBorrow(suite.ctx, tc.args.borrower)
-				suite.Require().True(f)
-			} else {
-				suite.Require().Error(err)
-				suite.Require().True(strings.Contains(err.Error(), tc.errArgs.contains))
+			for i, arg := range tc.args {
+				if i == 0 {
+					// Initialize test app and set context
+					tApp = app.NewTestApp()
+					ctx = tApp.NewContext(true, tmproto.Header{Height: 1, Time: tmtime.Now()})
```
drklee3, line 483: With the setup args moved to it's own field, we can then do the initialization before the loop without needing to check the index - only looping for borrows

x/hard/keeper/borrow_test.go
```diff
@@ -561,4 +751,32 @@ func (suite *KeeperTestSuite) TestValidateBorrow() {
...
 	)
 	suite.Require().NoError(err)
+
+	// now that it's all that you can borrow, we shouldn't be able to borrow anything
+	err = suite.keeper.Borrow(
+		suite.ctx,
+		borrower,
+		sdk.NewCoins(sdk.NewCoin("ukava", sdkmath.NewInt(2*USDX_CF))),
+	)
+	suite.Require().Error(err)
```
drklee3, line 761: Would be good practice to check the exact error values as much as possible instead of just being any error

Same for the other ones below

x/hard/keeper/borrow_test.go
```diff
@@ -39,389 +38,580 @@ func (suite *KeeperTestSuite) TestBorrow() {
...
+					borrower:                  sdk.AccAddress(crypto.AddressHash([]byte("test"))),
+					depositCoins:              []sdk.Coin{sdk.NewCoin("ukava", sdkmath.NewInt(100*KAVA_CF))},
+					previousBorrowCoins:       sdk.NewCoins(),
+					borrowCoins:               sdk.NewCoins(sdk.NewCoin("ukava", sdkmath.NewInt(20*KAVA_CF))),
+					expectedAccountBalance:    sdk.NewCoins(sdk.NewCoin("ukava", sdkmath.NewInt(20*KAVA_CF)), sdk.NewCoin("btcb", sdkmath.NewInt(100*BTCB_CF)), sdk.NewCoin("bnb", sdkmath.NewInt(100*BNB_CF)), sdk.NewCoin("xyz", sdkmath.NewInt(1))),
+					expectedModAccountBalance: sdk.NewCoins(sdk.NewCoin("ukava", sdkmath.NewInt(1080*KAVA_CF)), sdk.NewCoin("usdx", sdkmath.NewInt(200*USDX_CF)), sdk.NewCoin("busd", sdkmath.NewInt(100*BUSD_CF))),
+				},
+				{
+					borrower:                  sdk.AccAddress(crypto.AddressHash([]byte("test"))),
+					depositCoins:              []sdk.Coin{sdk.NewCoin("ukava", sdkmath.NewInt(100*KAVA_CF))}, // too much
```
drklee3, line 388: Use of `depositCoins` and non-borrow fields after the first `args` in the list don't seem to be used in the test, so this test case seems to be not actually testing as intended

The refactor of split up args would make it easier to avoid use of fields that won't be used and make it more clear what is being tested

drklee3 (changes requested):
Some suggestions on refactoring to make the test more clear, seems like it's easy to make test cases that don't actually test what we are trying to test right now

x/hard/keeper/borrow_test.go
sesheffield: I can update as suggested, but the current task is to add new tests to fix a specific bug. However, your suggestions are recommending to going back and fixing how all the tests are running and structured, which is distracting from the main focus.

drklee3: yeah understandable to avoid making changes outside goal of resolving the bug, though I think this still fits with testing this edge case since it requires your addition of multiple borrow cases which changes the structure - just want to make sure any modifications we make are clear and intentional even if it requires some shuffling around in order to support it

x/hard/keeper/borrow_test.go
```diff
@@ -561,4 +811,37 @@ func (suite *KeeperTestSuite) TestValidateBorrow() {
...
+	hard.BeginBlocker(suite.ctx, suite.keeper)
+
+	// Should error since ukava has become insolvent at this point
+	err = suite.keeper.Borrow(
+		suite.ctx,
+		borrower,
+		sdk.NewCoins(sdk.NewCoin("ukava", availableToBorrow.AmountOf("ukava"))),
+	)
+	suite.Require().Error(err)
+	suite.Require().True(strings.Contains(err.Error(), "protocol reserves exceed available cash"))
```
drklee3, line 834: not blocking merge, can be updated in the next pr if you want - just easier error checking

```suggestion
	suite.Require().ErrorContains(err, "protocol reserves exceed available cash")
```

x/hard/keeper/borrow_test.go
```diff
@@ -25,268 +24,515 @@ const (
...
+				},
+			},
+			borrows: []borrowArgs{
+				{
+					borrowCoins: sdk.NewCoins(sdk.NewCoin("ukava", sdkmath.NewInt(20*KAVA_CF))),
+					wantErr:     "",
+				},
+				{
+					borrowCoins: sdk.NewCoins(sdk.NewCoin("ukava", sdkmath.NewInt(20*KAVA_CF))),
+					wantErr:     "",
```
drklee3, line 456: Is this second borrow supposed to fail with "not enough collateral supplied by account"?

With the test name "... second insufficient funds blocks"

sesheffield: I thought I removed this one already since it's already being handled above in multiple test cases. good catch

sesheffield: [done](https://github.com/Kava-Labs/kava/pull/1981/commits/59624f31ff71e3c7aa37aca411f6fbe874b9d55a)

drklee3 (approved):
Looks good!

PR #1986, opened 2024-08-01
docs: update changelog for v0.26.2

pirtleshell:
The v0.26.2 release link is broken right now because the tag does not yet exist (this merged commit will be the tag).

The intention is to publish all currently-unreleased v0.26.x branch changes in a patch before publishing an iavl v1 release candidate in an effort to minimize the differences between a forthcoming iavl v1 version & current mainnet.


CHANGELOG.md
```diff
@@ -412,7 +419,8 @@ the [changelog](https://github.com/cosmos/cosmos-sdk/blob/v0.38.4/CHANGELOG.md).
 [#750]: https://github.com/Kava-Labs/kava/pull/750
 [#751]: https://github.com/Kava-Labs/kava/pull/751
 [#780]: https://github.com/Kava-Labs/kava/pull/780
-[unreleased]: https://github.com/Kava-Labs/kava/compare/v0.26.1...release/v0.26.x
+[unreleased]: https://github.com/Kava-Labs/kava/compare/v0.26.2...release/v0.26.x
+[v0.26.2]: https://github.com/Kava-Labs/kava/compare/v0.26.2...v0.26.1
```
drklee3, line 423: I think it should be the other way around, older tag first then newer in the url. Some previous versions also link it the opposite way and those don't work like `v0.26.1`

```suggestion
[v0.26.2]: https://github.com/Kava-Labs/kava/compare/v0.26.1...v0.26.2
```

nddeluca: Good catch

pirtleshell: thanks! fixed this and all the other incorrect links

PR #1987, opened 2024-08-01
deps: bump cosmos-sdk to v0.47.13-kava.1

pirtleshell:
## Description

Updates cosmos-sdk to kava's fork based on v0.47.13 (from v0.47.10)

## Checklist


go.mod
```diff
@@ -8,9 +8,9 @@ require (
 	cosmossdk.io/simapp v0.0.0-20231127212628-044ff4d8c015
 	github.com/Kava-Labs/opendb v0.0.0-20240719173129-a2f11f6d7e51
 	github.com/cenkalti/backoff/v4 v4.1.3
-	github.com/cometbft/cometbft v0.37.4
+	github.com/cometbft/cometbft v0.37.5
 	github.com/cometbft/cometbft-db v0.9.1
-	github.com/cosmos/cosmos-proto v1.0.0-beta.4
+	github.com/cosmos/cosmos-proto v1.0.0-beta.5
 	github.com/cosmos/cosmos-sdk v0.47.10
```
drklee3, line 14: Even though it's being replaced, should we update this to `v0.47.13` as well so it's not misleading?

drklee3 (approved):
Looks good, also went through the cosmos changes, fairly minor things

go.mod
nddeluca: ^ Agree

nddeluca:
Soo not to nitpick and it is not required for this PR, but @drklee3 set a good example of providing compare & changelogs for the dependency being updated https://github.com/Kava-Labs/kava/pull/1988.

Having the link to compare, etc is a nice helper during review.  Maybe we should make a git template for fork bumps?

nddeluca (changes requested):
There is actually a state breaking fix in v0.47.11, lets mark this commit with a `deps!` to signify and also discuss upgrade path.

pirtleshell:
The breaking change in cosmos-sdk from v0.47.10 -> v0.47.11 means this can't be put on `v0.26.2` without a coordinated upgrade so i am closing this PR.

PR #1990, opened 2024-08-02
fix(e2e): Use docker compose V2 for kvtool and Makefile

nddeluca:
This updates the kvtool submodule to the latest master, brining in @pirtleshell's fix for docker compose v2.  In addition, updates the makefile commands to also use V2 instead of V1.

This fixes failing e2e tests in actions without the `docker-compose` alias.

More details: https://docs.docker.com/compose/migrate/

KVTool Fix - https://github.com/Kava-Labs/kvtool/pull/201


Makefile
```diff
@@ -346,7 +346,7 @@ start-remote-sims:
 	# submit an array job on AWS Batch, using 1000 seeds, spot instances
 	aws batch submit-job \
 		-—job-name "master-$(VERSION)" \
-		-—job-queue “simulation-1-queue-spot" \
+		-—job-queue "simulation-1-queue-spot" \
```
drklee3, line 349: sneaky “

drklee3 (approved):

PR #1992, opened 2024-08-05
chore(make): Allow docker comamnd override and error if not installed

nddeluca:
This updates the makefile to allow the command used for docker to be overridden with the DOCKER variable, and adds a check that fails if docker is not installed.  Tested with `make DOCKER=foo`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ print-version:
 ################################################################################
 LEDGER_ENABLED ?= true
 
-DOCKER:=docker
+DOCKER ?= docker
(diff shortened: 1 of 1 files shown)


sesheffield (approved):

nddeluca:
It seems this causes more issues since Makefile is not always used in context's with docker installed

nddeluca:
Thinking more about this, this is not the correct path -- we also have users installing from source that do not have docker and with this change, they won't be able to run `make install`.  Closing, @sesheffield maybe there are other methods that are more local to the targets using docker, but I also think the errors should be pretty self explanatory?

PR #1993, opened 2024-08-07
chore(lint): Disable funlen for test functions

nddeluca:
This adds a regular expression that matches `func Test...` or `func (suite *Suite) Test...` style functions and disables the length check. An example from e2e tests that failed lint:

`func (suite *IntegrationTestSuite) TestEip712BasicMessageAuthorization()`

I tested with different variations locally on some existing functions and new functions.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -81,6 +81,13 @@ linters:
 #    - whitespace
     - wrapcheck
 
+issues:
+  exclude-rules:
+    # Disable funlen for "func Test..." or func (suite *Suite) Test..." type functions
+    # These functions tend to be descriptive and exceed length limits.
+    - source: "^func (\\(.*\\) )?Test"
+      linters:
+        - funlen
 
 linters-settings:
   errcheck:


.golangci.yml
```diff
@@ -81,6 +81,13 @@ linters:
 #    - whitespace
     - wrapcheck
 
+issues:
+  exclude-rules:
+    # Disable funlen for "func Test..." or func (suite *Suite) Test..." type functions
+    # These functions tend to be descriptive and exceed length limits.
+    - source: "^func (\\(.*\\) )?Test"
+      linters:
+        - funlen
```
sesheffield, line 90: rather than regex
```
issues:
  exclude-rules:
    - linters:
        - funlen
      path: _test\.go$
```
`path: _test\.go$ specifies that the exclusion should apply to files ending in _test.go.`

sesheffield (approved):

.golangci.yml
nddeluca: I thought about it, but it'd still be nice to validate function length across all other functions, including those defined in test files.  We should hold test and non-test code to the same standard, with the exception being functions that just reference test cases.

In addition, the regex above is smallest exception we can make, and unless we have a strong reason to, imo it's best to relax linting rules in small increments.

.golangci.yml
```diff
@@ -81,6 +81,13 @@ linters:
 #    - whitespace
     - wrapcheck
 
+issues:
+  exclude-rules:
+    # Disable funlen for "func Test..." or func (suite *Suite) Test..." type functions
+    # These functions tend to be descriptive and exceed length limits.
```
pirtleshell, line 87: love this description of what and why 👍

pirtleshell (approved):

PR #1999, opened 2024-08-08
feat!(precompile): Add registry and genesis tests

nddeluca:
Based on @evgeniy-scherbina's previous work, this adds a new precompile module which defines a contract module with an example noop contract that will be will be used for implementing test functions.

In addition, it defines a registry module that instantiates stateful precompile contracts and associates them with an address in a global registry defined at [kava-labs/go-ethereum](https://github.com/Kava-Labs/go-ethereum/tree/master/precompile/modules). See precompile/README.md for more information.

The go-ethereum and etheremint replace statements are updated to support these changes as well as an update to kvtool which includes genesis state for the registry.NoopContractAddress.

This update initializes the contract's EthAccount with a non-zero sequence, sets the code hash to keccak256(0x01), and sets the contract code to 0x01, and adds the contract address to the enabled list.  See `tests/e2e/e2e_precompile_genesis_test.go` for an overview of the expected genesis state for an enabled precompile.

Depends on https://github.com/Kava-Labs/kvtool/pull/206

### DIFF
--- a/precompile/README.md
+++ b/precompile/README.md
@@ -1,0 +1,27 @@
+# Precompile
+
+The `precompile` module defines stateful precompile contracts used for creating precompile contracts that enhance the EVM, enable interactions between cosmos state and EVM state, or are used soley for testing purposes.
+
+This module is made of two submodules:
+
+- `contracts` - Defines stateful precompiles and their constructors.
+- `registry` - Defines stateful precompile addresses and registers them with the global registry
+ defined at `github.com/kava-labs/go-ethereum/precompile/modules`.
+
+This is architected to isolate the dependency on the global registry (`github.com/kava-labs/go-ethereum/precompile/modules` package) to as few places as possible, have one source of truth for address registration (see `./registry/registry.go`), and isolate registration testing to one package.
+
+In order to use the precompile registry, it must be imported for it's init function to run and register the precompiles.  For the kava app, this is done in the `app.go` file with the import `_ "github.com/kava-labs/kava/precompile/registry"`.  This is done in the `app.go` since this is the main file used for app dependencies and so all modules using the app cause the registry to be loaded.  This is import for any consumers of the app outside of the kava cmd, as well as test code using the app for integration and unit testing.
+
+## Defining a new precompile
+
+1) Add the expected 0x address to the expected list in `./registry/registry_test.go`.
+2) Create a new sub-directory under `./contracts` with a `contract_test.go` and `contract.go` file.
+3) Implement `NewContract` function with associated tests in contract and contract test files.
+4) Add the contract registration to `./registry/registry.go`.
+
+## Contracts
+
+### Noop
+
+This contract is used for testing purposes only and should not be used on public chains.  The functions of this contract (once implemented), will be used to exercise and test the various aspects of the EVM such as gas usage, argument parsing, events, etc. The specific operations tested under this contract are still to be determined.
+
--- a/precompile/README.md
+++ b/precompile/README.md
@@ -1,8 +1,8 @@
 # Precompile
 
-The `precompile` module defines stateful precompile contracts used for creating precompile contracts that enhance the EVM, enable interactions between cosmos state and EVM state, or are used soley for testing purposes.
+The `precompile` package defines stateful precompile contracts used for creating precompile contracts that enhance the EVM, enable interactions between cosmos state and EVM state, or are used soley for testing purposes.
 
-This module is made of two submodules:
+This package is made of two subpackages:
 
 - `contracts` - Defines stateful precompiles and their constructors.
 - `registry` - Defines stateful precompile addresses and registers them with the global registry
@@ -10,7 +10,7 @@ This module is made of two submodules:
 
 This is architected to isolate the dependency on the global registry (`github.com/kava-labs/go-ethereum/precompile/modules` package) to as few places as possible, have one source of truth for address registration (see `./registry/registry.go`), and isolate registration testing to one package.
 
-In order to use the precompile registry, it must be imported for it's init function to run and register the precompiles.  For the kava app, this is done in the `app.go` file with the import `_ "github.com/kava-labs/kava/precompile/registry"`.  This is done in the `app.go` since this is the main file used for app dependencies and so all modules using the app cause the registry to be loaded.  This is import for any consumers of the app outside of the kava cmd, as well as test code using the app for integration and unit testing.
+In order to use the precompile registry, it must be imported for it's init function to run and register the precompiles.  For the kava app, this is done in the `app.go` file with the import `_ "github.com/kava-labs/kava/precompile/registry"`.  This is done in the `app.go` since this is the main file used for app dependencies and so all modules using the app cause the registry to be loaded.  This is important for any consumers of the app outside of the kava cmd, as well as test code using the app for integration and unit testing.
 
 ## Defining a new precompile
 
--- a/precompile/contracts/noop/contract.go
+++ b/precompile/contracts/noop/contract.go
@@ -1,0 +1,18 @@
+package noop
+
+import (
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/precompile/contract"
+)
+
+// NewContract returns a new noop stateful precompiled contract.
+func NewContract() (contract.StatefulPrecompiledContract, error) {
+	precompile, err := contract.NewStatefulPrecompileContract([]*contract.StatefulPrecompileFunction{})
+
+	if err != nil {
+		return nil, fmt.Errorf("failed to instantiate noop precompile: %w", err)
+	}
+
+	return precompile, nil
+}
--- a/precompile/contracts/noop/contract.go
+++ b/precompile/contracts/noop/contract.go
@@ -7,6 +7,11 @@ import (
 )
 
 // NewContract returns a new noop stateful precompiled contract.
+//
+//	This contract is used for testing purposes only and should not be used on public chains.
+//	The functions of this contract (once implemented), will be used to exercise and test the various aspects of
+//	the EVM such as gas usage, argument parsing, events, etc. The specific operations tested under this contract are
+//	still to be determined.
 func NewContract() (contract.StatefulPrecompiledContract, error) {
 	precompile, err := contract.NewStatefulPrecompileContract([]*contract.StatefulPrecompileFunction{})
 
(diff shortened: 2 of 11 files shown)


precompile/README.md
```diff
@@ -0,0 +1,27 @@
+# Precompile
+
+The `precompile` module defines stateful precompile contracts used for creating precompile contracts that enhance the EVM, enable interactions between cosmos state and EVM state, or are used soley for testing purposes.
+
+This module is made of two submodules:
+
+- `contracts` - Defines stateful precompiles and their constructors.
+- `registry` - Defines stateful precompile addresses and registers them with the global registry
+ defined at `github.com/kava-labs/go-ethereum/precompile/modules`.
```
sesheffield, line 9:
```suggestion
This package is made of two subpackages:

- `contracts` - Defines stateful precompiles and their constructors.
- `registry` - Defines stateful precompile addresses and registers them with the global registry
 defined at `github.com/kava-labs/go-ethereum/precompile/modules`.
```

A "submodule" typically refers to a separate module within a larger project, often managed with a separate go.mod file. Submodules are more about versioning and dependency management. In your structure, unless each of these directories (contracts, registry) has its own go.mod file, they are not submodules.

In Go, a "subpackage" refers to a package that is nested within another package directory. Each directory containing Go source files with a package declaration is considered a package. 

Based on the provided structure, contracts and registry are subpackages of the `/preccompiles` package. They are not submodules unless each has its own go.mod file and is managed independently as a separate Go module.

Given that we have an `/x` directory where subdirectories are referred to as modules, we should be careful not to mix up these terms.

precompile/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+This is architected to isolate the dependency on the global registry (`github.com/kava-labs/go-ethereum/precompile/modules` package) to as few places as possible, have one source of truth for address registration (see `./registry/registry.go`), and isolate registration testing to one package.
+
+In order to use the precompile registry, it must be imported for it's init function to run and register the precompiles.  For the kava app, this is done in the `app.go` file with the import `_ "github.com/kava-labs/kava/precompile/registry"`.  This is done in the `app.go` since this is the main file used for app dependencies and so all modules using the app cause the registry to be loaded.  This is import for any consumers of the app outside of the kava cmd, as well as test code using the app for integration and unit testing.
```
sesheffield, line 13:
```suggestion
In order to use the precompile registry, it must be imported for it's init function to run and register the precompiles.  For the kava app, this is done in the `app.go` file with the import `_ "github.com/kava-labs/kava/precompile/registry"`.  This is done in the `app.go` since this is the main file used for app dependencies and so all modules using the app cause the registry to be loaded.  This is important for any consumers of the app outside of the kava cmd, as well as test code using the app for integration and unit testing.
```

precompile/contracts/noop/contract.go
```diff
@@ -0,0 +1,18 @@
+package noop
+
+import (
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/precompile/contract"
+)
+
+// NewContract returns a new noop stateful precompiled contract.
```
sesheffield, line 9: Might be good to update
```
// NewContract returns a new noop stateful precompiled contract.
// 	This contract is used for testing purposes only and should not be used on public chains. 
//	The functions of this contract (once implemented), will be used to exercise and test the various aspects of
// 	the EVM such as gas usage, argument parsing, events, etc. The specific operations tested under this contract are
// 	still to be determined.
```
So future users are reminded with godocs

image

tests/e2e/e2e_precompile_genesis_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+			// accountErr is checked during assertions if we are expecting an account due to
+			// the service returning a not found error for precompiles that are not enabled
+			accountResponse, accountErr := suite.Kava.Grpc.Query.Auth.Account(
+				grpcGenesisContext, &authtypes.QueryAccountRequest{Address: sdkAddress.String()})
+			var account authtypes.AccountI
+			if accountErr == nil {
+				err = suite.Kava.EncodingConfig.Marshaler.UnpackAny(accountResponse.Account, &account)
+				suite.Require().NoError(err)
+			}
```
sesheffield, line 108:
```
accountResponse, accountErr := suite.Kava.Grpc.Query.Auth.Account(
grpcGenesisContext, &authtypes.QueryAccountRequest{Address: sdkAddress.String()})
if (accountErr != nil) != tf.expectIsEthAccount {
    suite.T().Errorf("expected account query to not error for address %s", tf.address)
    return
}

var account authtypes.AccountI
suite.Require().NoError(suite.Kava.EncodingConfig.Marshaler.UnpackAny(accountResponse.Account, &account))
```
little easier to read then having to jump down to see where err is being handled.

tests/e2e/e2e_precompile_genesis_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+			// We ensure both the evm json rpc and x/evm grpc repeat the correct and same code value
+			grpcCodeResponse, err := suite.Kava.Grpc.Query.Evm.Code(grpcGenesisContext,
+				&evmtypes.QueryCodeRequest{Address: evmAddress.String()})
+			suite.Require().NoError(err)
+			rpcCode, err := suite.Kava.EvmClient.CodeAt(context.Background(), evmAddress, genesisHeight)
```
sesheffield, line 114:
```
// We assert both methods of code retrieval report the same value
suite.Equalf(tf.expectCode, rpcCode, "expected code for address %s to match expected", tf.address)
```

moving https://github.com/Kava-Labs/kava/pull/1999/files#diff-d666ccf9b7197bc4365f483161de25f41d750da2d2a4b691d19749208b3f2241R159
to here

tests/e2e/e2e_precompile_genesis_test.go
```diff
@@ -0,0 +1,171 @@
...
+			suite.Require().NoError(err)
+
+			nonce, err := suite.Kava.EvmClient.NonceAt(context.Background(), evmAddress, genesisHeight)
+			suite.Require().NoError(err)
+
```
sesheffield, line 119:
```
suite.Equalf(tf.expectNonce, nonce, "expected nonce for address %s to match expected", tf.address)
```
little easier to read then having to jump down to see where err is being handled.

tests/e2e/e2e_precompile_genesis_test.go
```diff
@@ -0,0 +1,171 @@
...
+			accountResponse, accountErr := suite.Kava.Grpc.Query.Auth.Account(
+				grpcGenesisContext, &authtypes.QueryAccountRequest{Address: sdkAddress.String()})
+			var account authtypes.AccountI
+			if accountErr == nil {
+				err = suite.Kava.EncodingConfig.Marshaler.UnpackAny(accountResponse.Account, &account)
+				suite.Require().NoError(err)
+			}
+
+			// We ensure both the evm json rpc and x/evm grpc repeat the correct and same code value
+			grpcCodeResponse, err := suite.Kava.Grpc.Query.Evm.Code(grpcGenesisContext,
```
sesheffield, line 111:
```
// The GRPC endpoint returns []byte(nil) when the code is not in state, which is different from
// the rpc endpoint that returns []byte{}.
grpcExpectedCode := tf.expectCode
if len(grpcExpectedCode) == 0 {
    grpcExpectedCode = []byte(nil)
}
suite.Equalf(grpcExpectedCode, grpcCodeResponse.Code, "expected code for address %s to match expected", tf.address)
```

little easier to read then having to jump down to see where err is being handled.

precompile/README.md
nddeluca: I have a habit of referring to any unit of code as a module 😓.  Great suggestion and clarification and ya we should ensure when we refer to `./x` directories, we use the term cosmos module to make it distinct from go modules -- though, cosmos main has now made these independent go modules as well.

precompile/contracts/noop/contract.go
nddeluca: Great reminder to document as close to the code/functions as possible and take advantage of godoc

tests/e2e/e2e_precompile_genesis_test.go
```diff
@@ -0,0 +1,171 @@
...
+	"context"
+	"math/big"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+	ethermint "github.com/evmos/ethermint/types"
+	evmtypes "github.com/evmos/ethermint/x/evm/types"
+
```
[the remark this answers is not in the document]
nddeluca: This is actually in line with the standard, default, and localmodule organization

tests/e2e/e2e_precompile_genesis_test.go
nddeluca: Resolved the other similar comments - I explored the different variations here and opted to improve the comments.  This current organization is the most clear imo and is organized based on a "arrange, act, assert" approach.  With this e2e test, we focus on building a view of the blockchain state at block 1, then making assertions on that state.  There are probably better ways to make it more clear, but from that perspective moving the business logic type assertions up into the queries is a bit distracting.

The goal of the error assertions in the queries is to halt the test if we find data unavailable, which signals the state, setup, etc is invalid and running assertions does not make sense -- hence the Require() on these.

nddeluca: Also, with editors it is quite easy to jump to where variables are used, so imo the weight on query & assertion organization is more important.

precompile/contracts/noop/contract.go
```diff
@@ -0,0 +1,18 @@
...
+func NewContract() (contract.StatefulPrecompiledContract, error) {
+	precompile, err := contract.NewStatefulPrecompileContract([]*contract.StatefulPrecompileFunction{})
+
+	if err != nil {
+		return nil, fmt.Errorf("failed to instantiate noop precompile: %w", err)
+	}
+
+	return precompile, nil
+}
```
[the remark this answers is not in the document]
nddeluca: Great points, I added a ticket to revisit this feedback once more of the noop contract as well as other contracts are implemented.

I honestly didn't put a ton of thought into contract construction yet -- just enough to fill out the registry behavior but I do agree with your feedback but without having more code and use cases, it should be re-visited and non-blocking.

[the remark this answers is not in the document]
nddeluca: I also do like the attention to extensibility, I always prefer that as well

sesheffield (approved):
LGTM 🚀

PR #2008, opened 2024-08-16
chore(ci): update internal testnet genesis and seed to support committee voting

DracoLi:
## Description
- Adds committee MsgVote to evm `eip712_allowed_msgs` to allow for committee voting.
- Updates internal testnet seed script to submit a committee proposal for testing purposes.
- Changes gov voting period to 7 days to help with testing active proposals.

### DIFF
--- a/.github/scripts/seed-internal-testnet.sh
+++ b/.github/scripts/seed-internal-testnet.sh
@@ -185,6 +185,23 @@ sleep $AVG_SECONDS_BETWEEN_BLOCKS
 updatedEvmUtilParams=$(curl https://api.app.internal.testnet.us-east.production.kava.io/kava/evmutil/v1beta1/params)
 printf "updated evm util module params\n %s" , "$updatedEvmUtilParams"
 
+# submit a kava token committee proposal
+COMMITTEE_PROP_TEMPLATE=$(
+  cat <<'END_HEREDOC'
+{
+    "@type": "/cosmos.gov.v1beta1.TextProposal",
+    "title": "The next big thing signaling proposal.",
+    "description": "The purpose of this proposal is to signal support/opposition to the next big thing"
+}
+END_HEREDOC
+)
+committeeProposalFileName="$(date +%s)-committee-proposal.json"
+touch $committeeProposalFileName
+echo "$COMMITTEE_PROP_TEMPLATE" >$committeeProposalFileName
+
+# committee 4 is the hard token committee
+kava tx committee submit-proposal 4 "$committeeProposalFileName" --gas 2000000 --gas-prices 0.01ukava --from god -y
+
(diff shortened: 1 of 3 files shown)


.github/scripts/seed-internal-testnet.sh
```diff
@@ -185,6 +185,23 @@ sleep $AVG_SECONDS_BETWEEN_BLOCKS
...
+  cat <<'END_HEREDOC'
+{
+    "@type": "/cosmos.gov.v1beta1.TextProposal",
+    "title": "The next big thing signaling proposal.",
+    "description": "The purpose of this proposal is to signal support/opposition to the next big thing"
+}
+END_HEREDOC
+)
+committeeProposalFileName="$(date +%s)-committee-proposal.json"
+touch $committeeProposalFileName
```
nddeluca, line 199: We should be able to remove the touch since echo will create the file

.github/scripts/seed-internal-testnet.sh
```diff
@@ -185,6 +185,23 @@ sleep $AVG_SECONDS_BETWEEN_BLOCKS
...
+echo "$COMMITTEE_PROP_TEMPLATE" >$committeeProposalFileName
+
+# committee 4 is the hard token committee
+kava tx committee submit-proposal 4 "$committeeProposalFileName" --gas 2000000 --gas-prices 0.01ukava --from god -y
```
nddeluca, line 203: I like to use `--gas auto --gas-adjustment 1.5` for scripts like this unless I want to know about gas usage going over a certain threshold

nddeluca (approved):
Nice heredoc usage

.github/scripts/seed-internal-testnet.sh
nddeluca: Also, maybe it's worth using a variable so we can put a name to the `4`?

ci/env/kava-protonet/genesis.json
```diff
@@ -2172,7 +2172,7 @@
         "max_deposit_period": "172800s"
       },
       "voting_params": {
-        "voting_period": "600s"
+        "voting_period": "604800s"
```
pirtleshell, line 2175: is this being updated on purpose? it's more useful for our internal testnets to use short voting periods so that we can submit & pass test proposals (waiting 10min vs waiting 1 week)

DracoLi: yeah, frontend requested this so its up longer for testing on webapp. I suggest we increase it for now and can revert it back later.

.github/scripts/seed-internal-testnet.sh
DracoLi: good point, this one shouldn't go over. I'll update it. and create a var for the token committee

PR #2012, opened 2024-08-21
fix(x/precisebank): Avoid blocked addr error on SendCoinsFromAccountToModule

drklee3:
## Description
`SendCoinsFromAccountToModule` and `SendCoins` should not error even if the recipient module is blocked.

Blocked address check is only to be done by `SendCoinsFromModuleToAccount`

This fixes this bug and adds corresponding test.

Excluding e2e test for this specific issue at the moment: Unreliable way for a consistent fee_collector state as other tests may affect it


pirtleshell (approved):
thanks to @downing034 for catching this!

LGTM, thanks for adding a test

PR #2015, opened 2024-08-26
ci: disable cd-protonet workflow

sesheffield:
## Description

Disabling the protonet workflow to avoid any issues when working through resetting the network to work with new infra. 
This will be eventually re-enabled, so leaving as a comment. 

## Checklist


pirtleshell (approved):
👍

PR #2016, opened 2024-08-29
chore: Add hardhat and viem based E2E test suite

nddeluca:
This adds a E2E test suite using hardhat and viem for testing EVM compatibility, precompiles, and integrations with smart contracts.

Hardhat is chosen for it's familiarity and great tooling/dev experience.  Viem is chosen due to it's excellent typescript support, clean and modular API, and it's extensibility.  It's also a project that is extensively tested.

The test suite is written to run on both hardhat and kvtool networks in order to make assertions about compatibility and use hardhat's advanced features when necessary for unit testing supporting contracts.  The EVM based accounts are set up to be the same on both networks and the configuration is tested by signing transactions to ensure each environment is set up correctly.

CI is fully integrated to run both hardhat and kvtool networks with the test suite as well as ensure linting and formatting has been run.

Github action changes have been tested locally by:
```
act -W '.github/workflows/ci-default.yml' -j test-e2e-evm --bind
act -W '.github/workflows/ci-lint.yml' -j e2e-evm-lint
```
The `--bind` flag is required for kvtool to correctly mount the kava chain config.

PR #2018, opened 2024-09-06
test: Add auto-generating basic abi tests for solidity interfaces

nddeluca:
This adds solidity source files and tests that document and assert EVM behavior for solidity function ABI's.

Contracts are added in `ABI_BasicTests.sol` that define the different function modifiers, special functions cases, etc and the `abi_basic.test.ts` file asserts all expected behavior dynamically by reading the contract ABI's.

This is currently only testing existing behavior and uses an interface approach to separate the defined ABI from the concrete implementation on chain since these tests will be expanded to use the same interfaces against precompile addresses when run against the kvtool network.

The mocks will be kept as each stateful precompile we add will have an associated mock with the same external behavior (basic input validation, events, etc) so integrating protocols can use them in their test suites without having to always use the kava binary, which is much slower.

Mocks may be extended in the future to support more complex testing or assertions, but for now they are basic and will first be used to verify the behavior of stateful precompiles implementing solidity interfaces matches the expected behavior of the EVM & solidity compiled bytecode.

The empty account and EOA tests are added to increase our e2e testing of EVM behavior through kava & etheremint (hardhat helps us verify we conform to ethereum standards) since these code paths are being modified to support stateful precompiles.  They also serve as documentation for reference or learning.

In addition, PR feedback from @boodyvo, @pirtleshell, @downing034 in #2016 is included here.

### DIFF
--- a/.github/workflows/ci-default.yml
+++ b/.github/workflows/ci-default.yml
@@ -37,6 +37,8 @@ jobs:
         run: make docker-build test-e2e
   test-e2e-evm:
     runs-on: ubuntu-latest
+    env:
+      KAVA_TAG: local
     steps:
       - name: Checkout current commit
         uses: actions/checkout@v4
--- a/.github/workflows/ci-default.yml
+++ b/.github/workflows/ci-default.yml
@@ -61,6 +61,9 @@ jobs:
         run: npm install
         working-directory: tests/e2e-evm
       - name: Run test suite against hardhat network
+        run: npm run compile
+        working-directory: tests/e2e-evm
+      - name: Run test suite against hardhat network
         run: npm run test-hardhat
         working-directory: tests/e2e-evm
       - name: Build local image
--- a/.github/workflows/ci-lint.yml
+++ b/.github/workflows/ci-lint.yml
@@ -45,6 +45,12 @@ jobs:
       - name: Install npm dependencies
         run: npm install
         working-directory: tests/e2e-evm
+      - name: Run solhint
+        run: npm run solhint
+        working-directory: tests/e2e-evm
+      - name: Compile contracts and create artifcats
+        run: npm run compile
+        working-directory: tests/e2e-evm
       - name: Run linter
         run: npm run lint
         working-directory: tests/e2e-evm
--- a/tests/e2e-evm/test/abi_basic.test.ts
+++ b/tests/e2e-evm/test/abi_basic.test.ts
@@ -1,0 +1,162 @@
+import hre from "hardhat";
+import type { ArtifactsMap } from "hardhat/types/artifacts";
+import type { PublicClient, WalletClient, ContractName } from "@nomicfoundation/hardhat-viem/types";
+import { expect } from "chai";
+import { Address, toFunctionSelector, toFunctionSignature, concat } from "viem";
+import { getAbiFallbackFunction, getAbiReceiveFunction } from "./helpers/abi";
+import { whaleAddress } from "./addresses";
+
+interface ContractTestCase {
+  interface: keyof ArtifactsMap;
+  mock: ContractName<keyof ArtifactsMap>;
+}
+
+// ABI_BasicTests assert ethereum + solidity transaction ABI interactions perform as expected.
+describe("ABI_BasicTests", function () {
+  const testCases = [
+    // Test function modifiers without receive & fallback
+    { interface: "NoopNoReceiveNoFallback", mock: "NoopNoReceiveNoFallbackMock" },
+
+    // Test receive + fallback scenarios
+    { interface: "NoopReceiveNoFallback", mock: "NoopReceiveNoFallbackMock" },
+    { interface: "NoopReceivePayableFallback", mock: "NoopReceivePayableFallbackMock" },
+    { interface: "NoopReceiveNonpayableFallback", mock: "NoopReceiveNonpayableFallbackMock" },
+
+    // Test no receive + fallback scenarios
+    { interface: "NoopNoReceivePayableFallback", mock: "NoopNoReceivePayableFallbackMock" },
+    { interface: "NoopNoReceiveNonpayableFallback", mock: "NoopNoReceiveNonpayableFallbackMock" },
+  ] as ContractTestCase[];
+
+  //
+  // Client + Wallet Setup
+  //
+  let publicClient: PublicClient;
+  let walletClient: WalletClient;
+  before("setup clients", async function () {
+    publicClient = await hre.viem.getPublicClient();
+    walletClient = await hre.viem.getWalletClient(whaleAddress);
+  });
+
+  //
+  // Test each function defined in the interface ABI
+  //
+  // Only payable functions may be sent value
+  // Any function (payable, non-payable, view, pure) can be called via a transaction
+  // All functions can be provided calldata regardless of their arguments
+  for (const tc of testCases) {
+    describe(tc.interface, function () {
+      const abi = hre.artifacts.readArtifactSync(tc.interface).abi;
+      const receiveFunction = getAbiReceiveFunction(abi);
+      const fallbackFunction = getAbiFallbackFunction(abi);
+
+      let mockAddress: Address;
+      before("deploy mock", async function () {
+        mockAddress = (await hre.viem.deployContract(tc.mock)).address;
+      });
+
+      describe("State", function() {
+        it("has code set", async function() {
+          const code = await publicClient.getCode({ address: mockAddress });
+          expect(code).to.not.equal(0);
+        });
+
+        it("has nonce default of 1", async function() {
+          const nonce = await publicClient.getTransactionCount({ address: mockAddress });
+          expect(nonce).to.equal(1);
+        });
+
+        it("has a starting balance of 0", async function() {
+          const balance = await publicClient.getBalance({ address: mockAddress });
+          expect(balance).to.equal(0n);
+        });
+      });
+
+      for (const funcDesc of abi) {
+        if (funcDesc.type !== "function") {
+          continue;
+        }
+
+        const funcSelector = toFunctionSelector(toFunctionSignature(funcDesc));
+
+        describe(`ABI function ${funcDesc.name} ${funcDesc.stateMutability}`, function () {
+          const isPayable = funcDesc.stateMutability === "payable";
+
+          it("can be called", async function () {
+            const txData = { to: mockAddress, data: funcSelector, gas: 25000n };
+
+            await expect(publicClient.call(txData)).to.be.fulfilled;
+
+            const txHash = await walletClient.sendTransaction(txData);
+            const txReceipt = await publicClient.waitForTransactionReceipt({ hash: txHash });
+            expect(txReceipt.status).to.equal("success");
+          });
+
+          it(`can ${isPayable ? "" : "not "}be called with value`, async function () {
+            const txData = { to: mockAddress, data: funcSelector, gas: 25000n, value: 1n };
+
+            const txHash = await walletClient.sendTransaction(txData);
+            const txReceipt = await publicClient.waitForTransactionReceipt({ hash: txHash });
+            expect(txReceipt.status).to.equal(isPayable ? "success" : "reverted");
+          });
+
+          it("can be called with extra data", async function () {
+            const data = concat([funcSelector, "0x01"]);
+            const txData = { to: mockAddress, data: data, gas: 25000n };
+
+            const txHash = await walletClient.sendTransaction(txData);
+            const txReceipt = await publicClient.waitForTransactionReceipt({ hash: txHash });
+            expect(txReceipt.status).to.equal("success");
+          });
+        });
+      }
+
+      //
+      // Test ABI special functions -- receive & fallback
+      //
+      // Receive functions are always payable and can not receive data
+      // Fallback functions can be payable or non-payable and can receive data in both cases
+      const testName = `ABI special functions: ${receiveFunction ? "" : "no "}receive and ${fallbackFunction ? fallbackFunction.stateMutability : "no"} fallback`;
+      describe(testName, function () {
+        if (!receiveFunction && (!fallbackFunction || fallbackFunction.stateMutability !== "payable")) {
+          it("can not receive plain transfers", async function () {
+            const txData = { to: mockAddress, gas: 25000n, value: 1n };
+
+            const txHash = await walletClient.sendTransaction(txData);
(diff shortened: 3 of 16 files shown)


.github/workflows/ci-default.yml
```diff
@@ -58,6 +60,9 @@ jobs:
       - name: Install npm dependencies
         run: npm install
         working-directory: tests/e2e-evm
+      - name: Run test suite against hardhat network
```
pirtleshell, line 63:
```suggestion
      - name: Compile contracts and create artifcats
```

tests/e2e-evm/test/helpers/abi.ts
```diff
@@ -0,0 +1,21 @@
+import { Abi, AbiFallback, AbiReceive } from "abitype";
+
+export function getAbiFallbackFunction(abi: Abi): AbiFallback | undefined {
```
pirtleshell, line 3: ha such a golang guy rocking these for-loops

would recommend a single helper func like (i've converted to functional, keep your for loops if ya want 😄)
```ts
export const getAbiFunctionByType = (abi: Abi, funcType: string) => abi.find(f => f.type === funcName)
```

if you want to have the same funcs as here you can convert them to use that:
```ts
const getAbiFallbackFunction = (abi: Abi) => getAbiFunctionByType(abi, "fallback")
const getAbiReceiveFunction = (abi: Abi) => getAbiFunctionByType(abi, "receive")
```

tests/e2e-evm/test/abi_basic.test.ts
```diff
@@ -0,0 +1,162 @@
...
+      //
+      // Receive functions are always payable and can not receive data
+      // Fallback functions can be payable or non-payable and can receive data in both cases
+      const testName = `ABI special functions: ${receiveFunction ? "" : "no "}receive and ${fallbackFunction ? fallbackFunction.stateMutability : "no"} fallback`;
+      describe(testName, function () {
+        if (!receiveFunction && (!fallbackFunction || fallbackFunction.stateMutability !== "payable")) {
+          it("can not receive plain transfers", async function () {
+            const txData = { to: mockAddress, gas: 25000n, value: 1n };
+
+            const txHash = await walletClient.sendTransaction(txData);
```
nddeluca, line 124: I've been meaning to refactor this duplication out and create much more concise tests but I've deferred for now since I haven't landed on an API I like.

tests/e2e-evm/test/abi_basic.test.ts
```diff
@@ -0,0 +1,162 @@
...
+          const data = toFunctionSelector("does_not_exist()");
+          const txData = { to: mockAddress, data: data, gas: 25000n };
+
+          const txHash = await walletClient.sendTransaction(txData);
+          const txReceipt = await publicClient.waitForTransactionReceipt({ hash: txHash });
+          expect(txReceipt.status).to.equal(fallbackFunction ? "success" : "reverted");
+        });
+
+        if (fallbackFunction) {
+          it(`can ${fallbackFunction.stateMutability === "payable" ? "" : "not "}receive transfer with data or invalid function selector`, async function () {
```
nddeluca, line 150: I should create a helper here -- `isPayable(fallbackFunction)` that can be used with an ABI defined function

tests/e2e-evm/test/abi_basic.test.ts
```diff
@@ -0,0 +1,162 @@
...
+          continue;
+        }
+
+        const funcSelector = toFunctionSelector(toFunctionSignature(funcDesc));
+
+        describe(`ABI function ${funcDesc.name} ${funcDesc.stateMutability}`, function () {
+          const isPayable = funcDesc.stateMutability === "payable";
+
+          it("can be called", async function () {
+            const txData = { to: mockAddress, data: funcSelector, gas: 25000n };
```
nddeluca, line 85: Default gas should be a constant

tests/e2e-evm/test/abi_basic.test.ts
```diff
@@ -0,0 +1,162 @@
...
+        });
+
+        it("has a starting balance of 0", async function () {
+          const balance = await publicClient.getBalance({ address: mockAddress });
+          expect(balance).to.equal(0n);
+        });
+      });
+
+      for (const funcDesc of abi) {
+        if (funcDesc.type !== "function") {
```
nddeluca, line 75: Maybe deserves a comment?

tests/e2e-evm/test/abi_basic.test.ts
```diff
@@ -0,0 +1,162 @@
+import hre from "hardhat";
+import type { ArtifactsMap } from "hardhat/types/artifacts";
+import type { PublicClient, WalletClient, ContractName } from "@nomicfoundation/hardhat-viem/types";
+import { expect } from "chai";
+import { Address, toFunctionSelector, toFunctionSignature, concat } from "viem";
+import { getAbiFallbackFunction, getAbiReceiveFunction } from "./helpers/abi";
+import { whaleAddress } from "./addresses";
+
+interface ContractTestCase {
```
nddeluca, line 9: Should this type be inline like the test cases in helpers/abi.test.ts?

tests/e2e-evm/test/extensions/viem.ts
```diff
@@ -49,4 +50,18 @@ export function extendViem(hre: HardhatRuntimeEnvironment) {
 
   hre.viem.getWalletClient = (address: Address, walletClientConfig?: Partial<WalletClientConfig>) =>
     getWalletClient(address, { ...walletClientConfig, ...getChainConfig(hre) });
+
+  hre.viem.deployContract = (async <CN extends string>(
+    contractName: ContractName<CN>,
+    // eslint-disable-next-line  @typescript-eslint/no-explicit-any
+    constructorArgs: any[] = [],
```
nddeluca, line 57: `any` is used in the upstream type, so linter ignore was added instead of correcting it

tests/e2e-evm/test/abi_basic.test.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    // Test receive + fallback scenarios
+    { interface: "NoopReceiveNoFallback", mock: "NoopReceiveNoFallbackMock" },
+    { interface: "NoopReceivePayableFallback", mock: "NoopReceivePayableFallbackMock" },
+    { interface: "NoopReceiveNonpayableFallback", mock: "NoopReceiveNonpayableFallbackMock" },
+
+    // Test no receive + fallback scenarios
+    { interface: "NoopNoReceivePayableFallback", mock: "NoopNoReceivePayableFallbackMock" },
+    { interface: "NoopNoReceiveNonpayableFallback", mock: "NoopNoReceiveNonpayableFallbackMock" },
+  ] as ContractTestCase[];
```
pirtleshell, line 28: nit: i'd prefer to declare type this like `const testCases: ContractTestCase[] =  ...` rather than cast with `as`. 
it's a subtle difference, but i think of it like (declaring) "hey, compiler, these are supposed to be `ContractTestCase`s, plz help" vs (casting) "hey, compiler, these ARE `ContractTestCase`"

tests/e2e-evm/test/helpers/abi.ts
```diff
@@ -0,0 +1,21 @@
...
+  for (const f of abi) {
```
nddeluca, line 4: This is the golang dev in me -- I probably should just use Array#find or similar, tho maybe that's more difficult for typescript to infer the type 🤔

tests/e2e-evm/test/abi_basic.test.ts
```diff
@@ -0,0 +1,162 @@
...
+
+      let mockAddress: Address;
+      before("deploy mock", async function () {
+        mockAddress = (await hre.viem.deployContract(tc.mock)).address;
+      });
+
+      describe("State", function () {
+        it("has code set", async function () {
+          const code = await publicClient.getCode({ address: mockAddress });
+          expect(code).to.not.equal(0);
```
pirtleshell, line 60: can you confirm that an EOA or never-used address has code `0`? i am not 💯 familiar with these operations by i just want to confirm `getCode` isn't returning a string like `"0x"` (which also is not equal to the number `0`)

tests/e2e-evm/test/abi_basic.test.ts
```diff
@@ -0,0 +1,162 @@
...
+        describe(`ABI function ${funcDesc.name} ${funcDesc.stateMutability}`, function () {
+          const isPayable = funcDesc.stateMutability === "payable";
+
+          it("can be called", async function () {
+            const txData = { to: mockAddress, data: funcSelector, gas: 25000n };
+
+            await expect(publicClient.call(txData)).to.be.fulfilled;
+
+            const txHash = await walletClient.sendTransaction(txData);
+            const txReceipt = await publicClient.waitForTransactionReceipt({ hash: txHash });
```
pirtleshell, line 90: oh my, this is so much nicer than doing the same in golang 😂

tests/e2e-evm/test/abi_basic.test.ts
```diff
@@ -0,0 +1,162 @@
...
+            expect(txReceipt.status).to.equal("success");
+          });
+
+          it(`can ${isPayable ? "" : "not "}be called with value`, async function () {
+            const txData = { to: mockAddress, data: funcSelector, gas: 25000n, value: 1n };
+
+            const txHash = await walletClient.sendTransaction(txData);
+            const txReceipt = await publicClient.waitForTransactionReceipt({ hash: txHash });
+            expect(txReceipt.status).to.equal(isPayable ? "success" : "reverted");
```
pirtleshell, line 99: i'd imagine "success" and "reverted" are constants in `viem` we could use. or create our own. would prevent potential test typos going forward

tests/e2e-evm/test/abi_basic.test.ts
nddeluca: You are probably correct here, I believe this is suppose to be checked against undefined -- Viem converts the "0x" response to undefined, I think I confused this assetion with balance

tests/e2e-evm/test/abi_basic.test.ts
nddeluca: I 100% agree and it was the first thing I tried, but typescript was unable to correctly infer contract name type for the mock property with an `as ContractTestCase[]`, so didn't think it was worth duplicating.

Maybe there is another way to hint at the type?  Explicitly typing the strings in mock also worked, but it has to be repeated for every element.

tests/e2e-evm/test/abi_basic.test.ts
nddeluca: Good idea, especially since there is no typecheck on the status.  I'll create constants since https://github.com/wevm/viem/blob/main/src/utils/formatters/transactionReceipt.ts#L23 are typed string constants that are not exported individually.

At some point I plan to refactor these things to custom chai matchers that will reduce a lot the boilerplate and make the tests much more readable.

tests/e2e-evm/test/abi_basic.test.ts
nddeluca: +1 I do enjoy all the built-in functionality with viem/async/await.  The API's feel much better, though golang would be a bit easier to parallelize suites like this.

tests/e2e-evm/test/helpers/abi.ts
nddeluca: +1, I did a pretty fast extraction of those helpers and should be leveraging these other approaches

.github/workflows/ci-default.yml
nddeluca: 🍝

.github/workflows/ci-lint.yml
```diff
@@ -45,6 +45,12 @@ jobs:
       - name: Install npm dependencies
         run: npm install
         working-directory: tests/e2e-evm
+      - name: Run solhint
+        run: npm run solhint
+        working-directory: tests/e2e-evm
+      - name: Compile contracts and create artifcats
```
drklee3, line 51: artif🐱🐱🐱🐱
```suggestion
      - name: Compile contracts and create artifacts
```

tests/e2e-evm/test/abi_basic.test.ts
drklee3: types are weird here yeah since they're generated in the artifacts? I tried some wonky thing like this which works for this section but causes some `"Unsafe assignment of an error typed value."` issues below when using `hre.artifacts.readArtifactSync(tc.interface).abi;`

```typescript
interface ContractTestCase<T extends keyof ArtifactsMap> {
  interface: T;
  mock: ArtifactsMap[T]["contractName"];
}

// ABI_BasicTests assert ethereum + solidity transaction ABI interactions perform as expected.
describe("ABI_BasicTests", function () {
  const testCases: ContractTestCase<keyof ArtifactsMap>[] = [
    // Test function modifiers without receive & fallback
    { interface: "NoopNoReceiveNoFallback", mock: "NoopNoReceiveNoFallbackMock" },

    // Test receive + fallback scenarios
    { interface: "NoopReceiveNoFallback", mock: "NoopReceiveNoFallbackMock" },
    { interface: "NoopReceivePayableFallback", mock: "NoopReceivePayableFallbackMock" },
    { interface: "NoopReceiveNonpayableFallback", mock: "NoopReceiveNonpayableFallbackMock" },

    // Test no receive + fallback scenarios
    { interface: "NoopNoReceivePayableFallback", mock: "NoopNoReceivePayableFallbackMock" },
    { interface: "NoopNoReceiveNonpayableFallback", mock: "NoopNoReceiveNonpayableFallbackMock" },
  ];
...
```

tests/e2e-evm/test/abi_basic.test.ts
drklee3: Could we compare it with an code match from the `deployedBytecode` in the abi instead of a non-match?

tests/e2e-evm/test/abi_basic.test.ts
pirtleshell: heh, potentially hacky workaround that skirts the `ContractName` type troubles:
```
interface ContractTestCase {
  interface: keyof ArtifactsMap;
  mock: `${keyof ArtifactsMap}Mock`;
}
```
but not a major concern lol

pirtleshell (approved):
great test additions to endure we don't regress / alter fundamental evm functionality & assumptions 👍

tests/e2e-evm/test/abi_basic.test.ts
nddeluca: Ditto 👍

PR #2033, opened 2024-10-16
ci: update protonet reset workflow for new infra

pirtleshell:
## Description
Updates the reset network flow for protonet. The new flow should be more resilient. It consists of two steps:

1. stopping the chain services & removing all data
2. reconfiguring the chain & restarting the services

This should keep instances up to date with any infra enhancements going forward.

No changes to the data seeding have been made. The changes do not affect internal testnet. I will be replacing the infra & upgrading that network's flows next.

I have confirmed both the reset workflow & the existing seed workflow succeed via local workflow testing with `act`:
```
# save secrets to a local file, github-action.secrets, then...
act workflow_dispatch --secret-file github-action.secrets -W '.github/workflows/cd-protonet.yml'
```


.github/workflows/cd-protonet.yml
```diff
@@ -1,41 +1,26 @@
...
-      playbook-name: reset-protonet-playbook.yml
-      playbook-infrastructure-branch: master
-    secrets: inherit
+## run after every successful CI job of new commits to the master branch, or on manual dispatch
+on:
+  workflow_run:
+    workflows: [Continuous Integration (Kava Master)]
+    types:
+      - completed
+  workflow_dispatch:
```
pirtleshell, line 9: this workflow supports being triggered by either:
1. a successful run of the CI on merge to master
2. manual dispatch via the github UI

configuring it this way allows us to have a single file instead of two copies with different `on` conditions

PR #2035, opened 2024-10-17
ci: update internal testnet reset workflow

pirtleshell:
Updates CD pipeline for internal testnet, now on newer infra. The action uses the same workflows as the protonet flow, but with different variables.

Confirmed the reset + seeding of network works by running the workflow locally.

Additionally, removed all workflows and scripts that are no longer used.

The workflow is triggered in either of these cases:
1. A commit to master triggers a successful CI run AND the files in `ci/kava-internal-testnet` have changed
2. The workflow is manually triggered via the Github Actions UI

note the above differs from protonet, which deploys on any commit to master. even with `KAVA.VERSION` pinned to `master`, the network will not redeploy on new commits because the files in `ci/kava-internal-testnet` will not change. if the commit is updated or the genesis file changes, the network will be reset & deployed. otherwise, the network will only be reset if the workflow is manually dispatched.


nddeluca:
🚀

.github/workflows/cd-internal-testnet.yml
```diff
@@ -1,60 +1,45 @@
...
+
+## run after every successful CI job of new commits to the master branch, or on manual dispatch
 on:
   workflow_run:
     workflows: [Continuous Integration (Kava Master)]
     types:
       - completed
+  workflow_dispatch:
 
 jobs:
```
nddeluca, line 11: It's great to see re-use of the workflow jobs instead of duplicating them

.github/workflows/cd-internal-testnet.yml
```diff
@@ -1,60 +1,45 @@
...
+            ci/env/kava-internal-testnet/**
 
-  # start kava with new binary and genesis state on api, peer and seed nodes, place nodes in service once they start and are synched to live
-  start-chain-api:
-    uses: ./.github/workflows/cd-start-chain.yml
+  reset-and-restart-chain:
+    # only start cd pipeline if last ci run was successful (or it this is being manually dispatched)
+    if: |
+      github.event_name == 'workflow_dispatch' ||
+      (github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success' && needs.changed_files.outputs.changedInternalTestnetConfig == 'true')
```
pirtleshell, line 34: this is the logic for whether or not to actually reset the network.

it amounts to:
1. reset network if we are manually dispatching the flow. the manual dispatch is a specific choice. if i hit the reset network button, the network should reset regardless of what the content of the HEAD commit is.
2. when a commit to master passes the test CI, reset the network if the files in `ci/kava-internal-testnet` have changed. _only commits that include changes to `KAVA.VERSION` or the genesis file trigger an automated reset_

PR #2021, opened 2024-09-12
test(ante): Add test to cover various EIP712Domain fields

drklee3:
## Description
- Explicitly tests EIP712Domain values for both expected and unexpected
  - Ensure expected domain exclude `verifyingContract` and `salt` fields

Depends on Ethermint PR: https://github.com/Kava-Labs/ethermint/pull/75

Also manually tested EIP712 transaction with the expected EIP712Domain fields with MetaMask.

## Checklist
 - [x] Changelog has been updated as necessary.

PR #2042, opened 2024-10-28
fix!: bump ethermint to fix EIP712 message signing

pirtleshell:
## Description

Previous version of ethermint used a value for the `verifyingContract` in the domain separator that was not compliant with the spec. This field and `salt`, which was unused, have been removed.

This is a state breaking change because it changes which EIP712 txs are considered valid, and therefore, which txs are allowed to be included in a block.

## Checklist

### DIFF
--- a/.github/workflows/proto.yml
+++ b/.github/workflows/proto.yml
@@ -19,8 +19,8 @@ jobs:
       - run: make check-proto-lint
       - run: make check-proto-format
       - run: make check-proto-breaking-remote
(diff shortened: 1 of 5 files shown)


drklee3 (commented):
Can we include the test from #2021 to confirm the patch is integrated properly with the ante handler 😃

pirtleshell:
discussed with @drklee3 who will cherry-pick #2021 onto the release branch 🙏

PR #2046, opened 2024-11-01
chore: setup tcmalloc in rocksdb dockerfile

evgeniy-scherbina:


Dockerfile-rocksdb
```diff
@@ -37,4 +37,5 @@ RUN ldconfig
 # copy kava binary
 COPY --from=kava-builder /go/bin/kava /bin/kava
 
+ENV LD_PRELOAD="/usr/lib/x86_64-linux-gnu/libtcmalloc.so.4"
```
pirtleshell, line 40: can we also support the arm64 images by using `uname -m` instead of hardcoding for `x86_64`?

PR #2048, opened 2024-11-04
docs: call out new rocksdb configurations

pirtleshell:
## Description

Improved IAVL V1 migration guide. Will bring in the full changes to the migration docs for v0.27.


migrate/v0_26/iavl-v1.md
```diff
@@ -21,7 +21,18 @@ make install
 ```
 1. Start kava as usual: `kava start`
 
-**For `rocksdb` nodes:**
+## `rocksdb` nodes
```
pirtleshell, line 24: made these headings so they can be directly linked

PR #2055, opened 2024-11-15
docs: Update iavl v1 guide to include snapshot

nddeluca:
Quicksync now offers a IAVL v1 archive snapshot


karzak (approved):

pirtleshell (approved):
i wonder if we should update the `v0.26.2-iavl-v1` release notes to point to this branch's iavl v1 upgrade guide?

or we should make a similar update to `release/v0.26.x-iavl-v1`

PR #2064, opened 2025-02-12
chore(deps): Bump packet-forward-middleware 

drklee3:
## Description
- Upgrades packet-forward-middleware
- Other deps upgraded as a result of pfm
- swagger.yaml got removed from upstream ibc-go

Note: IBC test is disabled in e2e tests since it affects the upgrade test. Only upgrade test is enabled along with the other basic e2e tests, just to test state migration.

## Checklist


go.mod
```diff
@@ -14,12 +14,12 @@ require (
 	github.com/cometbft/cometbft v0.37.13
 	github.com/cometbft/cometbft-db v0.9.5
 	github.com/cosmos/cosmos-proto v1.0.0-beta.5
-	github.com/cosmos/cosmos-sdk v0.47.10
+	github.com/cosmos/cosmos-sdk v0.47.13
```
nddeluca, line 17: Should be 15?

nddeluca (approved):
Nice, thanks for the adding the upgrade to IBC 7.7 🎸

